PR #5, opened 2020-12-14
[PART 1] First draft of the SSC hardcoded form

fitore-cds:
[PART 1] The initial implementation of the SSC tech-exemption form has the foundation of the 5 pages and Notify submission of the session's data that was input in the form. The final form will come in the next PR.

Also in this PR:
- New environment variables for the SSC form (notify template, email)
- The error templates had a problem which this PR should fix
- Move notify code to a helper function for all forms to use


bryan-robitaille (approved):
Only really reviewed the 400 and 500 request response changes.
Will merge now and tweak the SSC form if needed in future PRs.

PR #15, opened 2021-01-11
Add Form UI components

fitore-cds:
# Summary | Résumé

This PR adds initial form UI components that will eventually be imported via an npm package.

- Components are mapped to JSON configuration via formBuilder
- A new sample form was added with `id=2` to showcase a contact form example, with new fields
- Styling and translation work will continue

---

To test locally, run the app and load this url http://localhost:3000/en/2 (or the French equivalent)


bryan-robitaille (approved):
Great work!  Let's get this in and keep iterating 👍

PR #16, opened 2021-01-11
Small quick changes

bryan-robitaille:
# Summary | Résumé

Quick fix for typescript warning about using type `any`.  Also quick type change for dataLayer formID input from number to string.


fitore-cds (approved):
Thank you!

PR #19, opened 2021-01-11
Esdc form

nmakuch:
# Summary | Résumé

> Added the ESDC form as `esdc.json`


fitore-cds (approved):
Wow, this is an intense form! Great job :-)

PR #22, opened 2021-01-11
Markdown refinement

bryan-robitaille:
# Summary | Résumé

This PR modifies the markdown.js file to TypeScript and removes additional line breaks that were being sent to Notify.


fitore-cds (approved):
:+1:

PR #31, opened 2021-01-19
added section titles and removed titles in fields

mcman12:
- Entered 2 plaintext elements for section titles
- removed the titles in each affected elements

### DIFF
--- a/forms/award.json
+++ b/forms/award.json
@@ -4,7 +4,7 @@
   "publishingStatus": true,
   "submission": {
     "templateID": "",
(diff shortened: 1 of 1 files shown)


mcman12:
@srtalbot  please review before Bryan approves.
https://platform-for-awardfixes-fjvwir.herokuapp.com/

srtalbot (approved):
Looks good, thanks for the changes. 

Tagging design for input, @caycefischer @nmakuch, but I expect we'll need different heading sizes for Form element titles (e.g., given name) and Section breaks (Nominee information)

PR #35, opened 2021-01-20
Feature/add referrerurl confirmation

fitore-cds:
# Summary | Résumé

An interim solution to pass a "Back/Retour" link to the confirmation page, from the JSON. The CDS intake form has an example: https://forms-staging.cdssandbox.xyz/en/1

Also, there's a separate PR for the Storybook documentation. We'll revisit as we define the FormBuilder.

ConfirmationReferrerURL

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,2 @@
 # Required for Notify integration
-NOTIFY_ENDPOINT=
 NOTIFY_API_KEY=
-NOTIFY_TEMPLATE_ID=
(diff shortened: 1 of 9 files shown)


bryan-robitaille (approved):
Let's get this merged :)

fitore-cds:
@bryan-robitaille I just pushed another tweak that simply passes the `required` field to the inputs and relies on the [built-in-browser validation](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#using_built-in_form_validation), for now (to be styled later)

BuiltinBrowserValidation

PR #36, opened 2021-01-20
Desc text

bryan-robitaille:
# Summary | Résumé

This PR adds descriptive text to a form input that follows WCAG guidance by associating the descriptive text to the input with `aria-describedby`.  The titles of the `plainText` have also been modified to H2 elements to ensure page structure for screen readers.


fitore-cds (approved):
Yay :+1:

PR #38, opened 2021-01-20
Add ESDC new form JSON file

caycefischer:
adding first draft of ESDC accessibility grant form

missing dynamic add-field blocks, etc

cc @srtalbot


mcman12 (dismissed):
@caycefischer can you change the form id (row 3) to 18 (8 is already taken). tx

caycefischer:
oops! done, @mcman12

bryan-robitaille (approved):
Just needed to fix a few sneaky issues.
1. The ID 18 was already taken as well so I swapped it out for ID 14.
2. There was an ID in Layout (26) that didn't exist in Elements so it was throwing an error.  I'll add code to ignore that type of bug in the future.

PR #39, opened 2021-01-22
Add En/Fr referrers and remove required fields

fitore-cds:
# Summary | Résumé

- The referrerUrl for the confirmation page now supports providing both EN & FR links.
- Added basic styles for headings in plainText
- Removed some `required` fields in sample JSON forms


bryan-robitaille (approved):
:) Lets get this merged!

PR #43, opened 2021-01-25
Dynamic row for Demo

bryan-robitaille:
# Summary | Résumé

This PR adds a dynamic row element that can leverage all the other question element types to create a repeating group.  This is not a final product but can be used to demonstrate the idea in a demo.  Currently the state is not hooked into the form submission.  
A subsequent PR will include the proper state handling.

To test visit: https://platform-for-dynamic-ro-uexwel.herokuapp.com/en/14


fitore-cds (approved):
Yay, another demo rush about to succeed! (some TypeScript warnings ask for attention)

PR #44, opened 2021-01-25
update copy and add attestation

caycefischer:
changes via cayce and hillary to polish up the form and include missing sections, fixed copy, etc 

FOR DEMO


bryan-robitaille (approved):
Nice!

PR #49, opened 2021-01-25
file input

nmakuch:
### Adding a link to the home page for the a11y funding page 

Screen Shot 2021-01-25 at 16 28 18

### Adding file upload buttons in 2 spots 

Screen Shot 2021-01-25 at 16 29 04

and 

Screen Shot 2021-01-25 at 16 29 39


fitore-cds (approved):
Thank you!

PR #50, opened 2021-01-26
Update docker ignore to force rebuild

bryan-robitaille:
# Summary | Résumé

This PR makes several additions to Docker Ignore file.  Mainly to ignore the local prebuilt static JS  and ensure a clean build on every image.


fitore-cds (approved):
Thank you!

PR #51, opened 2021-01-26
Fixes Warnings and Filters out Props sent to Submit API

bryan-robitaille:
# Summary | Résumé

This PR:
 - fixes COPY on the confirmation page:
 - Fixes a missing key on the component <Heading>
 - Parses the object sent to the Submit API to filter out any unwanted props in the Form json.


fitore-cds (approved):
Nicely done!

PR #60, opened 2021-02-04
Welcome page for Staging

bryan-robitaille:
# Summary | Résumé

This PR introduces a new page `sandbox` that will display the links to all forms that have their `publishingStatus` set to `false`.  The form titles are no longer using static i18n json but instead the internationalization is using the bilingual forms titles as set on the form json object.

The `welcome` page will only display the forms that have their `publishingStatus` set to `true` and implements the same i18n changes.

It is no longer required to manually add links to either page when a new form is uploaded as both pages are now dynamically driven by all forms under the `forms` folder at the root of the project.


fitore-cds (approved):
👍

PR #62, opened 2021-02-04
Add GA code

bryan-robitaille:
# Summary | Résumé

This PR adds the Google Analytics tracking code to the Base page.  The GA script will only render when NODE_ENV is set to production.  Once we have multiple deployment environments (staging and production) we can revisit this setup.

### DIFF
--- a/components/globals/Base.js
+++ b/components/globals/Base.js
@@ -7,9 +7,29 @@ import SkipLink from "./SkipLink";
 import Fip from "./Fip";
 
 const Base = ({ children }) => {
+  const isProduction = process.env.NODE_ENV === "production";
   return (
     <>
       <Head>
+        {isProduction && (
+          <React.Fragment>
+            <script
+              async
+              src="https://www.googletagmanager.com/gtag/js?id=G-KY2EVJV33K"
+            ></script>
+            <script
+              dangerouslySetInnerHTML={{
+                __html: `
+                window.dataLayer = window.dataLayer || [];
+              function gtag(){dataLayer.push(arguments);}
+              gtag('js', new Date());
+
+              gtag('config', 'G-KY2EVJV33K');
+              `,
+              }}
+            />
+          </React.Fragment>
+        )}
         <meta charSet="utf-8" />
         <meta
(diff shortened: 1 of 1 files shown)


components/globals/Base.js
```diff
@@ -7,9 +7,29 @@ import SkipLink from "./SkipLink";
...
     <>
       <Head>
+        {isProduction && (
+          <React.Fragment>
+            <script
+              async
+              src="https://www.googletagmanager.com/gtag/js?id=G-KY2EVJV33K"
+            ></script>
+            <script
+              dangerouslySetInnerHTML={{
```
fitore-cds, line 21: I can approve the PR as is, but was wondering if it's worth iterating without `dangerouslySetInnerHTML` via something like this (not installing the plugin, but adding a dom node?) https://github.com/alinemorelli/react-gtm/blob/master/src/TagManager.js#L41

Also, I remember something funky about client-side routing and maybe we don't need to tinker in the router to but rather set a "Google Tag Manager's History Change trigger": https://support.google.com/tagmanager/answer/7679322?hl=en

Happy to merge this for the demo and iterate later, let me know

bryan-robitaille: `dangerouslySetInnterHTML` is Reacts way of wrapping the example you provided instead of creating a separate DOM node and then injecting it.  [React DOM dangerouslySetInnerHTML](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml).  It basically just saves lines of code but if you want to iterate it out and inject it as a pure DOM we can do that to.

bryan-robitaille: We're currently not using Google Tag manager. (But we can setup if needed.) Since everything is SSR right now I was curious to see if all states would be currently handled by the base script.  Once we know from Research what they want to track I think it would be good at that point to ensure everything's captured?

fitore-cds: Sounds good, let's merge this and add a follow-up card for PM/Research to test and let us know if anything needs to be tweaked.

fitore-cds (approved):

PR #63, opened 2021-02-05
Layout Fix

nmakuch:
## This PR consists of the following

In doing so, I deleted some of the queries that were breaking due to some of the new classes etc. (I'll fix this when I start working on some of the smaller screen styles)

|  Before |  After  |
|---|---|
|  Screen Shot 2021-02-04 at 19 03 51 Screen Shot 2021-02-04 at 19 03 48  |  Screen Shot 2021-02-04 at 19 04 01 Screen Shot 2021-02-04 at 19 03 55 |


fitore-cds (approved):
👍

PR #65, opened 2021-02-05
Cypress form

bryan-robitaille:
# Summary | Résumé

This PR adds some basic cypress tests that check:
- The welcome page renders and language switcher works
- Completes and submits a form (currently CDS Intake Form)


fitore-cds (approved):
What a beauty!

PR #69, opened 2021-02-09
Add custom error pages with translations

fitore-cds:
# Summary | Résumé

As per the [Next.js error guidelines](https://nextjs.org/docs/advanced-features/custom-error-page), I've added two pages to handle 404 or 500 types of errors.

The translation probably needs tweaking, will review with @caycefischer and @bryan-robitaille.


bryan-robitaille (approved):
:) 🎉

PR #71, opened 2021-02-14
Style Cleanup + Media Queries update

nmakuch:
### Custom checkbox and radio inputs
> - Realized the checkbox and radio input styles were never transferred over to the client repo or storybook. Instead we were just rendering default checkboxes, which appear different in different browsers.
> - I've added some css to hide default checkboxes/radios and rendered some custom ones on top. The actual check or radio tick for the checked state are both svgs that I've added to the `public/img` folder
> - This is a first iteration and they can be adjusted later if needed

### GIF
https://recordit.co/rO6SLfHeFp


components/forms/Checkbox/Checkbox.tsx
```diff
@@ -23,18 +23,16 @@ export const Checkbox = (
...
+        className="gc-input-checkbox__input"
+        id={id}
+        type="checkbox"
+        name={name}
+        ref={inputRef}
+        {...inputProps}
+      />
       <label className="gc-checkbox-label" htmlFor={id}>
-        {label}
+        <span>{label}</span>
```
fitore-cds, line 35: Wondering if we could look at the needs of these label `<span>` to see if we need them here and in other places?

fitore-cds (approved):
The UI looks so much better, thank you!

PR #75, opened 2021-02-22
set publish status to false

mcman12:
# Summary | Résumé

the forests forms should not be shown on the main page, so setting the publish status to false.


mcman12:
@nmakuch just for awareness

bryan-robitaille (approved):

nmakuch:
> @nmakuch just for awareness

no prob, I'll probably do a small PR to update the styles on the sandbox page then

nmakuch:
going to merge this

PR #76, opened 2021-02-22
Stringify error before appending CSL

bryan-robitaille:
# Summary | Résumé
This PR does a small refactor to the logMessage functionality.  It now appends the 'client side log' after JSON stringify has run on the error message object/array


fitore-cds (approved):
👍

PR #81, opened 2021-02-24
Ircc form

nmakuch:


fitore-cds (approved):
👍

mcman12:
@nmakuch paragraph should not be in H2, should probably use "plaintext" components. tx

nmakuch:
chatted with Mario a bit on slack but I'm going to merge this in. The form itself is awkward (I converted as is with little design intervention) and has ridiculously long labels (sometimes 2 paragraphs :s). I think a better solution to this would be a content design review as the labels for the questions themselves should probably be reworked.

PR #83, opened 2021-03-02
Update the fileUpload component and tweak "preview" page

fitore-cds:
# Summary | Résumé

This PR contains two updates:
- File upload component: Formik receives the uploaded file, but a separate PR will handle the multipart/formdata update to pass to Notify
- Preview page tweaks (apply a classname to the Base component, via which we can style things on specific urls - will need later for departments)

[Note]: Happy to expand in person, but the `input type="file"` cannot be updated programatically, only via a user interaction. Therefore, as per [Formik suggestions](https://github.com/formium/formik/issues/45), I wrapped the UploadField component with a `Field` one that will allow Formik to receive the updated values via its onChange.


lib/formBuilder.tsx
```diff
@@ -286,6 +286,10 @@ const _getElementInitialValue = (
       dynamicRow[subElementID] = _getElementInitialValue(subElement, language);
     });
     return [dynamicRow];
+  } else if (element.properties.fileType) {
+    // For file attachments, we need several values like the FileName, FileReader base64 object and File object
+    nestedObj = { file: null, src: null, name: "" };
```
bryan-robitaille, line 291: After reviewing the logic I think this could be simplified to:
```
const nestedObj: Record<string, unknown> = {};
....
} else if (element.properties.fileType {
  // For file attachements, we need several values like FileName, FileReader base64 object and File object
  return {file: null, src: null, name: ""}
} else ...
```
It just saves from having to recast the value of nestedObj

bryan-robitaille (approved):
Looks good!  Just one comment regarding possibly refactoring one part of the code in formBuilder.

lib/formBuilder.tsx
fitore-cds: Agreed! Thank you, I fixed it and will tweak this solution after introducing `multipart/form-data`

PR #84, opened 2021-03-02
A11y fix form group

bryan-robitaille:
# Summary | Résumé

This PR replaces the label element on a field-set with the proper legend element.  The legend element is also styled like a field input label.


fitore-cds (approved):
👍

PR #85, opened 2021-03-03
Step 1 in setting up custom validation with Formik, only for required fields

fitore-cds:
# Summary | Résumé

This PR has the initial setup to run a custom validation function during form submission that doesn't rely on the built-in browser functionality. 
- ErrorMessage component is used to render field errors
- There's a list of errors shown at the top of the Form if invalid (needs Design help)
- Radio/Checkbox validation needs more work, will tackle in the next PR with the custom types (like email etc)
- I've also updated the JSON config for forms to introduce the `validation` object and updated documentation


bryan-robitaille (approved):
Great work!

PR #87, opened 2021-03-04
Need to include next-i18next config

bryan-robitaille:
# Summary | Résumé

This PR fixes an issue with the Docker Image.  The new i18n package requires the config file to also be available after the NextJS package is built.


fitore-cds (approved):
💯

PR #91, opened 2021-03-09
Add basic regex validation for common types

fitore-cds:
# Summary | Résumé

This PR adds the ability to specify the type of the field (email, number, text, phone, date) or a custom regex in the `validation` object in the Form config.

Currently, only the Platform Intake form takes these extra validators under account, but we should introduce them to our demos moving forward.

@bryan-robitaille I'm debating whether to introduce [Yup](https://github.com/jquense/yup) now or as a later iteration?


lib/validation.tsx
```diff
@@ -1,6 +1,14 @@
...
-import { FormikProps } from "formik";
+import {
+  FormValues,
+  DynamicFormProps,
+  InnerFormProps,
+  FormElement,
+  Responses,
+  ValidationProperties,
+} from "./types";
+import { FormikProps, FormikValues } from "formik";
```
bryan-robitaille, line 10: Just a quick removal of `FormikValues` which isn't being used in this file.

lib/validation.tsx
```diff
@@ -52,3 +69,41 @@ export const getErrorList = (
...
+    };
+  }
+
+  const REGEX_CONFIG: RegexProps = {
+    email: {
+      regex: /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$/,
+      error: t("input-validation.email"),
+    },
+    text: {
+      regex: /^[0-9a-zA-Z]+$/,
```
bryan-robitaille, line 87: Maybe we should add`-` as a valid character given hyphenated composed names?

lib/validation.tsx
```diff
@@ -52,3 +69,41 @@ export const getErrorList = (
...
+      error: t("input-validation.text"),
+    },
+    number: {
+      regex: /^[0-9]+$/,
```
bryan-robitaille, line 91: Maybe we should add `,` and `.` as valid characters as well given some people use `,` to separate every 10<sup>3</sup> and `.` due to floats.

bryan-robitaille (changes requested):
I think some quick changes to the regex's would really benefit the UX for now until we have more information coming back from research.

lib/validation.tsx
fitore-cds: Good catch, thank you!

lib/validation.tsx
fitore-cds: Yes, good idea I updated the regex for text and number.

bryan-robitaille (approved):
Awesome!

PR #92, opened 2021-03-10
rough draft of TSB/BST contact form

caycefischer:
cc @fitore-cds @srtalbot


mcman12 (approved):
Looks good, I changed the name of the file to add the .json extension

caycefischer:
Note: no context or preamble text on this form - will be coming from the TSB page https://www.tsb.gc.ca/eng/contactez-contact/index.html so will want to know from them what's required - specifically, to improve this form once we've reviewed with TSB and learned their needs, I'd suggest:

- adding guidance around when to use this form vs the other options on that page (reporting accidents, etc)
- adding in guidance around when to call vs when to use this form
- ??? more, as we learn from meeting with them

PR #106, opened 2021-03-22
Prod conversion

bryan-robitaille:
This PR modifies the Submission API to communicate with the AWS API Gateway which feeds the reliability queue.

Other small tweaks/fixes:

- Fix logging API that was receiving a non-compliant JSON format
- Add axios and swap out fetch.
- Add github action workflow to push built container to ECR in production environment
- Added error handling when the Submission API returns an error.  Shows Alert error at top of page.  Auto scroll to error alert.
- Added email preview on confirmation page when not in production


fitore-cds (approved):
This is so exciting!!!

PR #107, opened 2021-03-22
Add validation types to TSB form

fitore-cds:
# Summary | Résumé
- Validate email and text fields via `type` in the TSB form


bryan-robitaille (approved):
lgtm

PR #109, opened 2021-03-22
Error styling + small rich text update

nmakuch:
### 1. Updating the error message styles and error summary styles
- Small tweaks to both

### 2. Small update to rich text styles
- A little hard to style based on the rich text component but added some small changes


fitore-cds (approved):
💯

PR #110, opened 2021-03-23
Load Privacy content from markdown files

fitore-cds:
# Summary | Résumé

- This PR adds the ability for the app to load markdown files and render their content via RichText
- Markdown files can be added under `/public/static/content/[en|fr]`
- Added Privacy page in EN/FR and updated the Footer link. Terms page is there as a placeholder, meantime no links to it.
- The homepage is a simple splash page with two links to English/French, will ask Design to tweak its appearance

Note: I think we can optimize loading markdown files on the server-side, down the road. Had trouble loading the correct locale and decided to push the Privacy page as a first iteration.


bryan-robitaille (approved):
:)

PR #114, opened 2021-03-30
feat: add a11y tracking to staging env

maxneuvians:
This PR adds a call to the a11y tracking tools into the CI pipeline. After the client component has been deployed it will make an API call to the a11y tracking tools API that request a spidering of the public facing pages of both the english and french domains.


bryan-robitaille (approved):
👍

bryan-robitaille:
Noticed that both website start points will hit the language splash page.  Is the spider configuration `1` only going to go one level deep or it references another config option that has nothing to do with levels? lol

PR #115, opened 2021-03-31
Production

bryan-robitaille:
# Summary | Résumé

This PR adds:
- Github action to build and push container to the Production environment but NOT deploy it.
- Add an environment variable `PRODUCTION_ENV` that is available at both build and runtime to act as a feature flag between development, staging, and production.
    - Production:
      -  use lambda submit function and return status code based on it's results
      - do not preview notify template
      - add Google Analytics tag
      - set a timeout of 3 seconds on the browser side call to the NextJS submit api.
      - Make sandbox and unpublished form id's non-routable
    - Staging:
      - use lambda submit function in an async way and do not care about it's return value
      - use preview notify template and return value to browser
      - No Google Analytics tag
      - no timeout on browser side call to NextJS submit api
    -  Local Development:
      - Do not use lambda submit function
      - use preview notify template and return value to browser
      - No Google Analytics tag
      - no timeout on browser side call to NextJS submit api
- Add an alert error when the browser receives a 500 response from the NextJS submit API and scroll to error.
- Add a Notify Email Template preview on the form confirmation screen.
- Add new Google Analytics code id in header
- Only show `published forms` on the welcome page
- Make the sandbox page with unpublished forms not routable
- Make direct links to unpublished form id's not routable

# Reviewer checklist | Liste de vérification du réviseur
- [ ] Have you tested it? | L’avez-vous testé?
- [ ] Does this break existing functionality? | Est-ce que ça brise une fonctionnalité existante?
- [ ] Should this be split into smaller PRs to decrease change risk? | Est-ce que ça devrait être divisé en de plus petites demandes de tirage (« pull requests ») afin de réduire le risque lié aux modifications?
- [ ] Does this change the privacy policy? | Est-ce que ça entraîne une modification de la politique de confidentialité?
- [ ] Does this introduce any security concerns? | Est-ce que ça introduit des préoccupations liées à la sécurité?
- [ ] Does this significantly alter performance? | Est-ce que ça modifie de façon importante la performance?
- [ ] What is the risk level of using added dependencies? | Quel est le degré de risque d’utiliser des dépendances ajoutées?
- [ ] Should any documentation be updated as a result of this? (i.e. README setup, etc.) | Faudra-t-il mettre à jour la documentation à la suite de ce changement (fichier README, etc.)?


fitore-cds (approved):
Super exciting!

PR #119, opened 2021-04-07
NextJS version upgrade and form URL modification

brdunfield:
I didn't see any documentation on nextjs' [10.1.0 page](https://nextjs.org/blog/next-10-1) about any breaking changes, and the app seems to work fine locally for me.

I also added `/id/` to form URLs in this PR as requested by @bryan-robitaille . So a form that was previously at `/20` is now at `/id/20`.


fitore-cds (approved):
Looks great, thank you!

There is one "hidden" use-case to preview JSON config, wondering if you could add the `id` here too? (so  change to `pathname: "/id/preview-form"`) https://github.com/cds-snc/platform-forms-client/blob/main/pages/builder-preview.js#L14

bryan-robitaille:
We'll wait till the demo at 2:30 EST is completed before merging... otherwise poor @mcman12 might be a little freaked out if for some reason something breaks.

PR #122, opened 2021-04-07
Fix Erroring logic on submit.js API

bryan-robitaille:
# Summary | Résumé

This PR refactors the submit.js API endpoint to ensure that the response headers are only ever set once.  This was an issue given the async nature of the dependant functions and the unique circumstances of the staging environment sending to both Lambda and Notify.


fitore-cds (approved):
A super tricky one, fingers crossed!

PR #123, opened 2021-04-07
Enable customizing the text on the confirmation page, per form

fitore-cds:
# Summary | Résumé

This PR adds the ability to customize the text that appears on the confirmation page, by passing in markdown in the form's JSON config via `endPage`.

I updated the TSB one for now, as they're currently the only ones customizing the confirmation page. All other forms will continue showing the standard confirmation message upon submission.

**Note**: We'll revisit this solution once we introduce pagination


brdunfield (approved):
LGTM for an MVP and works great!

PR #127, opened 2021-04-14
Update git ignore to exclude storybook static directory

brdunfield:
I was running Storybook locally and noticed it added close to 50 files under this directory. We probably want to ignore this directory so nothing gets accidentally committed?


bryan-robitaille (approved):
👍

PR #128, opened 2021-04-15
Jest tests

bryan-robitaille:
# Summary | Résumé

Big PR that brings up our Jest testing to 71 tests with a coverage of over 90%.
- There are also some fixes along the way for a11y (aria-describedBy was not properly being filtered down to the component).
- There are failing tests that require fixes to the code.  The tests are correct but there are bugs in the actual code.  Those bugs will be addressed in a separate PR.
- There are still some outstanding tests on the `Page` components but given that they are basically just containers, it won't be prioritized for the time being.

Please ignore all the older commits referenced on this branch.. we won't see them after the squash on the main branch... Dirty history because of the origin ancestor being the production branch instead of main.


fitore-cds (approved):
Amazing! Herculean efforts, thank you🥇

PR #130, opened 2021-04-16
Preliminary pass at improving validation

brdunfield:
Changes:
- Added an "Alphanumeric" field type which only accepts letters and numbers for input. ( [WIP] just occurred to me that it may need to accept accented characters as well....)
- Added the name and text types. We probably don't need both of these, as the validation on the text type currently only is filtering out newline characters.
- Finessed the email type a little to look for at least two characters after the .
- Added a Trello card about date validation - I think this will be a separate issue and will not be touched in this PR, aside from potentially an error message update for the short term (needs translation)
- Added spaces to allowable characters for number fields

I have modified the error message for the new alphanumeric type as follows:
> Please enter letters and numbers only

This will need translation.
I also do not have an error message for the text type as I'm not sure whether this type is even necessary. If so, I am not sure What the error message should be.

As per @fitore-cds recommendation, I have put a preliminary message in this PR, and will create a Trello card to have design and mario validate and iterate if needed.

I have also added functionality to preserve the text field type when relevant. in the TextInput component (previously the type was hard coded to "text"). Most of our validation types don't map to an html input type, but a few like email and phone do, so I have mapped these and passed along the specific type. This also allows us to enable browser autofill for these fields, which was identified as a best practice in the research docs given to me by Hillary.


fitore-cds (approved):
💯

PR #131, opened 2021-04-20
french translations

mcman12:
# Summary | Résumé
- changed french translations to reflect english changes
- added phone number format in english validation 
-


brdunfield:
I think the phone number format is actually not needed. We support multiple types of format based on my manual tests. And from the research I've seen, being format agnostic is better for the user. It might be best to just strip out the phone number format from both the english and french messages?

fitore-cds (approved):
Merci!

brdunfield (approved):
LGTM!

PR #132, opened 2021-04-20
Fix TSB email validation bug

brdunfield:
A non email field was set to type: email, causing validation issues


bryan-robitaille (approved):
👍

PR #134, opened 2021-04-21
Pagination state sharing

brdunfield:
[WIP] As I need to clean up a few things in the TextPage component and fix all the tests.

This PR attempts to encompass existing pages like the confirmation page (and future pages such as a start page) into the DynamicForm component, so that state and formMetadata can more easily be shared.

To that end, I've created the concept of a form "step" in the code and in the routes. In the future, when we add multi part pagination, we can leverage this step value to load / navigate between different steps of forms.

Pages like the start and end pages can also be considered as "steps" of the form. Next will look for a value after the form ID in the URL to parse the step: `/id/{formid}/{stepvariable}` . The DynamicForm component now parses the step value found in the route, and can decide what to render based on what it receives. For now, the only "step" implemented is 'confirmation', viewed at `/id/{formid}/confirmation`. Any value other than 'confirmation' will just be ignored and the basic form page will render as usual.

TextPage.tsx is essentially a slightly generalized copy and paste of the old pages > confirmation.js file, which no longer exists.


components/containers/DynamicForm/DynamicForm.tsx
```diff
@@ -3,14 +3,20 @@ import classnames from "classnames";
...
 import Head from "next/head";
 import { Form } from "../../forms/Form/Form";
+import { TextPage } from "../../forms/TextPage/TextPage";
 import { getProperty, getRenderedForm } from "../../../lib/formBuilder";
 import { FormMetadataProperties } from "../../../lib/types";
 import { useRouter } from "next/router";
 
 interface DynamicFormProps {
   formMetadata: FormMetadataProperties;
+  step: string;
```
brdunfield, line 13: I think I actually don't need this

brdunfield:
[Share state across Form pages](https://trello.com/c/aV0jilCy/631-share-state-across-form-pages)

lib/dataLayer.tsx
```diff
@@ -207,11 +207,11 @@ async function _submitToApI(
             : "";
         router.push(
           {
-            pathname: `/${language}/confirmation`,
+            pathname: `/id/${formMetadata.id}/confirmation`,
```
bryan-robitaille, line 210: Thinking you meant to modify this line to `/${language}/id/${formMetadata.id}/confirmation` to render `/en/id/2/confirmation`

lib/dataLayer.tsx
```diff
@@ -207,11 +207,11 @@ async function _submitToApI(
...
             query: { ...referrerUrl, htmlEmail: htmlEmail, pageText: endPageText },
           },
           {
-            pathname: `/${language}/confirmation`,
+            pathname: `/id/${formMetadata.id}/confirmation`,
```
bryan-robitaille, line 214: Thinking you meant to modify this line to `/${language}/id/${formMetadata.id}/confirmation` to render `/en/id/2/confirmation`

brdunfield:
[Confirmation page needs to function like a Form step](https://trello.com/c/7qzSEASD/649-confirmation-page-needs-to-function-like-a-form-step)

lib/tests/dataLayer.test.js
```diff
@@ -136,14 +136,14 @@ describe("Submit Form data", () => {
       .toBeCalledTimes(1)
       .toBeCalledWith(
         expect.objectContaining({
-          pathname: "/en/confirmation",
+          pathname: "/id/1/confirmation",
```
bryan-robitaille, line 139: In order to support the language rendering this would need to check for `"/en/id/1/confirmation"`.

lib/tests/dataLayer.test.js
```diff
@@ -136,14 +136,14 @@ describe("Submit Form data", () => {
...
           query: {
             htmlEmail: undefined,
             referrerUrl: expect.any(String),
             pageText: expect.any(String),
           },
         }),
-        expect.objectContaining({ pathname: "/en/confirmation" })
+        expect.objectContaining({ pathname: "/id/1/confirmation" })
```
bryan-robitaille, line 146: In order to support the language rendering this would need to check for `"/en/id/1/confirmation"`.

bryan-robitaille (changes requested):
A few little tweaks just to enable the i18n routing for the confirmation page.  It's on my radar to write a test checking for the french submission confirmation path... my bad.

bryan-robitaille (approved):
😃  👍

PR #135, opened 2021-04-22
Fix blank confirmation page due to stringify/parse

fitore-cds:
# Summary | Résumé

Considering that the confirmation page doesn't receive the `pageText` from the URL query anymore, rather via the `formMetadata` property, we don't need to strigify/parse the values.


bryan-robitaille (approved):
Test on Heroku - Good to go!

PR #137, opened 2021-04-26
Remote File Inclusion - AWS Firewall blocking form submissions

bryan-robitaille:
# Summary | Résumé

This PR has a few small changes:
- Aligns logic around form values and form data to center on the `form.elements` array as the source of truth and leaves `form.layout` only the source of truth for form rendering
- Filters out any `richText` elements from the `formMetaData` passed to the submit API to ensure that any urls in richText do not block the form submission at the  AWS firewall. This is only needed until we can host the form templates in their own DB or API.
- Removes check for element ID's in the `layout` property that do no exist in elements.  This is likely to occur now that we are filtering out the richText elements from the array however their ID still appears in the `layout` property.  The associated test for this check is also removed.

### DIFF
--- a/lib/dataLayer.tsx
+++ b/lib/dataLayer.tsx
@@ -61,24 +61,19 @@ function _getSubmissionByID(formID: string): Record<string, unknown> {
 function _rehydrateFormResponses(payload: Submission) {
   const { form, responses } = payload;
   const rehydratedResponses: Responses = {};
-  form.layout.forEach((qID: string) => {
-    const question = form.elements.find((element: FormElement) => element.id === qID);
-    if (question) {
-      const response = responses[question.id];
-      switch (question.type) {
-        case "checkbox":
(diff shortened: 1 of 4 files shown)


fitore-cds (approved):
👍

brdunfield:
So, to check my understanding, rich text fields are blurbs / instructions we can inject into the forms, and don't actually have any submit data attached? In which case, it makes sense from a lot of angles to strip it out like you've done here. I don't think we ever really need to submit that, even once we have a DB etc...
If I'm correct with my understanding, then LGTM 🦑 !

brdunfield (approved):

bryan-robitaille:
That's exactly right @brdunfield.  The `richText` elements don't have any submit data attached.  Looking forward to only having to pass the form values back to the Submit API in the near future instead of all of the formMetaData and values.

PR #141, opened 2021-04-28
Add block style to logo so that focus state appears when the logo has focus

brdunfield:
I tried inline-block but that made our default SVG disappear. Block seems to work with both .pngs and .svgs


fitore-cds (approved):
👍 

@nmakuch FYI

brdunfield:
[Manual a11y review and fix for TSB Form](https://trello.com/c/jCSflmR3/653-manual-a11y-review-and-fix-for-tsb-form)

PR #142, opened 2021-04-29
Add react polyfills for IE

bryan-robitaille:
# Summary | Résumé

This PR adds polyfills for IE11 as well as stable browsers (>0.2% global usage)


brdunfield:
I am not sure what I should be looking for in order to test this (not sure what specific issues we had), but I do have access to Browserstack so happy to take a look once I know what to look for!

mcman12:
Looks great now, layout, left indent, etc.
Submit doesn,t seem to run though, get back to the same page without seeing confirmation. tx

brdunfield (approved):
LGTM, however just want to flag an issue with checkbox display in IE11 - we can log this as a separate bug so as not to hold up this PR.

image

PR #143, opened 2021-04-29
Render and read the asterisk on required fields more accessibly

brdunfield:
As per WCAG, meaningful content should not be inserted using CSS pseudo elements. This PR moves the asterisk indicated a required field into a <span> within the label element. I've also added an aria label to it so that screenreaders will read "required" instead of "star" or "asterisk" when reading it out.

I'm making this as a draft because currently the new asterisk is a separate readout from the label, requiring an extra tab to navigate through, and i'd like to improve that.
UPDATE: upon further research this is how screen readers are expected to work, and so I'm marking this as ready for review.

### DIFF
--- a/components/forms/Label/Label.tsx
+++ b/components/forms/Label/Label.tsx
@@ -8,10 +8,11 @@ interface LabelProps {
   error?: boolean;
   hint?: React.ReactNode;
   srOnly?: boolean;
+  required?: boolean;
 }
 
 export const Label = (props: LabelProps): React.ReactElement => {
-  const { children, htmlFor, className, error, hint, srOnly } = props;
+  const { children, htmlFor, className, error, hint, srOnly, required } = props;
 
   const classes = classnames(
     {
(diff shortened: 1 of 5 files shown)


brdunfield:
[Manual a11y review and fix for TSB Form](https://trello.com/c/jCSflmR3/653-manual-a11y-review-and-fix-for-tsb-form)

components/forms/Label/Label.tsx
```diff
@@ -24,7 +24,7 @@ export const Label = (props: LabelProps): React.ReactElement => {
 
   return (
     <label data-testid="label" className={classes} htmlFor={htmlFor}>
-      {children}
+      {children} {classes.includes("required") ? <span aria-label="required">*</span> : null}
```
fitore-cds, line 27: Happy to merge it the `className` detection, just wondering if it's worth passing the `required` prop to the Label component, as the corresponding fields get it too, e.g. https://github.com/cds-snc/platform-forms-client/blob/main/components/forms/TextInput/TextInput.tsx#L14

brdunfield: Might be better to have it match the other components. I can move it over!

fitore-cds: Thank you!

PR #146, opened 2021-04-30
A11y iterations for LanguageToggle and required fields

brdunfield:
# Summary | Résumé
Updated the accessibility of the LanguageToggle component, by removing the `<h2>` and replacing with a label.
My biggest outstanding complaint with the language toggle is that when you change the page while using a screenreader, the screenreader does not give any indication that the page or content has changed. Ideally it would read out the new title of the page or somehow otherwise indicate that the content has changed. It looks like this is an area of active work for next.js so we should keep an eye on their updates.

Also iterated on the required field (*) readout - as Screenreader was still reading out the asterisk.
Now it reads out 
`label name ->  "required" -> input description` in successive tabs. Still not perfect but I'm happier with it.

This iteration also required translating the "required" word so that it reads out correctly en français. Let me know if you'd rather I did the importing differently, as this necessitated me mocking out the library for tests.
Also @mcman12 if you could check the translation I used for "required" to make sure you're happy with it!

### DIFF
--- a/components/forms/Label/Label.tsx
+++ b/components/forms/Label/Label.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import classnames from "classnames";
+import { useTranslation } from "next-i18next";
 
 interface LabelProps {
   children: React.ReactNode;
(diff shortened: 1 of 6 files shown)


fitore-cds (approved):
👍  (I still wonder when to use `em` vs `i`) LGTM!

PR #147, opened 2021-05-03
Checkbox input mis-aligned on multi-line label

bryan-robitaille:
# Summary | Résumé

The checkbox and radio input element were normally hidden behind the styled label-input except when a label spanned multiple lines.  This PR aligns the input element and the label element so that one is always hidden by the other.

# Visual Changes
| Before | After |
|:-----:|:------:|
|Screen Shot 2021-05-03 at 12 12 54 PM|Screen Shot 2021-05-03 at 12 15 49 PM|

### DIFF
--- a/styles/_forms.scss
+++ b/styles/_forms.scss
@@ -174,7 +174,7 @@
   @apply block;
 
   .gc-input-checkbox__input {
-    @apply md:mr-4 mr-6 absolute w-10 h-10 bg-white-default border-2 border-transparent;
+    @apply absolute -left-full;
     & + .gc-checkbox-label {
       @apply relative cursor-pointer p-0;
     }
--- a/styles/_forms.scss
+++ b/styles/_forms.scss
@@ -174,7 +174,7 @@
   @apply block;
 
   .gc-input-checkbox__input {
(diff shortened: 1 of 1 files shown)


mcman12:
Check and it looks good. thanks Bryan

brdunfield:
I'm not sure that the left -100% is necessary, as when I disable it I don't see the original issue replicated.

I'm hesitant to style it off the page, as any user without CSS enabled won't be able to see it (probably not a super common issue these days, but something to consider..)

brdunfield:
On further reflection if they don't have CSS enabled it won't style off the page. It's monday

brdunfield:
Not something that's related to what you're fixing in this PR, but the vertical alignment of the checkboxes when the text gets longer than a line or two looks funky to me
Screen Shot 2021-05-03 at 10 37 18 AM

brdunfield (approved):
LGTM. I think checkbox styling will need iteration eventually as per my other comments, but this PR does what it says it does! 🦑

brdunfield (approved):
I like this approach a lot better! The alignment looks much cleaner too. 🦑

PR #148, opened 2021-05-03
Enable input type and autocomplete for "name" type 

brdunfield:
We already supported the "name" type in the json for validation purposes. This PR just extrapolates this and allows input type="name" and autocomplete="name" to be possible now when this type is used.

Enabling these properties when possible is a best practice for accessibility: https://www.w3.org/TR/WCAG21/#input-purposes


fitore-cds (approved):
LGTM!

PR #149, opened 2021-05-04
Wait for i18n to initialize

bryan-robitaille:
# Summary | Résumé

This PR implements a fix that is caused by the next-i18next package that leverages react-i18next under the hood.  It tries to use `Suspense` however given that some of the pages are rendered server side the `Suspense` function doesn't exist.

Solution posted here on  [NextJS github repo](https://github.com/vercel/next.js/issues/22508)


fitore-cds (approved):
Damn. Thanks for this one!

PR #150, opened 2021-05-04
Fix accessibility issues with error messages 

fitore-cds:
# Summary | Résumé

In this PR, there are two fixes for validation accessibility:
1. When errors appear at the top of the form, the focus will shift to that error container
2. When clicking on an error link in this container (error alert box), you’ll be “scrolledIntoView” of the label that sits above the input field and the field will have focus on it (previously it would scroll directly into the input field, so the label text couldn't be seen)

*Note: open to future improvements, if we find a way to not use JS

### DIFF
--- a/components/forms/Form/Form.tsx
+++ b/components/forms/Form/Form.tsx
@@ -1,8 +1,8 @@
-import React from "react";
+import React, { useEffect } from "react";
 import { withFormik, FormikProps } from "formik";
 import getConfig from "next/config";
 import { getFormInitialValues } from "../../../lib/formBuilder";
-import { validateOnSubmit, getErrorList } from "../../../lib/validation";
+import { validateOnSubmit, getErrorList, setFocusOnErrors } from "../../../lib/validation";
 import { submitToAPI } from "../../../lib/dataLayer";
 import { Button, Alert } from "../index";
 import { logMessage } from "../../../lib/logger";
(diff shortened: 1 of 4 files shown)


components/forms/Form/Form.tsx
```diff
@@ -19,18 +19,32 @@ const InnerForm = (props: InnerFormProps & FormikProps<FormValues>) => {
...
     <>
       {formStatus ? <Alert type="error" heading={formStatus} /> : null}
       {errorList ? (
-        <Alert type="error" heading={t("input-validation.heading")} validation={true}>
+        <Alert
+          type="error"
+          heading={t("input-validation.heading")}
+          validation={true}
+          id={errorId}
+          tabIndex={0}
```
brdunfield, line 46: Tab index! 👍 Great idea!

fitore-cds: :-)

components/forms/Form/Form.tsx
```diff
@@ -19,18 +19,32 @@ const InnerForm = (props: InnerFormProps & FormikProps<FormValues>) => {
...
 
   const errorList = props.errors ? getErrorList(props) : null;
+  const errorId = "gc-form-errors";
   const formStatus = props.status === "Error" ? t("server-error") : null;
   if (formStatus) {
-    window.scrollTo({
-      top: 0,
-      behavior: "smooth",
-    });
+    window.scrollTo({ top: 0, behavior: "smooth" });
```
brdunfield, line 25: It's a minor issue and doesn't affect functionality, but the "smooth" scroll doesn't appear to be working anymore.

brdunfield: it's probably being written by the useEffect below - in which case we can probably just remove this?

lib/validation.tsx
```diff
@@ -10,6 +10,71 @@ import {
...
+ * @param e The click event of the error link
+ * @param id The id of the input field that has the error and we need to focus
+ */
+const scrollErrorInView = (e: React.MouseEvent<HTMLAnchorElement, MouseEvent>, id: string) => {
+  e.preventDefault();
+  const labelElement = document.getElementById(`label-${id}`);
+  const inputElement = document.getElementById(id);
+  if (labelElement && inputElement) {
+    inputElement.focus();
+    labelElement.scrollIntoView();
```
brdunfield, line 74: 👍 👍

brdunfield (approved):
This is really cool, and I didn't know about useEffect! Happy to learn about this.

LGTM! 🦑 

The only possible change is looking at that window.scrollTo in Form.tsx that's no longer smooth scrolling, and see if it can be removed.

components/forms/Form/Form.tsx
fitore-cds: Good catch, this one pertains to server-side errors from the submission lambda (either a notify error, or connection etc). But, even though it's not part of the validation, you're completely right that it should have the same accessibility fix applied to it, within the useEffect. 

Updated the branch, thanks for your help!

PR #153, opened 2021-05-04
Scaffold out a basic templates endpoint

brdunfield:
Adds a templates API endpoint, and a couple new (but ugly) UI pages:
- Upload JSON - paste a JSON config file contents into the box and click the button - it will be uploaded into the DB
- DataView - displays all the "templates" that are currently stored in the DB

The templates API endpoint connects to the new "templates" lambda.

I am partitioning my work here so that I can unblock Bryan. I will iterate on functionality / UI / features in a new PR :)


bryan-robitaille (approved):
Look forward to seeing where you take it from here 😃

PR #158, opened 2021-05-05
Update partnerships_intake.json

srtalbot:
- Updated the content in the element one to set better expectations
- updated the email to the one provided by partnerships. Waiting on confirmation it's been set-up before going live. 
- linted :)


mcman12 (approved):
good to go

PR #167, opened 2021-05-10
Handle IE11's lack of support for pino logger

fitore-cds:
# Summary | Résumé

- IE11 doesn't support pino logger and the error from this module prevents our forms from submitting in that browser
- There's now a default logger object that relies on the console functions
- I've added a condition to asynchronously load pino in browsers that are not IE11
- *** Unrelated to pino, I've downloaded the [IE11 Tailwind polyfill](https://css-tricks.com/color-theming-with-css-custom-properties-and-tailwind/#ie11customproperties) into our /public folder just in case they stop serving/supporting it in the future

I could use a second set of eyes on the async loading so I'm doing a PR to the `polyfill` branch rather than merging directly.


fitore-cds:
Found a way to transpile `pino` for IE11 and that fixed the problem.

PR #168, opened 2021-05-10
Authentication using CDS Google accounts

bryan-robitaille:
# Summary | Résumé

This PR adds the ability to leverage Google accounts associated with the CDS organization to login to the future admin/form builder side of the application.

In order to simplify JWT token encryption/decryption and life cycle states of authentication we're using [NextAuth.js](https://next-auth.js.org/) to manage state and auth providers.

There are 2 ways to implement auth on a page...
Client side:
```
import { useSession } from 'next-auth/client'
const [session] = useSession();
// session will be null if no user is logged in
```
Server Rendered Side:
We have a wrapper function that can be implemented on getServerSideProps that will redirect a user who is not logged in and if there is a session available, inject it into the context which can then passed on as a prop.
```
export const getServerSideProps = requireAuthentication(async (context) => {
  if (context.user){
    // get some authenticated only data
  }
  if (context.locale) {
    return {
      props: {
        ...(await serverSideTranslations(context.locale, ["common", "admin-login"])),
      },
    };
  }
  return { props: {} };
});
```
The page will then receive all the props provided form the getServerSideProps and a prop called user which has the name, email and image of the user logged in.


brdunfield (approved):
There are a couple ESLint warnings in `/pages/admin/login.tsx` that we might want to look at.

Otherwise, LGTM and works well locally! 🦑

PR #169, opened 2021-05-11
Update TSB-contact.json

srtalbot:
Updating the form based on review with TSB:
- Updating checkbox copy, as provided by TSB
- Adding an "other" checkbox, as requested by TSB
- Adding a check-box option for copyright and permissions requests


mcman12 (approved):
tested and true

PR #178, opened 2021-05-13
New focus branch

nmakuch:
## This PR consists of the following:

Left out some of the buttons because they don't exist in the repo yet, could create the components next week or tomorrow if I have time. This would be a separate PR most likely.


forms/award.json
```diff
@@ -12,32 +12,7 @@
...
-      18,
-      19,
-      20,
-      21,
-      22,
-      23,
-      24,
-      25,
-      26,
+      1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26,
```
brdunfield, line 15: can we run prettier on this PR to get rid of all these unrelated changes?

brdunfield:
In the figma file, the focus style around a text link has some padding around the text. In the PR app, it seems to be missing
Screen Shot 2021-05-14 at 11 52 24 AM

brdunfield:
This seems to also apply to focus on the error text
Screen Shot 2021-05-14 at 12 02 37 PM

nmakuch:
> In the figma file, the focus style around a text link has some padding around the text. In the PR app, it seems to be missing
> Screen Shot 2021-05-14 at 11 52 24 AM

I found no padding was actually better. I've noticed with the padding it makes things on the page shift around as focus is moving from element to element.

brdunfield:
One other difference I notice between the figma file and the app:
In figma, the checkbox or filled radio button is blue to match the focus, while in the app it's still black.
Screen Shot 2021-05-14 at 12 03 59 PM

brdunfield:
> > In the figma file, the focus style around a text link has some padding around the text. In the PR app, it seems to be missing
> > Screen Shot 2021-05-14 at 11 52 24 AM
> 
> I found no padding was actually better. I've noticed with the padding it makes things on the page shift around as focus is moving from element to element.

That makes sense. Hmmm. My worry here is that without the padding the text looks visually difficult for me to read, because the white text is only 1px away from the white border. I'm not sure where this falls on the compliance side however, that's just my gut opinion.

nmakuch:
> > > In the figma file, the focus style around a text link has some padding around the text. In the PR app, it seems to be missing
> > > Screen Shot 2021-05-14 at 11 52 24 AM
> > 
> > 
> > I found no padding was actually better. I've noticed with the padding it makes things on the page shift around as focus is moving from element to element.
> 
> That makes sense. Hmmm. My worry here is that without the padding the text looks visually difficult for me to read, because the white text is only 1px away from the white border. I'm not sure where this falls on the compliance side however, that's just my gut opinion.

very valid point! maybe there's a way to combine it with the outline or boxshadow property to give it some extra padding without moving elements on the page

public/img/blue-checkmark.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
```
brdunfield, line 1: Can we get rid of the black checkmark and circle files? Or are they still being used elsewhere?

nmakuch: they're being used when the inputs don't have focus

brdunfield: Ah, right. Nevermind!

brdunfield (approved):
👍 Looks great!

PR #179, opened 2021-05-13
use repo prettier settings

bryan-robitaille:
# Summary | Résumé

This PR removes the prettier action that auto formats and recommits files as it was causing issues with the other actions.  There appears to be a new bug somewhere where the commit is causing checks to run with the new commit ref which wasn't happening in the past.  The action now only checks that the prettier format is ok and will fail if any changes are required.  I have also added the vscode config for the workspace to leverage the prettier config specified so that developers on the project don't need to remember to configure it specifically.


brdunfield (approved):
LGTM! 🦑

PR #183, opened 2021-05-17
tbs-review4

srtalbot:
- adding back button on end page in EN and FR
- Updating cypress test
- Validated JSON

Note: Mario, I've seen us use "retour" before, can you confirm that is correct?


mcman12 (approved):
tested french and english submissions plus back links

PR #185, opened 2021-05-18
Get vault submission to nextJS page

bryan-robitaille:
# Summary | Résumé

This PR creates a new page `/admin/vault` and a new api endpoint `/api/retrieval` that request through a lambda function submissions that have been stored in the vault for a specific form.  Both the `/admin/vault` page and `/api/retrieval` api are protected through auth.  Designs are obviously to be iterated upon.

| Search for Form 45 | Response 1 | Response 2 |
|:-----:|:----:|:----:|
|Screen Shot 2021-05-19 at 1 39 39 PM| Screen Shot 2021-05-19 at 1 40 14 PM | Screen Shot 2021-05-19 at 1 40 24 PM

### DIFF
--- a/pages/api/submit.js
+++ b/pages/api/submit.js
@@ -41,11 +41,12 @@ const callLambda = async (form, fields) => {
   const lambdaClient = new LambdaClient({ region: "ca-central-1" });
 
   const command = new InvokeCommand({
-    FunctionName: process.env.SUBMISSION_API ?? "",
+    FunctionName: process.env.SUBMISSION_API ?? "Submission",
     Payload: JSON.stringify({
       form,
       responses: fields,
       submission,
+      vault: true,
     }),
   });
   return await lambdaClient
--- a/lib/dataLayer.tsx
+++ b/lib/dataLayer.tsx
@@ -16,7 +16,7 @@ import {
 
 // Get the form json object by using the form ID
 // Returns => json object of form
-function _getFormByID(formID: string): Record<string, unknown> {
+function _getFormByID(formID: string): FormMetadataProperties {
   // Need to get these forms from a DB or API in the future
   let formToReturn = null;
   for (const form of Object.values(Forms)) {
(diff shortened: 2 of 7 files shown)


pages/api/submit.js
```diff
@@ -41,11 +41,12 @@ const callLambda = async (form, fields) => {
   const lambdaClient = new LambdaClient({ region: "ca-central-1" });
 
   const command = new InvokeCommand({
-    FunctionName: process.env.SUBMISSION_API ?? "",
+    FunctionName: process.env.SUBMISSION_API ?? "Submission",
     Payload: JSON.stringify({
       form,
       responses: fields,
       submission,
+      vault: true,
```
brdunfield, line 49: Not related to this PR, but eventually will we want this to be an option? As in, folks could use either email or vault (or any other) submission method we support?

bryan-robitaille: Exactly!

fitore-cds: Yup, I think I'll add a minimal "feature flag" functionality to our app, where we can turn features of/off via `.env` variables. This is a good candidate, and showing/hiding the "Login" link in the header is another :-)

brdunfield (approved):
LGTM and worked well when I tested locally. I'm not sure why your tests are failing - doesn't look like it failed on a file you changed.

fitore-cds (approved):
What a beauty, love seeing those screenshots! Post on the team channel?

pages/api/submit.js
bryan-robitaille: Just a reminder that env vars add complexity.  For every env var change there are 3-4 files in terraform that need to be changed and redeployed.

bryan-robitaille: On Portal we used a package that used a DB table to store flags that were controlled through Admin Page access.
Really help cut down on the number of infrastructure changes needed as new features were added.

PR #186, opened 2021-05-18
Minor update to the Dashboard page to link to the newly created admin pages

fitore-cds:
# Summary | Résumé

My humble contribution to the dashboard page as I'm catching up to the sandbox work :-)

### DIFF
--- a/pages/admin/index.tsx
+++ b/pages/admin/index.tsx
@@ -17,19 +17,36 @@ const AdminWelcome: React.FC<AdminWelcomeProps> = (props: AdminWelcomeProps) => 
   return (
     <>
       <h1 className="gc-h1">{t("title")}</h1>
-      <div>
-        <h2 className="gc-h2">
-          {i18n.language === "en" ? "Welcome" : "Bienvenue"} {user.name}!
-        </h2>
-        {user.image ? <img src={user.image} alt="profile" /> : null}
-        {t("logged-in")} {user.email} <br />
-        <Button type="button" onClick={() => signOut()}>
-          {t("button.logout")}
-        </Button>
+      <div className="grid grid-cols-3 gap-4">
+        <div className="w-60">
+          <h3 className="gc-h3">
+            {i18n.language === "en" ? "Welcome" : "Bienvenue"} {user.name}!
+          </h3>
+          {user.image ? <img src={user.image} alt="profile" className="rounded-full" /> : null}
+          <p className="text-sm">
+            {t("logged-in")} {user.email}
+          </p>
+          <Button type="button" onClick={() => signOut()}>
+            {t("button.logout")}
+          </Button>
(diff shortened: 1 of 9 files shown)


brdunfield:
PR app isn't working. Is that expected? Not sure the state of Heroku with sandbox apps

pages/api/submit.js
```diff
@@ -73,7 +73,7 @@ const previewNotify = async (form, fields) => {
   );
 
   const emailBody = await convertMessage({ form, responses: fields });
-  const messageSubject = `${form.emailSubjectEn ? form.emailSubjectEn : form.titleEn} Submission`;
+  const messageSubject = form.emailSubjectEn ? form.emailSubjectEn : form.titleEn;
```
brdunfield, line 76: Was this change intentional? Looks unrelated to the admin page so just checking :)

pages/admin/index.tsx
```diff
@@ -17,15 +18,36 @@ const AdminWelcome: React.FC<AdminWelcomeProps> = (props: AdminWelcomeProps) =>
...
+
+          <Button type="button" onClick={() => signOut()}>
+            {t("button.logout")}
+          </Button>
+        </div>
+
+        <div className="w-60">
+          <h3 className="gc-h3">Create Forms</h3>
+          <p>
+            <Link href="/admin/upload">View Form templates</Link>
```
brdunfield, line 39: This should probably say "Upload Form Templates" instead of "View"

pages/admin/index.tsx
```diff
@@ -17,15 +18,36 @@ const AdminWelcome: React.FC<AdminWelcomeProps> = (props: AdminWelcomeProps) =>
...
+          </p>
+        </div>
+        <div className="w-60">
+          <h3 className="gc-h3">View Existing Forms</h3>
+          <p>
+            <Link href="/admin/view-templates">View Form templates</Link>
+          </p>
+          <p>
+            <Link href="/admin/vault">View Form submissions</Link>
```
brdunfield, line 48: this doesn't technically exist yet but it will soon so I don't have a problem leaving it here

pages/api/submit.js
bryan-robitaille: Yes, that was part of another PR that got merged into Main to support the Transport Safety Board requested changes.  It aligns with what was done on the main branch.

fitore-cds: Yes, I snuck this in - it's unrelated but I keep forgetting to update. Thanks both :-)

pages/admin/index.tsx
fitore-cds: Good catch, thank you! Fixed :-)

brdunfield (approved):
Looks great! Thank you! 🦑

PR #187, opened 2021-05-18
Expose GET method in api/templates, and don't crash the app if the lambda times out

brdunfield:
# Summary | Résumé
Restructuring the api/templates route to actually accept and handle post POST and GET requests, to hopefully unblock @bryan-robitaille 

Also added a quick fix to just return null if the lambda has timed out instead of bringing down the whole app. Will brainstorm methods to reduce lambda timeouts tomorrow!

### DIFF
--- a/pages/admin/upload.js
+++ b/pages/admin/upload.js
@@ -41,7 +41,10 @@ const handleSubmit = async (e) => {
     headers: {
       "Content-Type": "multipart/form-data",
     },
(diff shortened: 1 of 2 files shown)


bryan-robitaille (approved):
Once I added the TEMPLATES_API env var locally it worked :). Noticed on the first run it threw back an error but on subsequent runs everything worked.

PR #190, opened 2021-05-19
Focus feedback from Julianna and Eric

nmakuch:
## This PR consists of the following:

I wasn't happy with some inconsistencies in the focus states for the forms components and submit button, but after some discussion with Eric and some feedback from Julianna I've made some updates with their help.

The changes are as follows:

### One consistent focus colour
- Links previously had a separate focus colour but have been updated along with the form inputs to have one darker shade of blue that I was previously using for active states in my previous PR briefly.

### Updating the buttons to have a small outline
- Worried the colour switch alone won't be enough for visually impaired users
- The buttons have been updated to have a small solid white box shadow (in the case this button is ever applied to a darker background), followed by a 2 px blue outline to stand out more.

### DIFF
--- a/public/img/blue-checkmark.svg
+++ b/public/img/blue-checkmark.svg
@@ -3,7 +3,7 @@
 <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
 	 viewBox="0 0 24 24" style="enable-background:new 0 0 24 24;" xml:space="preserve">
 <style type="text/css">
-	.st0{fill:#001F97;}
+	.st0{fill:#3500C7}
 </style>
 <path class="st0" d="M20.3,2L9,13.6l-5.3-5L0,12.3L9,21L24,5.7L20.3,2z"/>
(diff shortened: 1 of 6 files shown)


brdunfield:
This looks pretty good to me. My only worry is that the blue is dark enough that it might be difficult to distinguish it from the black non focussed elements? Particularly in a radio or checkbox situation? Was that already discussed with Eric / Julianna?

I really like the button outline change!

nmakuch:
> This looks pretty good to me. My only worry is that the blue is dark enough that it might be difficult to distinguish it from the black non focussed elements? Particularly in a radio or checkbox situation? Was that already discussed with Eric / Julianna?
> 
> I really like the button outline change!

It was, but they haven't seen it live, I've actually just sent them the heroku link to check out so I'll make sure to get the OK from them before doing anything

brdunfield (approved):
Perfect! Well LGTM codewise, so once you have the approval on the colour you're good to go 👍

nmakuch:
> Perfect! Well LGTM codewise, so once you have the approval on the colour you're good to go 👍

Sounds good :D Thanks Bethany!

PR #192, opened 2021-05-19
Add white text to active state for mobile

nmakuch:
## Context

With the way `hover` works on mobile it was triggering the hover state and the active state at the same time. The result was blue text on a dark background (gross!). 

The active styles have been updated to have white text so they will overwrite the hover state as the button is touched.


fitore-cds (approved):
👍

PR #193, opened 2021-05-20
Refactor templates

bryan-robitaille:
# Summary | Résumé
This PR extracts the logic from the templates API into the dataLayer lib.  This allows us to call the lambda functionality server side from `getServerSideProps` on the NextJS Pages instead of writing fetch/axios requests that connect to an API endpoint.  The is also the refactor of `getFormByID, getFormByStatus, getSubmissionByID` to use the stored forms in AWS Aurora DB instead of using the local versions in the app.

The `crudTemplates` function is essentially that.  It's built to handle all the Create, Read, Update, and Delete of the templates that are stored in the AWS RDS Cluster.

Next Steps...  
- align with @brdunfield as to how this affects her current draft PR and provide any help as needed to fix conflicts.
- remove the id from the FormMetadataProperties object as it is now defined in the DB as the primary key.
- write new tests that cover the `getFormByID, getFormByStatus, getSubmissionByID` that require significant mocking.

### DIFF
--- a/lib/dataLayer.tsx
+++ b/lib/dataLayer.tsx
@@ -74,10 +74,10 @@ async function _crudTemplates(payload: CrudTemplateInput): Promise<CrudTemplateR
       if (response.FunctionError) {
         //throw Error("Templates API could not process json");
         // temporary more graceful failure here
-        logMessage.info("Lambda Client not successful");
+        logMessage.info("Lambda Template Client not successful");
         return null;
       } else {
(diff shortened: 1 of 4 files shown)


bryan-robitaille:
Will create a more generic function that also checks if it's being used client or server side and either call the API or use the direct function accordingly.

brdunfield:
This is looking really great so far. I don't think the merge will be that bad either.

fitore-cds (approved):
💯

PR #194, opened 2021-05-21
quick fix to make the form links on /welcome-bienvenue and /sandbox work

brdunfield:
Previously they were linking to /undefined because we've removed id from the JSON


brdunfield:
This is addressed in Bryan's PR, so closing this one :)

PR #195, opened 2021-05-21
Focus colour tweak after some Sim Daltonism tests

nmakuch:
The bright blurple was failing some basic colour blindness checks so it needs to be tweaked. Worked with Eric to settle on this more diluted version of the previous colour.


brdunfield (approved):
LGTM! 🦑

PR #198, opened 2021-05-25
Uncomment cypress for sandbox branch

brdunfield:
# Summary | Résumé

We previously had to comment out the cypress specs because of the id changes. This PR updates the URLs to pull from the correct IDs. It also pulls the form data from the DB in order to match against the page info.

I also deleted all the local copies of the form json files, as we're no longer using them.


fitore-cds (approved):
Why is deleting files so satisfying? :-)

PR #199, opened 2021-05-26
Hide header/footer if embedding a form within an iframe

fitore-cds:
# Summary | Résumé

- If we pass the `?embed=true` parameter to a form url, the header and the footer won't be shown
- Updated docs to indicate how to embed in an iframe

```
<iframe src="https://forms-formulaires.alpha.canada.ca/id/20?embed=true" title="Form description title"></iframe>
```


brdunfield (approved):
This LGTM!
I think in the future we may need to iterate on things like CSS margins and breakpoints for embedded forms depending on how the embed feature is actually being used downstream, but this is great for now!

PR #202, opened 2021-05-26
[WIP] Show admin navigation on authenticated pages

fitore-cds:
# Summary | Résumé

- This is the first iteration of a basic admin navigation that appears below the FIP if the user is authenticated
- The `Base` component will receive `user` as a prop, if it's on an authenticated page


brdunfield:
It doesn't seem like Logging out works for me? (Tested locally)

fitore-cds:
@brdunfield Thank you, I fixed the "Logout" in the nav - turns out I couldn't rely on the `Link` component so switched it to a `button`. Will leave it with Nick to improve :-)

brdunfield:
The PR review app seems to not have built properly.. /welcome-bienvenue is totally broken. But it works locally so I'm not too concerned. 🤔

brdunfield (approved):
This looks great! Thanks Fitore! 🦑

PR #203, opened 2021-05-27
Add redirects and confirmations for uploads and deletes

brdunfield:
# Summary | Résumé
This PR adds several small UI updates to help give feedback on user actions.

- When creating a form, the page redirects to the settings page for the newly create form on success.
- When updating the JSON of an existing form on the settings page, a small success message appears beside the upload button once the request is successful
- When deleting a form, the user is first asked to confirm the deletion before it goes through - to make deleting a little less easy to accidentally do


bryan-robitaille (approved):
🎉  LGTM 👍

PR #204, opened 2021-05-27
Suspense

bryan-robitaille:
# Summary | Résumé

This PR introduces a new component that can be used as a Loading / Busy spinner.  It leverages the [react-spinners](https://github.com/davidhu2000/react-spinners) library.  The spinner is currently set on the Submit form page while the form validates and connects to the lambda to handle the form submission.

A11y - Tested in Mac OS Voice Over.  It was awesome.

| French Spinner | English Spinner|
|:----:|:----:|
|french_suspense|english_suspense|

### DIFF
--- a/components/globals/Loader.tsx
+++ b/components/globals/Loader.tsx
@@ -1,0 +1,23 @@
+import React from "react";
+import { css } from "@emotion/react";
+import Loader from "react-spinners/PacmanLoader";
+
+interface CustomLoaderProps {
+  loading: boolean;
+  message: string;
+}
+
+const override = css`
+  display: block;
+`;
+
+export const CustomLoader = ({ loading, message }: CustomLoaderProps): React.ReactElement => {
+  return (
+    <div role="status">
+      <p className="pb-6">{message}</p>
+      <Loader loading={loading} color="#00703C" size="100px" margin="2px" css={override} />
+    </div>
+  );
+};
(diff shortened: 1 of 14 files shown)


brdunfield:
Do you think we could also add this to the new admin pages? Particularly upload and settings when the form json is uploaded?

components/globals/Loader.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import { css } from "@emotion/react";
+import Loader from "react-spinners/PacmanLoader";
```
brdunfield, line 3: As much as I love it, I feel like the PacmanLoader might not be the _best_ choice.. 🤔

bryan-robitaille: lol I know... It's more of a fun placeholder until design provides their input.  I can also swap it out for something more 'standard'.

brdunfield:
> > Do you think we could also add this to the new admin pages? Particularly upload and settings when the form json is uploaded?
> 
> Definitely I can amend this PR. Wanted to check in with you first @brdunfield because I know you were modding those pages for state updates as well.

Yep! All my changes should have been in: https://github.com/cds-snc/platform-forms-client/pull/203
which was merged yesterday!

brdunfield (approved):
LGTM! 🦑

PR #207, opened 2021-06-02
Add that tests also run on push to branches

bryan-robitaille:
# Summary | Résumé

This PR ensures that the automated tests are run during both the PR as well as the merge.  If the tests were successful on the PR then they *should* also pass on the Merge however it's best to be safe rather then sorry.  This follows the same deployment process as Covid Alert Portal.


fitore-cds (approved):
👍

PR #212, opened 2021-06-09
Added "Delete all responses" per form to the vault

brdunfield:
# Summary | Résumé

After a form with at least one entry is looked up in the vault, a "Delete all Responses" button now appears below the form entries. This uses similar logic to the Delete Form button I implemented on the Settings page (in that there's a confirmation and second button before the request goes through).

I also added a small useEffect() hook to allow the "Look up Responses" button to display a different amount of entries if it has changed since the page was loaded. Example - if you have a vault tab open and a tab open and submit a form, the vault tab wouldn't pick up on the new submission without a hard refresh.

Feel free to use form 196 or a new form to test.

I added this function to the vault page instead of as an option when deleting a form on the Settings page to avoid outstanding policy questions that will be answered in the future. I felt this was the safest implementation to leverage the feature for now.

### DIFF
--- a/pages/admin/vault.tsx
+++ b/pages/admin/vault.tsx
@@ -35,6 +35,13 @@ const FormResponse = ({
   const [submissionID, setSubmissionID] = useState("");
   const { t } = useTranslation("admin-vault");
 
+  useEffect(() => {
+    if (Items.length != submissionArray.length) {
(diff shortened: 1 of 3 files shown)


bryan-robitaille (approved):
LGTM!

PR #220, opened 2021-06-11
Multiple choice validation

brdunfield:
# Summary | Résumé

This PR allows for proper validation of Multiple Choice fields (Radio, Checkbox):
- These fields now correctly display the 'required' asterisk / screen reader elements
- These fields are now actually validated against (i.e. if they are empty and required, the form will not submit)

Luckily this doesn't appear to have broken submission. Tested with form 52 (contains a required radio), and form 42 (contains some non required checkboxes).


lib/formBuilder.tsx
```diff
@@ -136,46 +135,50 @@ function _buildForm(element: FormElement, lang: string): ReactElement {
...
+          label: choice,
+          required: isRequired,
+        };
       });
 
       return (
         <FormGroup name={id} ariaDescribedBy={description ? `desc-${id}` : undefined}>
           <div className="focus-group">
-            <legend className="gc-label">{labelText}</legend>
+            {labelComponent}
```
brdunfield, line 150: This change is to make sure we show the asterisk and screen reader elements for a "required" field

bryan-robitaille (approved):
LGTM! 🎉

PR #225, opened 2021-06-15
Update README.md

millette:
# Summary | Résumé

Remove English paragraph from French section.


bryan-robitaille:
Thank you for your PR @millette .  Not only have you found an issue with our README but apparently one of our GitHub actions as well.  We'll try to fix the action in the next day or two so that we can get your PR approved and merged.
Sorry for the delay and thanks for your understanding.

bryan-robitaille (approved):
Thank you for catching this and your patience in getting this merged.

PR #227, opened 2021-06-15
Add api route to view app version

brdunfield:
# Summary | Résumé
Adds `/api/version` which shows the current app version from the VERSION file, and also the current contents of the changelog file.

Also syncs the version in package.json with the new VERSION file. We'll need to note that we need to keep these values in sync (not that anything bad will happen if we don't, it's just good practice)


bryan-robitaille (approved):
LGTM!

PR #228, opened 2021-06-15
Make form templates page more readable

brdunfield:
# Summary | Résumé

A quick sandbox update to the 'view form templates' admin page.

Since we now have a form settings page, and we have quite a lot of forms, printing out the form JSON on this page was not very useful, and not very readable.
I've streamlined the page to just show the id and internal name of the form, and the full JSON can be viewed on the settings page by clicking the Edit button

Screen Shot 2021-06-15 at 12 19 10 PM

### DIFF
--- a/components/containers/Dashboard/DataView.tsx
+++ b/components/containers/Dashboard/DataView.tsx
@@ -46,10 +46,14 @@ const DataElement = (props: { template: FormDBConfigProperties }): React.ReactEl
 
   let title = "";
   if (template.formConfig) {
-    if (i18n.language === "en" && template.formConfig.internalTitleEn)
-      title = template.formConfig.internalTitleEn;
-    else if (i18n.language === "fr" && template.formConfig.internalTitleFr)
-      title = template.formConfig.internalTitleFr;
+    if (i18n.language === "en")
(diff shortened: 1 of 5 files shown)


components/containers/Dashboard/DataView.tsx
```diff
@@ -34,33 +34,37 @@ export const DataView = (props: DataViewProps): React.ReactElement => {
...
-  };
-
   const redirectToSettings = (formID: number) => {
     router.push({
       pathname: `/${i18n.language}/id/${formID}/settings`,
     });
   };
 
+  let title = "";
+  if (template.formConfig) {
```
bryan-robitaille, line 48: Wondering if we should set a fall back here in case internalTitleEn/Fr doesn't exist (the type def has it as possibly null) and leverage the `formConfig.form.titleEn/Fr` which we know can't be null.

bryan-robitaille (approved):
🦩  Pure Awesomeness 🦩 
Just one little comment on possibly adding a fallback.

components/containers/Dashboard/DataView.tsx
brdunfield: Good call. I'll add that in.

Issue #215, opened 2021-06-10
Add Feature Flags to the app


brdunfield:
This will allow us to work on larger features more in sync with staging / production, without being widely available.

bryan-robitaille:
Merged into sandbox.

PR #229, opened 2021-06-15
Feature flags

bryan-robitaille:
# Summary | Résumé
This PR introduces the ability to set feature flags through the admin panel in the GCForms app.
References Issue #215 

To check if a feature flag is available client side use the `useFlag("flagName")` hook from `@lib/hooks/flags`.
To check if a feature flag is available server side use `checkOne("flagName")` function from `@lib/flags`

The feature flags are initialized by an object stored in the `./pages/api/flags/[key]/check` file.  Next JS doesn't have a nice way of initializing the application without ejecting and using a custom server.  By including it in the check api code but external to the exported block the code is only run when the file is first initialized by NextJS.  The initialization code will check to see if there are any new feature flags that need to be created with their default values or remove any expired/retired flags from stored Array.

There is also a Redis mock that runs in memory in case the env var `REDIS_URL` is not set.  This will facilitate jest testing in multiple scenarios.

## Architecture

| Client Side  | Server Side APIs | Server Side Data Layer | Redis Instance |
| ------- | --------- | --------- | -------- |
| `useFlag()` Hook | `enable` / `disable` / `check` / endpoints | `checkOne()` / `checkAll()` / `enableFlag()` / `disableFlag()` / `createFlag()` / `removeFlag()` | |

## Screen Shots
Translations and text are only temporary until we get official copy
| English | French |
| ---- | ---- |
|Screen Shot 2021-06-15 at 5 18 48 PM|Screen Shot 2021-06-15 at 5 19 25 PM|

If you've made it down the PR this far then there's also a small bonus.....
This PR also reintroduces hot module reload for translations.  This allows developers to modify and create translations keys in real time without having to restart the local development server each time to see the changes take effect.


brdunfield (commented):
Woohoo hot translation reload!

I'm wondering if we should document the feature flag use, as well as which ones are currently supported, either in the readme or in storybook docs somewhere?

Also, just noting there's some outstanding eslint warnings and cypress tests. I'm not sure how big of a lift getting that rolling again will be 😬

bryan-robitaille:
Updated the README with the env configuration for local development and fixed issues that the Cypress tests brought to light.
For docs about which features are supported I think we can do that in another PR as these were only really taken for proof of functionality.  We'd probably want to figure out how we want to lay them out eventually or if we want to link various features as a dependencies.

PR #231, opened 2021-06-16
Remove annotations and commenting from GitHub actions for now

bryan-robitaille:
# Summary | Résumé

Remove annotations from GitHub actions so that external collaborators can not have github actions fail on PR's due to lack of write access.

Will improve by implementing pull_request_target in separate action that handles artifacts and writes to the PR in the future.


brdunfield (approved):
🚢

PR #232, opened 2021-06-16
Round out component validation tests

brdunfield:
# Summary | Résumé
Recently, #220 exposed that our validation was non functional for checkboxes and radios.

In this PR I have written more tests around validation for existing components, to ensure that both a) the asterisk and other associated 'required' elements render on a required field, and b) an error is thrown if the field is empty.

I heavily rewrote validation.test.js to leverage jest.each using test cases. I converted all the existing text validation cases to this format, and then added one for testing whether a required field was filled for each other type of input element we currently have.


bryan-robitaille (approved):
🧪   This is just AWESOME!  🧪 
Can't wait to see the next Tests enhancement.

PR #233, opened 2021-06-17
Initialize Feature flags

bryan-robitaille:
# Summary | Résumé

This PR removes the need for the `check` API to be called to initialize the feature flags.  The initialization for development and production will now run as a separate script as part or `yarn dev` or `yarn start`.  The initial flag settings are declared in `./flag_initialization/default_flag_settings.json`.
If no REDIS_URL is declared the application will still function by falling back on an in-memory mock version of Redis.  This in-memory version is initialized the first time a Redis connection is required.  This situation will likely arise during automated testing so to ensure that no functionality is hidden from the tests all Flags are initialized enabled.


flag_initialization/index.js
```diff
@@ -0,0 +1,73 @@
...
+      for (const key in currentFlags) {
+        if (typeof initialFlags[key] === "undefined" || initialFlags[key] === null) {
+          console.log(`Removing flag: ${key} from flag registry`);
+          await removeFlag(key, redis);
+        }
+      }
+      return checkAll(redis);
+    })
+    .then(async (currentFlags) => {
+      console.log("Checkign for New Flags");
```
brdunfield, line 51: Nitpicky but there is a typo here! (Checkign)

brdunfield (approved):
Works great! LGTM aside from that tiny typo :)

🦑

flag_initialization/index.js
bryan-robitaille: Good catch!

PR #240, opened 2021-06-24
Temp routing

bryan-robitaille:
# Summary | Résumé
This PR enables temporary routing for existing published form Ids.  A grace period of 1 week will be provided to allow for our clients to change their website links to leverage the newly assigned form Ids after which the temporary routing in code will be removed.

This PR also fixes:
 - French translation issues on the 404 page.
- Adds context.locale to redirects invoked in the getServerSideProps functions to ensure the user is sent to the proper locale page.


brdunfield (approved):
Awesome! 🦑

PR #261, opened 2021-06-30
Quick fixes

bryan-robitaille:
# Summary | Résumé

This PR removes left over debugging `console.logs()`, reduces the production level logger level to `warn` from `info`, and fixes an issue when the Reliability Queue feature is turned off.  

When the 'Send to Reliability Queue' feature is turned off the Submit API was returning an error to the form client when it should have been returning a 200 success.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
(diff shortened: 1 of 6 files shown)


brdunfield:
Looks like the changes to one of the files triggered a few lint warnings..

bryan-robitaille:
Adjusted Jest test to look for expected `warn` level instead of `info` level.

brdunfield (approved):
LGTM! 🦑

PR #270, opened 2021-07-06
Optimize page generation

### LINKED ISSUE #263: Address Template Lambda request overload
During performance testing it was found that at roughly 30 requests/s the Template Lambda became overloaded and returned 429 status (TooManyRequests) errors.  The Template Lambda is currently invoked from many different application pages (Not just the form filling pages).  
Possible Mitigation Strategies:
1. Implement a cache on the Next JS server side to serve up template configurations.  When the JSON upload function is triggered it invalidates the cache and refetches the changes with the Template Lambda.
2. Leverage the RDS read-only endpoint and connect directly to it from the Next JS application instead of using the Template Lambda to fetch form templates.

bryan-robitaille:
# Summary | Résumé

Leverages a cache to now serve up frequently used Template API requests.  Cache is relatively short lived as the entry will expire after a random time of between 30-60 seconds. 

# Unresolved work / Out of scope | Travail non résolues ou hors sujet
Including a non client side compatible package in `@lib/dataLayer` raised the issue of sharing a lib that is used both client and server side.  A custom webpack config had to be included in `next.config.js` to ensure the application could build.  The importing of the `@lib/cache` module into `@lib/dataLayer` had the effect of trying to import `ioredis` into client side code which threw many errors about missing modules (dns, tls, net, etc.)

## Tech Debt / Refactoring Work
We should better organize/separate server side functions from client side to ensure there are no cross contamination in terms of imported modules or uncertainty of what code is being run where.

closes #263


brdunfield (approved):
👍 
LGTM!
I tested the /welcome-bienvenue and /sandbox pages locally, and verified that my local lambda was not being called with every page load.

I didn't test loading a specific form because I don't currently have any on my local platform, but I can put some in and test if we're worried about it.

Issue #222, opened 2021-06-14
Add Authorization Model to application


brdunfield:
An Authorization Model supports user management, allowing us to offer more features securely without risk of someone seeing data they are not authorized to see.

Rough overview:

We already have basic user management through Google SSO.
- Build out user management more:

- [x]  Persist user accounts in the DB
- [x]  Extend default user model to include admin role

- Add / build out the concept of an organization

- [ ] Forms belong to organizations
- [ ] Users also belong to organizations
- [ ] Users can only view forms that are attached to their organization
- [ ] Basic admin page(s) to manage organizations and users

 
The will require in some changes to how forms are saved / retrieved within the app.

PR #292, opened 2021-07-08
Auth model: Organisations

brdunfield:
# Summary | Résumé

Part of #222 

This PR adds the client side functionality to create and manage "organisations" within the app. A future PR will allow users to be assigned to these organisations, to facilitate the authorization model.

Things done as part of this PR:
- CRUD support for organisations
- `/organisations` API route
- Organisation overview page, at `/admin/organisations` (similar to `/admin/view-templates`)
- Individual organisation settings pages at `/admin/organisations/[id]`
- Create organisation page at `/admin/organisations/create`
- organisations lambda calls are cached

- Abstracted template deletion button into its own component so it could be easily re-used (this is a component that prompts an "Are you sure" and a second button before executing the deletion. I still consider this to be a temporary component until this flow can be better designed)

- Users are now persisted in the database, and the default user model has been extended to allow for admin roles and organisation assignment (although this is not yet being used)


components/containers/Admin/Organisations/CreateOrganisation.tsx
```diff
@@ -0,0 +1,59 @@
+import React from "react";
+import axios from "axios";
+import { useTranslation } from "next-i18next";
+import Head from "next/head";
+import { Label } from "../../../forms/Label/Label";
+import { TextInput } from "../../../forms/TextInput/TextInput";
+import { Button } from "../../../forms/Button/Button";
```
bryan-robitaille, line 7: just an fyi that these types of imports have been setup so that you can call `import { Button } from "@components/forms"`

components/containers/Admin/Organisations/OrganisationSettings.tsx
```diff
@@ -0,0 +1,73 @@
+import React, { Fragment } from "react";
+import { useTranslation } from "next-i18next";
+import Head from "next/head";
+import { Organisation } from "@lib/types";
+import { DeleteButton } from "../../../forms/Button/DeleteButton";
```
bryan-robitaille, line 5: This can be added to the forms package index in `/.components/forms/index.ts` for easier imports.

components/containers/Admin/Organisations/OrganisationSettings.tsx
```diff
@@ -0,0 +1,73 @@
...
+      method: "DELETE",
+      organisationID: organisationID,
+    },
+    timeout: 0,
+  })
+    .then((serverResponse) => {
+      //success - redirect to view-templates page
+      return serverResponse;
+    })
+    .catch((err) => {
```
bryan-robitaille, line 30: We might want to test this pattern of catching the error.  I just noticed this pattern in a couple of places.
- If there's an error in axios it'll get caught in the `.catch(err)` block.
- The error is then logged to console.   (We should probably migrate these to logMessage.error eventually)
- The body of the error is then returned to `resp` which returns the key `status`.  My concern if that if the `status` key doesn't exist in the error object then `undefined` is being returned.  
- I'm wondering if we should bubble up an error or explicitly return a `status: 500` to ensure the `status` key is always available.

I know the logic that calls this function explicitly checks for a `200` status code and if not send the error message to console again.  We can refactor this in a future PR as I know this is just to continue the work with the DB in staging.

lib/dataLayer.tsx
```diff
@@ -510,10 +512,151 @@ async function _submitToAPI(values: Responses, formikBag: FormikBag<DynamicFormP
...
+          };
+        default:
+          return {
+            data: {},
+          };
+      }
+    });
+  }
+
+  const lambdaClient = new LambdaClient({
```
bryan-robitaille, line 626: Just a placeholder but performance testing showed that we should pull the LambdaClient initialisations out of the function level and up to the module level.  Out of scope for this PR but just an FYI.  We'll be able to reuse the same `lambdaClient` for all the invocations in this file.

pages/admin/organisations/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { requireAuthentication } from "../../../lib/auth";
+import { crudOrganisations } from "@lib/dataLayer";
+
+export const getServerSideProps = requireAuthentication(async (context) => {
+  {
+    // getStaticProps is serverside, and therefore instead of doing a request,
+    // we import the invoke Lambda function directly
+    const lambdaResult = await crudOrganisations({ method: "GET" });
+
+    const organisations =
```
bryan-robitaille, line 12: There's a nifty new ES feature that can reduce the complexity of this type of call.
It's also a lot less of a pain to write out...

`const organizations = lambdaResult?.data?.records? ?? []`

bryan-robitaille (approved):
We can definitley refactor this in the future as this code will probably be a little fluid until we get Organisations fully flushed out.  For now let's get it up in place so we can continue testing.

components/containers/Admin/Organisations/OrganisationSettings.tsx
brdunfield: For now I'll just modify it to return the entire resp object if there is no status. The DeleteButton component is looking specifically for a return of exactly 200, so if anything else is passed it will just error out anyways.

We can refactor this better in the future as needed.

Issue #306, opened 2021-07-19
Triage radio button issue experienced on JAWS


srtalbot:
Research participant was unable to progress to the next screen when interacting with radio buttons on the [Public Service Award of Excellence](https://forms-staging.cdssandbox.xyz/id/1) demo form.

srtalbot:
User profile: Government-user providing feedback on accessibility considerations.

Form: https://forms-staging.cdssandbox.xyz/id/1

Description of issue:

- The participant was using JAWS to review the form, and had significant difficulty understanding what was on the page, as it was saying that radio button selections were invalid and not advancing to the next item of focus as expected.
- The grouping in the radio buttons were off; the participant was unable to hear how many items were in the list. 

Triaging: Moderate to critical, requires more investigation.

brdunfield:
I did some investigating, and used VoiceOver to compare our radio buttons to some example radio button examples on a11y documentation sites. I can definitely spot the difference, in our app it doesn't read out the grouping at all compared to the examples. Hard to describe over text, but the point is I can isolate the problem pretty easily without needing any extra or outside help.

I will try to do an effort estimation shortly as I learn more about what could be causing this!

PR #312, opened 2021-07-21
Group radio and checkboxes correctly for screen reader support

brdunfield:
# Summary | Résumé
Fix for #306 .

This PR slightly adjusts the markup in radio and checkbox groups in order to better support Screen readers.

Previously, screen readers would not identify checkbox and radio groups as "groups" for navigation purposes, but would rather just read out the label and inputs separately. This is not the correct behaviour and caused confusion in user testing.

VoiceOver now correctly identifies the fieldset element around these elements as a group. It seems like the key was using a legend instead of a label in these groups, and eliminating extraneous divs.


fitore-cds (approved):
💯

Issue #291, opened 2021-07-08
Jest is warning of deprecated methods


brdunfield:
Started seeing this today:
```
Warning: toBeInTheDOM has been deprecated and will be removed in future updates. Please use toBeInTheDocument for searching the entire document and toContainElement for searching a specific container.
```

We aren't using toBeInTheDOM, however we're using toHaveDescription. Guessing this is related. We'll need to have a look at Jest's release documentation and see if a refactor is necessary on our end.

PR #327, opened 2021-08-03
Fixed use of deprecated Jest toHaveDescription function

craigzour:
# Summary | Résumé

Fix for #291 .

Now using the official replacement for `toHaveDescription` https://github.com/testing-library/jest-dom#tohavedescription


bryan-robitaille (approved):
👍

PR #363, opened 2021-08-06
Fix/dropdown initial value validation error

### LINKED ISSUE #309: Dropdown error validation issue
Recreate issue:

- Drop down menu
- placeholder text for two options
- Select first option
- "Submit"
- Error messages shows that the field is require to be completed

craigzour:
# Summary | Résumé

Fixes Dropdown error validation issue.

closes #309


bryan-robitaille:
This raises an interesting concept if we 'should' be setting initial values for drop downs / radios / lists.  If it's a mandatory field the user should have a required action to choose a selection.  If we set the initial value so that it passes validation a user could have missed the field but yet still submit a response.
I think we'll need to dig deeper and ensure that no initial value is set for required fields as well review the logic [here] (https://github.com/cds-snc/platform-forms-client/blob/1b980c8aa6e24f55f8f9e57da4e0e3268cab4627/lib/validation.tsx#L107) as to why the shape of data for a drop down is not being captured as set even when a value is selected.

lib/tests/validation.test.js
```diff
@@ -86,30 +86,23 @@ const testCases = [
     required: true,
   },
   {
-    fieldType: "richText",
```
craigzour, line 89: @bryan-robitaille  I removed this type of validation because the `richText` component does not appear to be something the user interacts with.

lib/validation.tsx
```diff
@@ -84,37 +130,20 @@ export const validateOnSubmit = (values: FormValues, props: DynamicFormProps): R
...
 
-    const currentValidation = (currentItem.properties?.validation as ValidationProperties) || {};
-    const formikValue = values[item];
-    // A non-array object indicates an unset field (this is how Formik sends it to us)
-    // Discard these values, and convert anything else to a string
-    const currentValue: string = (
-      typeof formikValue == "object" && !Array.isArray(formikValue) ? "" : (formikValue as string)
-    ).toString();
-    const currentRegex = getRegexByType(currentValidation.type, props.t, currentValue);
+    if (formElement.properties.validation) {
```
bryan-robitaille, line 137: Nice, this really simplifies the logic

bryan-robitaille (approved):
Looks great!

Issue #354, opened 2021-08-06
Document Upload Functionality - NextJS API Submission Function


bryan-robitaille:
Linked to #280 
- NextJS API Submission Function
  
 - [x]  Handle files received through formidable and use aws/sdk to save files to S3 bucket.

- [x] Pass urls for saved files in S3 to Lambda Submission Function
 

- [X] If Lambda submission errors out delete files in S3 bucket.
- [X] Reviewed

PR #364, opened 2021-08-11
Feature/354

falila:
# Summary | Résumé

> Enable an  uploaded file to be stored in a s3 bucket and persisted via the submission lambda. 
> [Trello](https://github.com/cds-snc/platform-forms-client/issues/354)
>  Issue #354 

---

1. Upload a file.
2. Check the dev bucket tempXXX to see the file.

---

# Reviewer checklist | Liste de vérification du réviseur

This is a suggested checklist of questions reviewers might ask during their
review | Voici une suggestion de liste de vérification comprenant des questions
que les réviseurs pourraient poser pendant leur examen :

- [ ] Does this meet a user need? | Est-ce que ça répond à un besoin utilisateur?
- [ ] Is it accessible? | Est-ce que c’est accessible?
- [ ] Is it translated between both offical languages? | Est-ce dans les deux
      langues officielles?
- [ ] Have you tested it? | L’avez-vous testé?
- [ ] Are there automated tests? | Y a-t-il des tests automatisés?
- [ ] Does this cause automated test coverage to drop? | Est-ce que ça entraîne
      une baisse de la quantité de code couvert par les tests automatisés?
- [ ] Does this break existing functionality? | Est-ce que ça brise une
      fonctionnalité existante?
- [ ] Should this be split into smaller PRs to decrease change risk? | Est-ce
      que ça devrait être divisé en de plus petites demandes de tirage (« pull
      requests ») afin de réduire le risque lié aux modifications?
- [ ] Does this change the privacy policy? | Est-ce que ça entraîne une
      modification de la politique de confidentialité?
- [ ] Does this introduce any security concerns? | Est-ce que ça introduit des
      préoccupations liées à la sécurité?
- [ ] Does this significantly alter performance? | Est-ce que ça modifie de
      façon importante la performance?
- [ ] What is the risk level of using added dependencies? | Quel est le degré de
      risque d’utiliser des dépendances ajoutées?
- [ ] Should any documentation be updated as a result of this? (i.e. README
      setup, etc.) | Faudra-t-il mettre à jour la documentation à la suite de ce
      changement (fichier README, etc.)?


bryan-robitaille (changes requested):
I know there are a lot of comments but let's chat on how we can approach the modifying of the logic in `submit.tsx` for the `pushFileToS3` to descope it a bit.

lib/types.tsx
```diff
@@ -205,3 +205,22 @@ export interface CrudOrganisationResponse {
...
+
+export type SuccessData = {
+  Location: string;
+  ETag?: string;
+  Key?: string;
+};
+
+export type UploadFailure = {
+  isValid: false;
+  errorReason: any | string;
```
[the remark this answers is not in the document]
falila: fixed

pages/api/submit.tsx
```diff
@@ -181,4 +175,43 @@ const processFormData = async (
...
 
+/**
+ * Push a given file to a temporairy S3 and supply the resulting url to the submission's lambda.
+ * @param fileOrArray
+ * @param reqFields
+ * @param key
+ */
+const pushFileToS3 = async (
+  fileOrArray: formidable.File,
+  reqFields: Responses | undefined,
```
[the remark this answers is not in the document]
falila: I've removed the parameter reqFields since we are not using it anymore.

pages/api/submit.tsx
```diff
@@ -181,4 +175,43 @@ const processFormData = async (
...
+  fileName: string
+): Promise<void> => {
+  const bucketName: string = process.env.AWS_BUCKET_NAME as string;
+
+  readStream2buffer(fs.createReadStream(fileOrArray.path))
```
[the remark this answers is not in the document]
falila: i did refactor this method and moved it inside uploadFileToS3. I guess narrowing the scope.

lib/types.tsx
```diff
@@ -205,3 +205,21 @@ export interface CrudOrganisationResponse {
     ];
   };
 }
+
+// File Upload Result Type
+export type UploadSuccess = {
+  isValid: true;
```
bryan-robitaille, line 211: When defining a type or an interface it's best practice to avoid using actual values.
For example `isValue: true` would be usually typed as `isValid: boolean`

lib/types.tsx
```diff
@@ -205,3 +205,21 @@ export interface CrudOrganisationResponse {
...
+  successValue: SuccessData;
+};
+
+export type SuccessData = {
+  location: string;
+  data?: string;
+};
+
+export type UploadFailure = {
+  isValid: false;
```
bryan-robitaille, line 221: When defining a type or an interface it's best practice to avoid using actual values.
For example `isValue: true` would be usually typed as `isValid: boolean`

pages/api/s3-upload.ts
```diff
@@ -0,0 +1,72 @@
+import { ReadStream } from "fs";
+import fs from "fs";
+import { v4 as uuid } from "uuid";
+import { UploadResult, UploadFailure, UploadSuccess } from "../../lib/types";
+import { S3Client, PutObjectCommand, GetObjectCommand } from "@aws-sdk/client-s3";
+import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
+import formidable from "formidable";
+
+const s3Client = new S3Client({ region: process.env.AWS_BUCKET_REGION }); // could be hard coded
```
bryan-robitaille, line 9: If we're not going to hard code it than let's set a default in case the env var is missing.  If we were to merge this PR the functionality wouldn't actually work on Staging because we would also need to do a infrastructure release that injects an env into the fargate instances.
ex. `{region: process.env.AWS_BUCKET_REGION ?? "ca-central-1"}`

pages/api/s3-upload.ts
```diff
@@ -0,0 +1,72 @@
...
+    // setting the parameters
+    const uploadParams = {
+      Bucket: bucketName,
+      Body: data,
+      Key: `${bucketName}/user_file/${new Date().toISOString().slice(0, 10)}/${uuid()}.${filePath}`,
+    };
+
+    await s3Client.send(new PutObjectCommand(uploadParams));
+    const signedUrl = await getSignedUrl(s3Client, new GetObjectCommand(uploadParams), {
+      expiresIn: 3600 * 24 * 4,
```
bryan-robitaille, line 35: Nice! Good thinking of implementing an expiry here.

pages/api/submit.tsx
```diff
@@ -128,14 +117,27 @@ const processFormData = async (
...
       const fileOrArray = value;
       if (!Array.isArray(fileOrArray)) {
         if (fileOrArray.name) {
-          reqFields[key] = fileOrArray.name;
+          const result = await pushFileToS3(fileOrArray, fileOrArray.name);
+          if (result.isValid) reqFields[key] = result.successValue.location;
+          else throw new Error(result.errorReason);
         }
+      } else if (Array.isArray(fileOrArray)) {
+        fileOrArray.map(async (fileItem) => {
```
bryan-robitaille, line 129: I do this a lot by default as well but usually if we don't need a return value or array from the `map` call it can be switched to a `forEach` instead which returns type void.

pages/api/submit.tsx
```diff
@@ -181,4 +183,28 @@ const processFormData = async (
...
 };
 
+/**
+ * Push a given file to a temporairy S3
+ * @param fileOrArray
+ * @param reqFields
+ * @param key
+ */
+const pushFileToS3 = async (
+  fileOrArray: formidable.File,
```
bryan-robitaille, line 193: Could be renamed to just `file` to avoid confusion as through the logic tests earlier in the parent function the array type is already test for.  The type on this input parameter is also `formidable.File` so if Typescript detected that an array could be passed it would throw an error/warning.

pages/api/submit.tsx
```diff
@@ -181,4 +183,28 @@ const processFormData = async (
...
+  let uploadResult: UploadResult;
+  try {
+    uploadResult = await uploadFileToS3(fileOrArray, bucketName, fileName);
+    if (!uploadResult.isValid) {
+      throw new Error(uploadResult.errorReason);
+    }
+  } catch (error) {
+    throw new Error(error);
+  } finally {
+    fs.unlinkSync(fileOrArray.path);
```
bryan-robitaille, line 206: This is good for now.  We can open a new issue later that defines more enhanced error handing for when to delete files from S3 if there is an error later in the calling function.

.dockerignore
```diff
@@ -5,7 +5,6 @@ terraform
 .DS_Store
 .next
 .vscode
-.env
```
bryan-robitaille, line 8: `.env` should be ignored by Docker because for security we don't want secret values stored by accident in an image.

tsconfig.json
```diff
@@ -20,6 +20,6 @@
       "@components/*": ["components/*"]
     }
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "lib/tests/s3-upload.test.js"],
```
bryan-robitaille, line 23: This was probably an auto addition by typescript module of VSCode. 'js' files don't need to be transcompiled by the Typescript engine.

lib/tests/s3-upload.test.js
```diff
@@ -0,0 +1,37 @@
...
+      expect(error).toBeTruthy();
+    });
+  });
+});
+
+describe("uploadFileToS3", () => {
+  it("Should call upload method with incorrect file path", async () => {
+    jest.spyOn(global, "Date").mockImplementationOnce(() => new Date("2021-08-12"));
+
+    await uploadFileToS3({ path: "./test" }, "temp-s3-upload-testing", "test").then((data) => {
```
bryan-robitaille, line 23: Modified the test to check for the bad file path.  Passing in a Buffer object was throwing an error because `path` did not exist on the Buffer object.

bryan-robitaille (changes requested):
Nearly there.  Let's clean up some of the commented out code and one last quick refactor for renaming variables and this should be ready to merge :)

pages/api/s3-upload.ts
falila: oaky. Lets for now hard code it. And we'll refactor it once the infra is ready.

lib/types.tsx
falila: Will address this one in a coming PR.

bryan-robitaille (approved):
Made a few small changes to simplify the types that were created as well as implementing some of the changes that I had requested but were not completed.

PR #371, opened 2021-08-17
Feature/354

falila:


pages/api/submit.tsx
```diff
@@ -117,28 +117,29 @@ const processFormData = async (
...
+        fileOrArray.forEach(async (fileItem, index) => {
           if (fileItem.name) {
-            const { isValid, result } = await pushFileToS3(fileItem, fileItem.name);
-            //TODO @bryan-robitaille.
-            if (isValid)
-              /** what's the best approach to store these links in response object*/
-              // Need to check to see if the below is easy to parse out when processing the form submission
-              reqFields[key] = result;
-            else throw new Error(result);
+            logMessage.info(`uploading: ${_key} - filename ${fileItem.name} `);
```
bryan-robitaille, line 135: changed console.log() to logMessage.info() so that messages only appear in staging

pages/api/submit.tsx
```diff
@@ -117,28 +117,29 @@ const processFormData = async (
...
 
-    for (const [key, value] of Object.entries(files)) {
+    for (const [_key, value] of Object.entries(files)) {
       const fileOrArray = value;
       if (!Array.isArray(fileOrArray)) {
         if (fileOrArray.name) {
-          const { isValid, result } = await pushFileToS3(fileOrArray, fileOrArray.name);
-          if (isValid) reqFields[key] = result;
-          else throw new Error(result as string);
+          logMessage.info(`uploading: ${_key} - filename ${fileOrArray.name} `);
```
bryan-robitaille, line 124: changed console.log() to logMessage.info() so that messages only appear in staging

pages/api/submit.tsx
```diff
@@ -117,28 +117,29 @@ const processFormData = async (
...
+          const { isValid, key } = await pushFileToS3(fileOrArray);
+          if (isValid) {
+            uploadedFilesKeyUrlMapping.set(fileOrArray.name, key);
+            reqFields[_key] = key;
```
bryan-robitaille, line 128: Added return value (file path) to the reqForms form field property otherwise the value would never be passed to the submission lambda function.

pages/api/submit.tsx
```diff
@@ -117,28 +117,29 @@ const processFormData = async (
...
-            if (isValid)
-              /** what's the best approach to store these links in response object*/
-              // Need to check to see if the below is easy to parse out when processing the form submission
-              reqFields[key] = result;
-            else throw new Error(result);
+            logMessage.info(`uploading: ${_key} - filename ${fileItem.name} `);
+            const { isValid, key } = await pushFileToS3(fileItem);
+            if (isValid) {
+              uploadedFilesKeyUrlMapping.set(fileItem.name, key);
+              reqFields[`${_key}-${index}`] = key;
```
bryan-robitaille, line 139: Added return value (file path) to the reqForms form field property otherwise the value would never be passed to the submission lambda function.

lib/s3-upload.ts
```diff
@@ -69,4 +66,34 @@ const readStream2buffer = (fileStream: ReadStream): Promise<Buffer> => {
...
 };
 
-export { uploadFileToS3, readStream2buffer };
+/**
+ * Push a given file to a temporary S3
+ * @param fileOrArray
+ * @param reqFields
+ * @param key
+ */
+const pushFileToS3 = async (file: formidable.File): Promise<UploadResult> => {
```
bryan-robitaille, line 75: Refactored and moved to lib/s3-upload

lib/s3-upload.ts
```diff
@@ -2,46 +2,43 @@ import { ReadStream } from "fs";
...
-import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
+import { logMessage } from "./logger";
+import { S3Client, PutObjectCommand, DeleteObjectCommand } from "@aws-sdk/client-s3";
+
 import formidable from "formidable";
 
-const s3Client = new S3Client({ region: "ca-central-1" });
+const s3Client = new S3Client({ region: process.env.AWS_REGION ?? "ca-central-1" });
+
+const bucketName: string =
```
bryan-robitaille, line 12: Added check env name for reliability s3 bucket and fallback

lib/s3-upload.ts
```diff
@@ -2,46 +2,43 @@ import { ReadStream } from "fs";
...
  */
-const uploadFileToS3 = async (
-  file: formidable.File,
-  bucketName: string,
-  filePath: string
-): Promise<UploadResult> => {
+const uploadFileToS3 = async (file: formidable.File): Promise<UploadResult> => {
   try {
     const data = await readStream2buffer(fs.createReadStream(file.path));
+    const objectKey = `form_attachments/${new Date().toISOString().slice(0, 10)}/${uuid()}.${
```
bryan-robitaille, line 24: refactored to simplify path name by removing bucketname.
Also removed fetching the signed url.  The signed url could be used to gain unauthorized access to the file as it provides a url with credentials.  The GetObject command for s3 only requires the Key which we will use in the Lambda's with existing credentials.  I should have provided more detail in the beginning around the architecture for the solution.

lib/s3-upload.ts
```diff
@@ -2,46 +2,43 @@ import { ReadStream } from "fs";
...
- * @param bucketName
- * @param filePath
  * @returns
  */
-const uploadFileToS3 = async (
-  file: formidable.File,
-  bucketName: string,
-  filePath: string
-): Promise<UploadResult> => {
+const uploadFileToS3 = async (file: formidable.File): Promise<UploadResult> => {
```
bryan-robitaille, line 21: refactored to use only the file as a parameter.  `filePath` parameter with the file name which is already included on the `file` object.

bryan-robitaille (approved):
Tested with the refactoring and the file paths will be passed to the submission Lambda.  Ready to merge to main when you have a chance.

PR #421, opened 2021-09-09
Remove ML API call from app

bryan-robitaille:
# Summary | Résumé
Removes the Mailing List API call from the front end app.


Moro-Code (approved):
LGTM

PR #442, opened 2021-09-24
Fix/dynamicrow

### LINKED ISSUE #413: Error validation not working on Copyright Board of Canada form

falila:
# Summary | Résumé
## updated
New implementation of dynamic row . 

closes #413 

Known issue :  * When a validation occurs on a field in dynamic row's elements. we lost the red colour around  it.

> Form: https://forms-staging.cdssandbox.xyz/id/76

Without filling in any of the required fields I should not be able to submit the form.

- [ ] Does this meet a user need? | Est-ce que ça répond à un besoin utilisateur?
- [ ] Is it accessible? | Est-ce que c’est accessible?
- [ ] Is it translated between both offical languages? | Est-ce dans les deux
      langues officielles?
- [X] Have you tested it? | L’avez-vous testé?
- [ X] Are there automated tests? | Y a-t-il des tests automatisés?
- [ ] Does this cause automated test coverage to drop? | Est-ce que ça entraîne
      une baisse de la quantité de code couvert par les tests automatisés?
- [ ] Does this break existing functionality? | Est-ce que ça brise une
      fonctionnalité existante?
- [ ] Should this be split into smaller PRs to decrease change risk? | Est-ce
      que ça devrait être divisé en de plus petites demandes de tirage (« pull
      requests ») afin de réduire le risque lié aux modifications?
- [ ] Does this change the privacy policy? | Est-ce que ça entraîne une
      modification de la politique de confidentialité?
- [ ] Does this introduce any security concerns? | Est-ce que ça introduit des
      préoccupations liées à la sécurité?
- [ ] Does this significantly alter performance? | Est-ce que ça modifie de
      façon importante la performance?
- [ ] What is the risk level of using added dependencies? | Quel est le degré de
      risque d’utiliser des dépendances ajoutées?
- [ ] Should any documentation be updated as a result of this? (i.e. README
      setup, etc.) | Faudra-t-il mettre à jour la documentation à la suite de ce
      changement (fichier README, etc.)?


Moro-Code (changes requested):
Hi Raph

I realize this is only a draft but I'm going to give a detailed review on what I expect in order to save you time when finishing up this PR. 

A couple of general comments.

- Please refine your comments and be clear and detailed as to your intention of the code you're trying to explain when writing them. This will help the reviewer immensely understand what you're thinking and where the your headspace is at. 
- The simpler the code the better. Always optimize your code for readability
- Please be consistent in naming variables i.e. `elem` or `element`. These simple things do make a difference. Consistent and accurate variable naming means less cognitive load for the individual reading your code

With that last point in mind I believe what you're trying to do can be achieved in a simpler fashion. I will write some code snippets to give you an idea of what I'm thinking. Please not however I have not tested this code but its merely to give you an idea 

validateOnSubmit function
```ts
/**
 * validateOnSubmit is called during Formik's submission event to validate the fields
 * @param values
 * @param props
 */
 export const validateOnSubmit = (values: FormValues, props: DynamicFormProps, errors: Responses = {} ): Responses => {
   for ( const item in values ) {
     // get the form element from the formConfig elements array using the find method
     const formElement = props.formConfig.elements.find((element) => element.id === parseInt(item));

    if(!formElement) return errors;
    
   // if a form element can be validated then validate the value and if there are
  // any errors add them to the results object
    if(formElement.properties.validation){
     const result = isFieldResponseValid(
          values[item],
          formElement.type
          formElement.properties.validation,
          formElement,
          props.t
      );
     if( results  && typeof result === "object") {
          errors[item] = {
             ...errors,
             ...results
          }
     }else if (results){
        // it is a string in this case 
        errors[item] = result
     }
    }
  }
}
```

isFieldResponseValid function
```ts
const isFieldResponseValid = (
  value: unknown,
  componentType: string,
  formElement: FormElement,
  validator: ValidationProperties,
  t: TFunction
): string | null | Results => {
  switch (componentType) {
    case "textField": {
      const typedValue = value as string;
      if (validator.required && !typedValue) return t("input-validation.required");
      const currentRegex = getRegexByType(validator.type, t, value as string);
      if (validator.type && currentRegex && currentRegex.regex) {
        // Check for different types of fields, email, date, number, custom etc
        const regex = new RegExp(currentRegex.regex);
        if (typedValue && !regex.test(typedValue)) {
          return currentRegex.error;
        }
      }
      break;
    }
    case "textArea": {
      const typedValue = value as string;
      if (validator.required && !typedValue) return t("input-validation.required");
      break;
    }
    case "checkbox":
    case "radio":
    case "dropdown": {
      if (validator.required && value === undefined) return t("input-validation.required");
      break;
    }
    case "fileInput": {
      //TODO need refactoring.
      const typedValue = value as { file: File; src: FileReader; name: string; size: number };
      if ((validator.required && typedValue === undefined) || typedValue === null)
        return t("input-validation.required");
      if (validator.required && typedValue && typedValue.file === null)
        return t("input-validation.required");
      // Size limit is 8 MB
      if (typedValue && typedValue.size > 8000000) return t("input-validation.file-size-too-large");
      if (
        typedValue &&
        typedValue.file &&
        acceptedFileMimeTypes.split(",").find((value) => value === typedValue.file.type) ===
          undefined
      ) {
        return t("input-validation.file-type-invalid");
      }
      break;
    }
    case "dynamicRow":
        // in the case of the dynamicRow we want to validate the subElements 
        if( formElement.properties.subElements){
         // get the subElements from the dynamic row element
          const subElements = formElement.properties.subElements
         // set up object to store results
          const errors: Results = {}
          for ( const index in subElements ) {
            const subElement = subElements[index]
            const subElementId = extractSubElementId(elem)
            if( subElementId && subElement.properties.validation) {
             // I assume you still have to build in looping over rows Raph 
              const rowValues = value["0"]
             // get the results for the subElement
              const results = isFieldResponseValid(
                rowValues[subElementId],
                subElement.type,
                subElement.properties.validation,
                subElement,
                props.t
              )
             
             if(validatorResults){
                results[elem.subId] = result as string
             }
           }
          }
          return errors
        }
    case "richText":
      break;
    default:
      throw `Validation for component ${componentType} is not handled`;
  }
  return null;
};
```
Again the code above may not be correct but it should give you a guideline. I will probably have much more comments when you move your PR from draft mode but this should suffice for now and I would also like to get to my professional development 😅

lib/tests/dynamicRow.validation.test.js
```diff
@@ -0,0 +1,85 @@
+import { extractSubElementId } from "../validation";
```
Moro-Code, line 1: Is there  any particular reason a separate test file exists. There is already a validation test file

lib/tests/dynamicRow.validation.test.js
```diff
@@ -0,0 +1,85 @@
...
+        size: 0,
+      },
+      9: "",
+    },
+  ],
+};
+
+describe("Dynamic rows validator", () => {
+  test("Parses form values data", () => {
+    expect(mockDynaRowFormRes.length === 3);
```
Moro-Code, line 53: What is this testing exactly ?

lib/tests/dynamicRow.validation.test.js
```diff
@@ -0,0 +1,85 @@
...
+  });
+});
+
+describe("extractSubElementId: dynamic rows sub's ids validator", () => {
+  test("Should extract element id ", () => {
+    const elem = dynForm.formConfig.elements.find((element) => element.id === 3);
+    expect(
+      extractSubElementId(elem.properties.subElements.find((subElem) => subElem.id === 4)) === "0"
```
Moro-Code, line 61: Why not pass in the data you expect directly rather than doing all these find operations

lib/tests/dynamicRow.validation.test.js
```diff
@@ -0,0 +1,85 @@
...
+    );
+  });
+
+  test("Should have a property type subId", () => {
+    const elem = dynForm.formConfig.elements.find((element) => element.id === 3);
+    expect(elem.properties.subElements.find((subElem) => subElem.id === 4)).toHaveProperty("subId");
```
Moro-Code, line 67: There is no need to write tests for test data. Tests should target production code

lib/tests/dynamicRow.validation.test.js
Moro-Code: Do not use boolean operators in expect statements. The syntax here should be `expect(...).toEqual("0")`

lib/tests/dynamicRow.validation.test.js
```diff
@@ -0,0 +1,85 @@
...
+    expect(extractSubElementId(subElem) === 1);
+  });
+
+  test("Should thrown an exception invalid formElement", () => {
+    expect(extractSubElementId).toThrow("Invalid formElement");
+  });
+
+  test("Should thrown an exception invalid id", () => {
+    const elem = dynForm.formConfig.elements.find((element) => element.id === 12);
+    expect(() => extractSubElementId(elem)).toThrow();
```
Moro-Code, line 83: What is this testing ?

lib/tests/dynamicRow.validation.test.js
```diff
@@ -0,0 +1,85 @@
...
+
+  test("Should return a valid sub element id 1", () => {
+    const elem = dynForm.formConfig.elements.find((element) => element.id === 7);
+    const subElem = elem.properties.subElements.find((subElem) => subElem.id === 9);
+    expect(subElem).toHaveProperty("subId");
+    expect(extractSubElementId(subElem) === 1);
+  });
+
+  test("Should thrown an exception invalid formElement", () => {
+    expect(extractSubElementId).toThrow("Invalid formElement");
```
Moro-Code, line 78: Function isn't called here ?

lib/tests/formBuilder.test.js
```diff
@@ -2,7 +2,7 @@ import React from "react";
 import { cleanup, render, screen } from "@testing-library/react";
 import { Form } from "../../components/forms";
 import { getProperty, getFormInitialValues, getRenderedForm } from "../formBuilder";
-import TestForm from "./testData";
+import { testForm } from "./testData";
```
Moro-Code, line 5: Is their any particular reason for this alteration ?

lib/tests/dataLayer.test.js
```diff
@@ -10,7 +10,7 @@ import {
   getFormByStatus,
 } from "../dataLayer";
 import { logMessage } from "../logger";
-import testForm from "./testData";
+import { testForm } from "./testData";
```
Moro-Code, line 13: Why was this altered ?

lib/validation.tsx
```diff
@@ -106,26 +108,61 @@ const isFieldResponseValid = (
...
+      //TODO need refactoring.
       const typedValue = value as { file: File; src: FileReader; name: string; size: number };
-      if (validator.required && typedValue.file === null) return t("input-validation.required");
+      if ((validator.required && typedValue === undefined) || typedValue === null)
+        return t("input-validation.required");
+      if (validator.required && typedValue && typedValue.file === null)
+        return t("input-validation.required");
       // Size limit is 8 MB
-      if (typedValue.size > 8000000) return t("input-validation.file-size-too-large");
+      if (typedValue && typedValue.size > 8000000) return t("input-validation.file-size-too-large");
```
Moro-Code, line 118: Please limit the scope of your PR to the issue you are solving. If you are touching on other code in your PR give the team a heads up so as to avoid merge conflicts

lib/validation.tsx
```diff
@@ -106,26 +108,61 @@ const isFieldResponseValid = (
...
+        const subElements = formElement.properties.subElements.filter(
+          (element) => element.type !== "richText"
+        );
+        // set up object to store results
+        const errors: Responses = {};
+        for (const index in subElements) {
+          const subElement = subElements[index];
+          const subElementId = extractSubElementId(subElement);
+          if (subElement.properties.validation && value) {
+            const temValue = value as any;
```
Moro-Code, line 142: This should not be type any... We know what the types can be this should be a union type of these values. Or you can use `unknown`

Moro-Code, line 142: why another variable here

lib/validation.tsx
```diff
@@ -106,26 +108,61 @@ const isFieldResponseValid = (
...
+            const subElementValues = temValue["0"][parseInt(subElementId)];
```
Moro-Code, line 143: Does this mean we're only validating for the first row only ?

Moro-Code (changes requested):
Hey Raph... Some changes and context needed here. I also don't see any tests for the dynamic row validation. I would like to see unit tests and also end to end tests confirming that this works.

lib/tests/formBuilder.test.js
falila: "TestForm" was exported as default from the module. I added another variable then i change them in lowercase to make  the export consistent with the variable name.

lib/validation.tsx
falila: This was an issue that i found went uploading a file in a dynamic Row. And i added a fix. i also had a chat with Clement so that he was aware of it. I will try not  next time.

lib/validation.tsx
falila: No , each element (dynamic row)   has a  array of size 1 that contain all its values.  So this is needed in order to get the subsequent  value or object .

lib/tests/dynamicRow.validation.test.js
falila: throw "Invalid id";

lib/tests/dynamicRow.validation.test.js
falila: this syntax throw an error like "toEqual is not a function" etc ... when testing  reason why.

lib/tests/dynamicRow.validation.test.js
falila: Dynamic row will get more complicated as time goes so i though it might be good idea to separate. But yes it can be merged into it.

lib/validation.tsx
Moro-Code: All good thanks for adding context. Could you post in forms-dev next time so we're all aware.

lib/tests/dynamicRow.validation.test.js
Moro-Code: Then you would just need to do the following `expect(extractSubElementId(elem)).toThrow()`. Why the extra arrow function

falila:
@Moro-Code i will add  test/coverage for dynamic row validation. I had to re-write the tests since the implementation has changed.

lib/tests/dynamicRow.validation.test.js
Moro-Code: Here is a list of matchers jest supports https://jestjs.io/docs/expect

lib/tests/formBuilder.test.js
Moro-Code: You could still have a default export while exporting other variables.

lib/validation.tsx
Moro-Code: @falila 

When I look at the data by consol logging the value that gets passed in to the isFieldResponseValid that does not seem to be the case. In this case here you would only get the first row and validate the subElements only for that row 

Screen Shot 2021-09-27 at 4 13 05 PM

lib/validation.tsx
```diff
@@ -106,26 +108,59 @@ const isFieldResponseValid = (
...
+      //TODO need refactoring.
       const typedValue = value as { file: File; src: FileReader; name: string; size: number };
-      if (validator.required && typedValue.file === null) return t("input-validation.required");
+      if ((validator.required && typedValue === undefined) || typedValue === null)
+        return t("input-validation.required");
+      if (validator.required && typedValue && typedValue.file === null)
+        return t("input-validation.required");
       // Size limit is 8 MB
-      if (typedValue.size > 8000000) return t("input-validation.file-size-too-large");
+      if (typedValue && typedValue.size > 8000000) return t("input-validation.file-size-too-large");
```
bryan-robitaille, line 118: There's a nice shorthand that exists now with ES2020 that allows for Optional Chaining.  [Reference Doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

It allows for expressions like `if( typedValue && typedValue.size)` to be rewritten as `if(typedValue?.size)` which greatly reduces the overhead and cognitive overload when writing these types of boolean logic checks.

lib/validation.tsx
```diff
@@ -106,26 +108,59 @@ const isFieldResponseValid = (
...
-    case "dynamicRow": {
-      // Not implemented yet
+    case "dynamicRow":
+      // in the case of the dynamicRow we want to validate the subElements
+      if (formElement.properties.subElements) {
+        // get the subElements from the dynamic row element
+        const subElements = formElement.properties.subElements;
+        // set up object to store results
+        const errors: Responses = {};
+        for (const index in subElements) {
```
bryan-robitaille, line 136: There is also a for... of that can be used on Arrays that returns the value at the index instead of the index in the Array.  This would save from having to declare an additional variable.
For example:
```js
for (const index in subElements){
  const subElement = subElements[index];
  const subElementId = extractSubElementId(subElement);
```
can be easily rewritten to use a for ... of:
```js
for ( const subElement of SubElements){
  const subElementId = extractSubElementId(subElement);
```

falila: Great point!

lib/tests/dynamicRow.validation.test.js
falila: I tested that an error was thrown while the function gets executed.
Note: toThrow executes a function passed in expect and verifies that it throws an error.

lib/tests/dynamicRow.validation.test.js
falila: We should wrap the resolve function call in another function that is passed to expect.
see the doc https://jasmine.github.io/2.0/introduction.html.

lib/validation.tsx
```diff
@@ -106,26 +108,61 @@ const isFieldResponseValid = (
...
       }
       break;
     }
-    case "dynamicRow": {
-      // Not implemented yet
+    case "dynamicRow":
+      // in the case of the dynamicRow we want to validate the subElements
+      if (formElement.properties.subElements) {
+        // get the subElements from the dynamic row element
+        const subElements = formElement.properties.subElements.filter(
```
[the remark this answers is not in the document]
falila: Please refer to my comment.

lib/tests/dynamicRow.validation.test.js
falila: drop

lib/validation.tsx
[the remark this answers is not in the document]
falila: Please see comments and the draft PR summary.

tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "typeRoots": ["./types", "./node_modules/@types"],
     "target": "es5",
+    "downlevelIteration": true,
```
falila, line 5: Why do we need to "downlevel" ?

falila (commented):
Good work.

tsconfig.json
Moro-Code: Needed to allow looping iterator objects with es5 rather than just arrays

bryan-robitaille: https://www.typescriptlang.org/tsconfig#downlevelIteration

bryan-robitaille (approved):
Looks good.
While testing I found a new bug that's outside of the scope of this PR as it already exists in the main branch.
In a nutshell when the page renders after a validation check, and errors are present, the additional rows for a dynamic row do not render.  When clicking 'add a row' the values are not lost and are rendered in place.

Issue #355, opened 2021-08-06
Document Upload Function - Lambdas


bryan-robitaille:
Linked to #280 
- Lambda Submission Function
  - Little change should be necessary...
  - Ensure file urls are passed through to DynamoDB formData
- Reliability Queue Lambda
  - Little changes in main code body
  - Ensure file urls are passed through to Notify and Vault processing functions
- Notify processing function (Reliability Lambda lib)
  - Read up on Notify's file attachment API structure: [Notify API](https://documentation.notification.canada.ca/en/send.html#sending-a-file-by-email)
  - Fetch appropriate files from S3 storage and process into type required by Notify (base64 encoded string)
  - On successful submission to Notify delete files from S3.
- Vault processing function (Reliability Lambda lib)
  - Read files from S3 and move to Vault S3 storage
  - Delete files from temporary storage

craigzour:
Reopening because even if the PR has been merged we are still testing and debugging live in staging

PR #445, opened 2021-09-28
Feature/handle file input in dynamic row components

craigzour:
# Summary | Résumé

- Modified S3 uploaded file access path generation
  - Before: form_attachments/2021-09-28/0e0e0429-08b1-4d6e-894c-856d7b2cda4d.hotdogs.pdf
  - After: form_attachments/2021-09-28/0e0e0429-08b1-4d6e-894c-856d7b2cda4d/hotdogs.pdf

It will allow us to decompose the path on the AWS lambda side and easily get the filename which will be the last path component.

- Added support for file input components within dynamic row ones

When we input a file using the appropriate component within a dynamic row one we are not pushing that file to S3 and thus not processing it in the Reliability lambda.

---

Part of the work related to https://github.com/cds-snc/platform-forms-client/issues/355

This is not easy to test as it is pretty much transparent for the end user. We will need to see how it behaves in staging and we will have to tweak some of the lambda logic as well.


craigzour:
Currently reviewing my own code and testing different types of forms to make sure I did not break anything.

lib/tests/dataLayer.test.js
```diff
@@ -134,28 +149,48 @@ describe("Extract Form data", () => {
   });
 
   test("Build form data object", () => {
+    const expectedOutput = {
+      1: "test response",
+      2: "test response 2",
+      4: "Manitoba",
+      5: "Student",
+      6: '{"value":[]}',
+      [`7-0-0`]: "Doe",
```
Moro-Code, line 158: Minot nitpick: These could just be strings rather than template strings

lib/dataLayer.tsx
```diff
@@ -421,53 +475,87 @@ function _buildFormDataObject(form: PublicFormSchemaProperties, values: Response
...
+): [string, string | Blob][] {
+  if (element.type === "dynamicRow") {
+    if (element.properties.subElements === undefined) return [];
+
+    const subElements = element.properties.subElements.filter(
+      (element) => !["richText"].includes(element.type)
+    );
+
+    const responses = value as Responses[];
+    return responses
```
Moro-Code, line 501: perhaps add a comment here about whats happening so its clear

Moro-Code (changes requested):
Great work and the tests look good! There are some minor comments I had. In general would be great to add more comments especially around more complex pieces of code to give context to the reviewer.

lib/dataLayer.tsx
```diff
@@ -294,27 +294,80 @@ async function _getSubmissionByID(formID: string): Promise<SubmissionProperties
...
+    });
+
+  return rehydratedResponses;
+}
+
+function _rehydrateDynamicRowResponses(responses: [string, Response][]) {
+  const rehydratedResponses: Responses[] = [];
+
+  let currentResponse: Responses = {};
+  let currentResponseIndex: string | undefined = undefined;
```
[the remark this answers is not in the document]
craigzour: Doing this would mean that the `foreach` loop just below starts at index 1 instead of 0 since we would have already taken that first element into account. In the end it would just make the code a bit more cluttered.

lib/dataLayer.tsx
```diff
@@ -421,53 +475,87 @@ function _buildFormDataObject(form: PublicFormSchemaProperties, values: Response
...
+    );
+
+    const responses = value as Responses[];
+    return responses
+      .map((response, responseIndex) => {
+        return subElements
+          .map((subElement, subElementIndex) => {
+            const result = _handleFormDataType(subElement, response[subElementIndex]);
+            return result
+              ? ([[`${element.id}-${responseIndex}-${subElementIndex}`, result[1]]] as [
```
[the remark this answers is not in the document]
craigzour: In fact this is a tuple inside an array and I used it as a trick to avoid having to deal with the undefined response that can be returned from the `_handleFormDataType ` function. It allows me to return an empty array if the response is undefined and then it perfectly slides in the `flat` logic at the end which is needed because we are using a `map` in a `map` and we do not want to get an array of arrays.

lib/dataLayer.tsx
```diff
@@ -421,53 +475,99 @@ function _buildFormDataObject(form: PublicFormSchemaProperties, values: Response
   const formData = new FormData();
   form.elements = form.elements.filter((element) => !["richText"].includes(element.type));
```
bryan-robitaille, line 476: We need to be careful when we modify a passed in object to a function.  This does modify the object higher up in the scope.  In this case it flows all the way back up to `formikBag.props`.  If we're not sure if modifying the object higher in the scope will cause issues it's aways better to clone it and then do manipulations.
In the scope of this function it seems like we're only using `form.elements` so we could just create a const for the manipualted array and leverage it in the scope wihtin this function.

lib/dataLayer.tsx
```diff
@@ -421,53 +475,99 @@ function _buildFormDataObject(form: PublicFormSchemaProperties, values: Response
...
+
+    /**
+     * We are creating a new data structure (to be passed to the submit API) from the multiple responses that could have been entered
+     * for the dynamic row component we are currently processing. */
+    return (
+      responses
+        .map((response, responseIndex) => {
+          return subElements
+            .map((subElement, subElementIndex) => {
+              const result = _handleFormDataType(subElement, response[subElementIndex]);
```
bryan-robitaille, line 511: The _handleFormDataType is receiving the wrong response for the subElement.  The response property for the subElement is the index of the subElement in the unmodified form config array.  By removing the richtext element the array has been shifted and subElement with ID 1 is being passed in response[0].
This could be rewritten like: 
```js
const result = _handleFormDataType(subElement, response[subElement.id]);
```

bryan-robitaille (changes requested):
Unfortunatley when I tested the code with form 79 which includes a file input and dropdown in the dynamic row it was not working.  I debugged and found the issue.  

I also included some general code cleanup outside of the scope of this PR that I ran across while doing the review.

lib/dataLayer.tsx
craigzour: Discovered another issue with your approach but I found a new way to fix it :)

PR #447, opened 2021-09-29
Feature/form templates json validation

jeberhardt:
# Summary | Résumé

This will validate the JSON that is being uploaded for a form.

Launch application and upload a form. You can test for a form failing by removing any of the top-level elements: "form", "submission", "internalTitleEn", "internalTitleFr", "publishingStatus". The form will not be uploaded, and an error will appear in the Javascript console in the browser.

>> The schema will need to be updated to be more accurate, describing the required / optional properties, and the data structure.
>> The error will need to be presented properly to the user in the UI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Harmonized RichText and Checkbox/Radio label max width depending on screen ratio.
+
 ## [1.0.2] 2021-09-21
 
 ### Fixed
--- a/components/forms/Checkbox/Checkbox.tsx
+++ b/components/forms/Checkbox/Checkbox.tsx
@@ -1,14 +1,22 @@
 import React from "react";
 import { Field } from "formik";
 import { MultipleChoiceProps } from "../../../lib/types";
+import { Description } from "../Description/Description";
 
 export const Checkbox = (
   props: MultipleChoiceProps & JSX.IntrinsicElements["input"]
 ): React.ReactElement => {
-  const { id, label, required, name } = props;
+  const { id, label, required, name, ariaDescribedBy } = props;
 
   return (
     <div className="gc-input-checkbox">
+      {ariaDescribedBy ? (
+        <Description id={id} className="gc-form-group-context">
+          {ariaDescribedBy}
+        </Description>
+      ) : (
+        ""
+      )}
       <Field
         className="gc-input-checkbox__input"
         id={id}
--- a/components/forms/FormGroup/FormGroup.test.js
+++ b/components/forms/FormGroup/FormGroup.test.js
@@ -42,15 +42,22 @@ describe.each([["en"], ["fr"]])("Generate a form group", (lang) => {
           ? radioButtonData.properties.descriptionEn
           : radioButtonData.properties.descriptionFr;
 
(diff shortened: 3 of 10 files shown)


Moro-Code (changes requested):
Code looks fantastic ! Awesome work but we need unit tests on the middleware you created at a minimum

craigzour (approved):

tests/templates.test.js
```diff
@@ -0,0 +1,118 @@
...
+
+jest.mock("next-auth/client");
+
+jest.mock("@aws-sdk/client-lambda", () => {
+  return {
+    LambdaClient: jest.fn(() => {
+      return {
+        send: jest.fn((command) => {
+          const encoder = new TextEncoder(),
+            decoder = new TextDecoder();
```
Moro-Code, line 20: If these are going to be on separate lines anyways. Might as well declare them in separate lines. I was a little confused reading this

tests/templates.test.js
```diff
@@ -0,0 +1,118 @@
...
+        Origin: "http://localhost:3000",
+      },
+      body: JSON.stringify({
+        formConfig: brokenFormTemplate,
+        method: "INSERT",
+      }),
+    });
+
+    await templates(req, res);
+    expect(res.statusCode).toBe(400);
```
Moro-Code, line 116: Something else to check is if you're getting the error messages you expect in the response

tests/templates.test.js
```diff
@@ -0,0 +1,118 @@
...
+        }),
+      };
+    }),
+    InvokeCommand: jest.fn((payload) => {
+      return payload;
+    }),
+  };
+});
+
+describe("Test JSON validation scenarios", () => {
```
Moro-Code, line 64: Overall these tests look good ! Well done ! There are some additional test cases like for example the lambda client throwing an error but this is outside the scope of this PR

lib/jsonValidator/jsonValidator.ts
```diff
@@ -0,0 +1,31 @@
+import { Schema, Validator } from "jsonschema";
+import { NextApiRequest, NextApiResponse } from "next";
+
+export type validateOptions = {
+  jsonKey: string;
+};
+
+const validate = (
```
Moro-Code, line 8: I still need unit tests for this. You wrote unit tests for the template handler but this doesn't provide adequate coverage for this file. We need unit tests specifically for the validate middleware

lib/http/allowed.ts
```diff
@@ -0,0 +1,37 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import { getSession } from "next-auth/client";
+
+const isRequestAllowed = (
```
Moro-Code, line 4: I still need unit tests for this. You wrote unit tests for the template handler but this doesn't provide adequate coverage for this file. We need unit tests specifically for the isRequestAllowed middleware

lib/jsonValidator/jsonValidator.ts
```diff
@@ -0,0 +1,31 @@
+import { Schema, Validator } from "jsonschema";
+import { NextApiRequest, NextApiResponse } from "next";
+
+export type validateOptions = {
```
Moro-Code, line 4: This should be capitalized if its a type

Moro-Code (changes requested):
Great work adding tests for the API handler itself. This will certainly help the team to increase test coverage of the code base. You're still missing unit tests for the individual middleware functions. I also left some minor comments

falila (approved):
Great work @jeberhardt. 
A small detail which maybe out of the scope of this PR. See pic. 
IMO , The red circle's content should be removed whenever there is an error otherwise  users might be confused. 
Screen Shot 2021-10-18 at 1 38 21 PM

PR #460, opened 2021-10-06
Harmonized RichText and Checkbox/Radio label max width depending on screen ratio

### LINKED ISSUE #414: Richtext component width is different from checkbox

craigzour:
# Summary | Résumé

- Harmonized RichText and Checkbox/Radio label max width depending on screen ratio

---

closes #414 

- Using form http://localhost:3000/en/id/85 you can see that the width of RichText and Checkbox/Radio label is harmonized when resizing the browser window


Moro-Code:
So that this does not get stale. I propose we create a new issue for this and address it separately https://github.com/cds-snc/platform-forms-client/issues/479

PR #470, opened 2021-10-13
Delete Dynamic Rows

jeberhardt:
# Summary | Résumé

Remove "Item #" from dynamic row items.

- Open a form that uses dynamic rows. (For example, Form 76).
- Add a row for "Representative Information"
- Note that **Item #x** no longer appears, and all other functionality remains the same.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


components/forms/DynamicRow/DynamicRow.test.js
```diff
@@ -71,7 +71,7 @@ const formConfig = {
 describe.each([
   ["en", "Add Row", "Item"],
```
Moro-Code, line 72: remove Item and Article from the array

Moro-Code (changes requested):
Minot comment. Good work

PR #489, opened 2021-10-20
fix: restore dynamicRows's state when errors occur

falila:
# Summary | Résumé
Closes https://github.com/cds-snc/forms-staging-terraform/issues/76

Whatever the validation errors are triggered we must ensure that all additional rows and values are reinstated. 

---

1. Load any form with dynamic rows group. (copy form staging  https://forms-staging.cdssandbox.xyz/en/id/68) 
2. Add a few rows with "add Row" button.
3. Then submit the form
4. It should be rendered with all the components and errors that are linked to them.
5. Check that rows and values are in sync.


Moro-Code:
This PR has no changes... created from #512

PR #510, opened 2021-10-25
Refactor datalayer code

falila:
# Summary | Résumé
PR is aiming to decouple data access (crud/lambdas) functions call from the client side.
1. In this iteration, all Crud functions were moved into a separate file and the common func grouped until we merge template and retrieval. This way, the following PRs are much easier to review.
2. The updated PR will focus on splitting shared (client/server) functions into small portions that are either going to be used from client or server side.


Moro-Code (approved):
LGTM

PR #525, opened 2021-11-10
Create secure API to retrieve bearer token for a specific form

### LINKED ISSUE #484: Create secure API to retrieve bearer token for a specific form
As an admin user I would like to be able to retrieve a bearer token for a specific form securely so that I may use it to authenticate against GC Form services

Acceptance Criteria:
- [x] An API route in NextJS that accepts GET request in which the path contains the ID for which form the bearer token should be retrieved from. 
- [x] The path should be `pages/api/id/[form]/bearer.js`
- [x] Should only allow requests that have a valid user session from next auth. For any invalid or non existing sessions a 403 status code should be returned 
- [x] Should a record exist with the specified Form ID and the session is valid a 200 response should be returned with the bearer token in the body of the response
- [x] Should a record not exist with the specified Form ID then a 404 response should be returned 
- [x] Adequate unit testing on all levels

falila:
# Summary | Résumé

Create secure API to retrieve bearer token for a specific form 
closes #484

| Insert Image showing before the change was made | 
Screen Shot 2021-11-12 at 1 32 48 PM
 |

# Test instructions | Instructions pour tester la modification.

1. Open postman UI
2. restart and run your migration
3. Add a token to template table
4. Paste `http://localhost:3000/api/id/{firmID}/bearer` in Postman replace formID by yours.
5. Get the cookie object from Chrome devTools and add it to Postman header.
6. Hit the url and it should return your bearer token. 

- [X] Have you completely tested the functionality of change introduced in this PR? Is the PR solving the problem it's meant to solve within the scope of the related issue?
- [X] The PR does not introduce any new issues such as failed tests, console warnings or new bugs.
- [X] If this PR adds a package have you ensured its licensed correctly and does not add additional security issues?
- [X] Is the code clean, readable and maintainable? Is it easy to understand and comprehend.
- [X] Have you removed unnecessary lines such as `console.log` statements?
- [X] Does your code have adequate comprehensible comments? Do new functions have [docstrings](https://tsdoc.org/)?
- [X] Have you modified the change log and updated any relevant documentation?
- [X] Is there adequate test coverage? Both unit tests and end-to-end tests where applicable?


lib/integration/dbConnector.ts
```diff
@@ -0,0 +1,17 @@
+import { Client } from "pg";
+import { logMessage } from "../logger";
+
+export const dbConnector = async (): Promise<Client> => {
+  const connectionString: string | undefined = process.env.DATABASE_URL;
+  if (connectionString) {
+    logMessage.debug("Connexion initialization");
```
Moro-Code, line 7: small nit: Connection is spelled wrong

lib/constant.ts
```diff
@@ -0,0 +1,8 @@
+// Template table raw queries
+export const TemplateSelectTokenByFormIdQuery =
```
Moro-Code, line 2: I doubt these queries are used more than once in the code base. Would just use strings rather than adding an additional file and variables that need to be managed

lib/integration/dbConnector.ts
```diff
@@ -0,0 +1,17 @@
...
+    const client = new Client({ connectionString });
+    //Attempt a connexion
+    await client.connect();
+    return client;
+  }
+  throw new Error("connexion error");
+  // connect by using rds client like so
```
Moro-Code: Remove not needed code.

lib/integration/queryManager.js
```diff
@@ -0,0 +1,34 @@
+import { logMessage } from "../logger";
+import { dbConnector } from "./dbConnector";
+
+export const executeQuery = async (sql, params) => {
+  const client = await dbConnector();
+  return client
```
Moro-Code, line 6: fix mixed async await and promise stying.

Should be 

`try{
     const data = await client.query()
}catch(err){
 ...
}`

lib/integration/queryManager.js
```diff
@@ -0,0 +1,34 @@
+import { logMessage } from "../logger";
+import { dbConnector } from "./dbConnector";
+
+export const executeQuery = async (sql, params) => {
+  const client = await dbConnector();
```
Moro-Code, line 5: pass in the client instead of instantiating it from here. Otherwise everytime I call this executeQuery function a new connection is established and returned.

lib/integration/queryManager.js
```diff
@@ -0,0 +1,34 @@
+import { logMessage } from "../logger";
```
Moro-Code, line 1: Why is this a JS file here. If we're adding new things to the code base its a good opportunity to write them in TypeScript

lib/integration/queryManager.js
```diff
@@ -0,0 +1,34 @@
...
+    .catch((error) => {
+      logMessage.error(`{"error: "${formatError(error)}"}`);
+      throw new Error(error);
+    });
+};
+function buildResult(data) {
+  //Result list
+  let result = [];
+  if (data.rowCount > 0) {
+    for (let record in data.rows) {
```
Moro-Code, line 18: can use a map function for this

lib/integration/dbConnector.ts
```diff
@@ -0,0 +1,17 @@
+import { Client } from "pg";
```
Moro-Code, line 1: You need tests for this file

lib/integration/queryManager.js
Moro-Code, line 1: You need tests for this file

pages/api/id/[form]/bearer.ts
```diff
@@ -0,0 +1,42 @@
...
+    res.status(500).json({ error: `Error on Server Side ${err}` });
+  }
+  /**
+   * Will return a bearer token if there is one associated with a given formID
+   * it might return a Not found (404) if token doest exist ( undefined or null)
+   * otherwise 400 as status code
+   * @param formID
+   * @returns
+   */
+  async function getToken(formID: string) {
```
Comment (Moro-Code) [inline pages/api/id/[form]/bearer.ts:23]: Function in a function why ? This should be separated out

lib/integration/queryManager.js
```diff
@@ -0,0 +1,34 @@
+import { logMessage } from "../logger";
+import { dbConnector } from "./dbConnector";
+
+export const executeQuery = async (sql, params) => {
+  const client = await dbConnector();
+  return client
+    .query(sql, params)
+    .then((data) => buildResult(data))
```
Moro-Code, line 8: Don't think this is a good idea. This function should just execute the query and handle any errors. buildResult should be called separately

pages/api/id/[form]/bearer.ts
```diff
@@ -0,0 +1,42 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import { getSession } from "next-auth/client";
+import * as Constants from "../../../../lib/constant";
+import { executeQuery } from "../../../../lib/integration/queryManager";
+
+export const retrieve = async (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
+  try {
+    const session = await getSession({ req });
```
Comment (Moro-Code) [inline pages/api/id/[form]/bearer.ts:8]: Good opportunity for a middleware function here. See James's code https://github.com/cds-snc/platform-forms-client/blob/main/lib/middleware/httpRequestAllowed.ts

pages/api/id/[form]/bearer.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+  /**
+   * Will return a bearer token if there is one associated with a given formID
+   * it might return a Not found (404) if token doest exist ( undefined or null)
+   * otherwise 400 as status code
+   * @param formID
+   * @returns
+   */
+  async function getToken(formID: string) {
+    type BearerResponse = {
```
Comment (Moro-Code) [inline pages/api/id/[form]/bearer.ts:24]: Create a type for the Template record as a whole. Also this should be in a separate file so it could be leveraged elsewhere. This file seems like a good place https://github.com/cds-snc/platform-forms-client/blob/main/lib/types.tsx

pages/api/id/[form]/bearer.ts
```diff
@@ -0,0 +1,42 @@
...
+      bearer_token: string;
+    };
+    if (formID) {
+      //Fetching the token return list of object or an empty array
+      const data = await executeQuery(Constants.TemplateSelectTokenByFormIdQuery, [formID]);
+      if (data && data.length > 0) {
+        const { bearer_token } = data[0] as BearerResponse;
+        if (!bearer_token || bearer_token === null)
+          return res.status(404).json({ error: "Not Found" });
```
Comment (Moro-Code) [inline pages/api/id/[form]/bearer.ts:33]: Would not return a 404 here. The row was found but it contained no bearer token. Should just return { token: null}

pages/api/id/[form]/bearer.ts
```diff
@@ -0,0 +1,42 @@
...
+    await getToken(req.query.form as string);
+  } catch (err) {
+    res.status(500).json({ error: `Error on Server Side ${err}` });
+  }
+  /**
+   * Will return a bearer token if there is one associated with a given formID
+   * it might return a Not found (404) if token doest exist ( undefined or null)
+   * otherwise 400 as status code
+   * @param formID
+   * @returns
```
Comment (Moro-Code) [inline pages/api/id/[form]/bearer.ts:21]: Remove this annotation if you're not going to use it

pages/api/id/[form]/bearer.ts
```diff
@@ -0,0 +1,42 @@
...
+      return res.status(403).json({ error: "Access Denied" });
+    await getToken(req.query.form as string);
+  } catch (err) {
+    res.status(500).json({ error: `Error on Server Side ${err}` });
+  }
+  /**
+   * Will return a bearer token if there is one associated with a given formID
+   * it might return a Not found (404) if token doest exist ( undefined or null)
+   * otherwise 400 as status code
+   * @param formID
```
Comment (Moro-Code) [inline pages/api/id/[form]/bearer.ts:20]: need to describe the parameter

tests/retrieveBearerToken.test.js
```diff
@@ -0,0 +1,109 @@
...
+      user: { email: "admin2@cds.ca", name: "Admin2 user", image: "null" },
+    };
+
+    client.getSession.mockReturnValueOnce(mockSession);
+    fetchMock.mockResponseOnce(JSON.stringify({ token: "akalkdfljasdlkfdsljasfdskfj" }));
+
+    const { req, res } = createMocks({
+      method: "GET",
+      headers: {
+        "Content-Type": "application/json",
```
Moro-Code, line 26: There shouldn't be a body in a get request

tests/retrieveBearerToken.test.js
```diff
@@ -0,0 +1,109 @@
...
+
+jest.mock("next-auth/client");
+
+describe("Test bearer token retrieve api endpoint", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    fetchMock.enableMocks();
+  });
+
+  it("Shouldn't allow GET request without a valid form's id", async () => {
```
Moro-Code, line 14: How are you mocking the database here ?

tests/retrieveBearerToken.test.js
```diff
@@ -0,0 +1,109 @@
...
+        form: "11",
+      },
+    });
+
+    await retrieve(req, res);
+
+    expect(res.statusCode).toBe(403);
+  });
+
+  it("Should allow a request and return Not found (400) resource", async () => {
```
Moro-Code, line 58: Not found is 404

tests/retrieveBearerToken.test.js
```diff
@@ -0,0 +1,109 @@
...
+  });
+
+  it("Shouldn't allow GET request without a valid form's id", async () => {
+    const mockSession = {
+      expires: "1",
+      user: { email: "admin2@cds.ca", name: "Admin2 user", image: "null" },
+    };
+
+    client.getSession.mockReturnValueOnce(mockSession);
+    fetchMock.mockResponseOnce(JSON.stringify({ token: "akalkdfljasdlkfdsljasfdskfj" }));
```
Moro-Code, line 21: Not entirely sure what this is doing. I see it in some other existing test files although I don't see why fetch is being mocked here

tsconfig.json
```diff
@@ -21,6 +21,6 @@
       "@components/*": ["components/*"]
     }
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "lib/integration/queryManager.js", "lib/integration/queryManager.js"],
```
Moro-Code, line 24: duplicated and this should be a TS file

tests/retrieveBearerToken.test.js
```diff
@@ -0,0 +1,109 @@
+import { createMocks } from "node-mocks-http";
+import client from "next-auth/client";
+import { retrieve } from "../pages/api/id/[form]/bearer";
+import fetchMock from "jest-fetch-mock";
```
Moro-Code, line 4: Lets discuss Monday about these tests because I'm not exactly sure they're testing what needs to be tested

Moro-Code (changes requested):
Good work but lots of work needed.

lib/integration/dbConnector.ts
falila: This is a pros name from Clientconfig object that is passed to client.

tests/retrieveBearerToken.test.js
falila: I did the db test but forgot it in a "stash tree". Will bring it back.

tests/retrieveBearerToken.test.js
falila: Good catch!

lib/integration/dbConnector.ts
Moro-Code: Keeping the same name in a log string will prevent creating an extra variable ? Not sure I understand. I mean `Connexion initialization` should be spelled `Connection initialization` and in fact it might be good to add detail here. `Postgres database connection initialization`

tests/retrieveBearerToken.test.js
Moro-Code: I don't mean testing the database I mean mocking out the database connection

lib/integration/dbConnector.ts
falila: I didn't get you were pointing at log string. It's been removed any.

pages/api/id/[form]/bearer.ts
Comment (falila) [reply pages/api/id/[form]/bearer.ts:10]: I've used hooked James function. But you can already see that i kept the session validation inside my handler. The reason is that the isRequestAllow isn't generic enough to cover this case. So i wonder if you wanted to created a new middleware for this Endpoint as well. There's a way to make a generic middleware for this purpose but it might be out of this PR scope. 
ps:  Also updated James code  ( a starting point).

lib/integration/queryManager.ts
```diff
@@ -0,0 +1,30 @@
+import { logMessage } from "../logger";
+import dbConnector from "./dbConnector";
+import { QueryArrayResult } from "pg";
+
+const queryManager = {
+  async executeQuery(sql: string, params: [string]): Promise<QueryArrayResult<unknown[]>> {
+    try {
+      // Get a client
+      const client = dbConnector();
+      //Stablish a connexion
```
Moro-Code, line 10: Minor but do you mean "Establish a connection".

lib/integration/dbConnector.ts
```diff
@@ -0,0 +1,10 @@
...
+
+const dbConnector = (connexion?: string): Client => {
+  const connectionString = connexion ?? process.env.DATABASE_URL;
+  if (!connectionString || connectionString === undefined)
+    throw Error("Invalid db configuration string");
+  const client = new Client({ connectionString });
```
Moro-Code, line 7: Just return the result of this. No need for a variable

lib/integration/queryManager.ts
```diff
@@ -0,0 +1,30 @@
+import { logMessage } from "../logger";
+import dbConnector from "./dbConnector";
+import { QueryArrayResult } from "pg";
+
+const queryManager = {
```
Moro-Code, line 5: Why is this in an object ? Just export each individually

lib/integration/queryManager.ts
```diff
@@ -0,0 +1,30 @@
...
+  async executeQuery(sql: string, params: [string]): Promise<QueryArrayResult<unknown[]>> {
+    try {
+      // Get a client
+      const client = dbConnector();
```
Moro-Code, line 9: Pass this in instead of instantiating here. In this way the same client can be used over and over again rather than creating a new one each time this is called.

lib/integration/queryManager.ts
```diff
@@ -0,0 +1,30 @@
...
+      //Stablish a connexion
+      await client.connect();
+      if (!sql) throw new Error("Invalid query param");
```
Moro-Code, line 12: This should be moved up before everything in the function. Please work on writing clearer error messages. If I'm seeing this in the logs it would be hard to tell where this came from or would take a lot more effort. Something like 

`Empty string passed in to queryManager` would be a lot more clearer

lib/integration/queryManager.ts
```diff
@@ -0,0 +1,30 @@
...
+import dbConnector from "./dbConnector";
+import { QueryArrayResult } from "pg";
+
+const queryManager = {
+  async executeQuery(sql: string, params: [string]): Promise<QueryArrayResult<unknown[]>> {
+    try {
+      // Get a client
+      const client = dbConnector();
+      //Stablish a connexion
+      await client.connect();
```
Moro-Code, line 11: Passing in the client instead will allow you to remove this line

lib/integration/queryManager.ts
```diff
@@ -0,0 +1,30 @@
...
+    } catch (error: unknown) {
+      logMessage.error(`{"error: "${error}"}`);
+      throw new Error(error as string);
+    }
+  },
+
+  getResult(data: QueryArrayResult): unknown[][] {
+    if (data.rowCount > 0) {
+      return data.rows.map((record) => {
+        return record;
```
Moro-Code, line 24: What does this function do exactly ? The map here just returns the same array !

lib/middleware/httpRequestAllowed.ts
```diff
@@ -8,9 +8,19 @@ const isRequestAllowed = (
   return async (req: NextApiRequest, res: NextApiResponse): Promise<unknown> => {
     try {
       const session = await getSession({ req });
-      const requestBody = JSON.parse(req.body);
-      if (requestBody.method && methods.includes(requestBody.method)) {
-        switch (requestBody.method) {
+      let method;
+      let isAllow = false;
```
Moro-Code, line 12: You could write this entire thing in way fewer lines 

```JS
const requestBody = JSON.parse(req.body);
const method =  requestBody.method && methods.includes(requestBody.method) ? requestBody.method : req.method
switch(method){
  ...
}
```

Also if you're changing this file tests need to be reevaluated as well to reflect new code paths

lib/tests/queryManager.test.js
```diff
@@ -0,0 +1,41 @@
...
+  const mClient = {
+    connect: jest.fn(),
+    query: jest.fn(),
+    end: jest.fn(),
+  };
+  return jest.fn(() => mClient);
+});
+
+describe("Testing queryManager and dbConnector", () => {
+  let client;
```
Moro-Code, line 14: probably better to define this within the tests instead as a const

lib/tests/queryManager.test.js
```diff
@@ -0,0 +1,41 @@
+import queryManager from "../integration/queryManager";
+import dbConnector from "../integration/dbConnector";
+
+jest.mock("../integration/dbConnector", () => {
```
Moro-Code, line 4: If you're mocking dbConnector you're not testing dbConnector. You'll need to write tests for dbConnector but good use of a mock for queryManager. Although you can see how this is a lot simpler if you pass in the things the function needs rather than having it as a side effect

lib/integration/dbConnector.ts
```diff
@@ -0,0 +1,10 @@
+import { Client } from "pg";
+
+const dbConnector = (connexion?: string): Client => {
```
Moro-Code, line 3: Not sure again why this is intentionally misspelled. Could just be called connection and it wouldn't break anything

lib/tests/queryManager.test.js
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  it("Should query the db and return an object", async () => {
+    client = dbConnector("postgres://test:test@localhost:5432/test");
+    client.query.mockResolvedValueOnce({ rows: [], rowCount: 1 });
+    const result = await queryManager.executeQuery(
+      "SELECT bearer_token FROM templates WHERE id = ($1)",
+      [1]
+    );
+    expect.assertions(result);
```
Moro-Code, line 32: Not really entirely sure what this expect does. Would probably be better to use a toEqual here and make sure it matches the mocked object returned

lib/tests/queryManager.test.js
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  it("Should throw an error b/c an Invalid query param", async () => {
+    client = dbConnector("postgres://test:test@localhost:5432/test");
+    expect(async function () {
```
Moro-Code, line 37: Test the error string. See if it's the actual error you expect. To make this easier you could use a try catch block here  with await since the test function in asynchronous anyways

pages/api/id/[form]/bearer.ts
Comment (Moro-Code) [reply pages/api/id/[form]/bearer.ts:10]: I just meant write a simple middleware to test if a session exists. Not related to whether the method is allowed. No worries at this point its an extra thing.

pages/api/id/[form]/bearer.ts
```diff
@@ -0,0 +1,45 @@
...
+    res.status(500).json({ error: `Error on Server Side ${err}` });
+  }
+};
+/**
+ * Will return a bearer token if there is one associated with a given formID
+ * It might return a Not found (404) if token doest exist ( undefined or null)
+ * otherwise 400 as status code.
+ * @param formID A form id
+ */
+async function getToken(res: NextApiResponse, formID: string) {
```
Comment (Moro-Code) [inline pages/api/id/[form]/bearer.ts:25]: probably best to separate this out in a separate file in the lib folder so it can be utilized elsewhere. This can even be made into a middleware where the resulting handler gets injected with a token. In any case in the interest of having this PR done we can skip this for now but just a note in the future

tests/retrieveBearerToken.test.js
```diff
@@ -0,0 +1,136 @@
...
+      headers: {
+        "Content-Type": "application/json",
+      },
+      query: {
+        form: "", //An empty form ID
+      },
+    });
+
+    await retrieve(req, res);
+
```
Moro-Code, line 32: Would be good to have an expect for the body of the response as well

tests/retrieveBearerToken.test.js
```diff
@@ -0,0 +1,136 @@
...
+      method: "GET",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      query: {
+        form: "999999999",
+      },
+    });
+    await retrieve(req, res);
+
```
Moro-Code, line 133: write expect for body of response as well

tsconfig.json
```diff
@@ -21,6 +21,6 @@
       "@components/*": ["components/*"]
     }
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "lib/integration/queryManager.ts"],
```
Moro-Code, line 24: There should be no need to specifically include this

tests/retrieveBearerToken.test.js
```diff
@@ -0,0 +1,136 @@
+import { createMocks } from "node-mocks-http";
+import client from "next-auth/client";
+import { retrieve } from "../pages/api/id/[form]/bearer";
+import queryManager from "../lib/integration/queryManager";
+
+jest.mock("next-auth/client");
+
+describe("Test bearer token retrieve API endpoint", () => {
```
Moro-Code, line 8: Don't see any tests for the case there is an error with the DB connection or something

tests/retrieveBearerToken.test.js
```diff
@@ -0,0 +1,136 @@
...
+      },
+      query: {
+        form: "12",
+      },
+    });
+
+    await retrieve(req, res);
+    const expectedValue = { token: "toekakdnaodk" };
+    expect(res.statusCode).toBe(200);
+    expect(JSON.parse(res._getData())).toEqual(expect.objectContaining(expectedValue));
```
Moro-Code, line 90: Just use `data[0]` instead of another variable but this is minor

Moro-Code (changes requested):
More changes needed but the PR should be good to go as long as you address each one of them

lib/integration/queryManager.ts
falila: In order to mock these functions  i either have to create to separate module  (file) which i found heavy for this purpose.So i create a namespace instead.

tests/retrieveBearerToken.test.js
falila: There's a  tiny difference between these two variables. `Data` doesn't contain token property and i want to make sure the value is passed to the response as `token: xxxxxxxxx` and not `baerer_token : xxxxx` instead.

falila: But yes it can be replaced by the value itself.

lib/integration/queryManager.ts
falila: It's been already refactored. 
The concern is to only extract the result from metadata  that is returned.

`rowCount: 1
oid: null
rows: [object Object]
fields: [object Object]
_parsers: function noParse (val) {
  return String(val);
}
_types: [object Object]
RowCtor: null
rowAsArray: false`

lib/middleware/httpRequestAllowed.ts
falila: Yeah i know but you may have forget that  there's  no requestBody in the context of bearer Api. Meaning we do not send any data like template does. So requestBody will be undefined.

pages/api/id/[form]/bearer.ts
Comment (falila) [reply pages/api/id/[form]/bearer.ts:11]: Done!

lib/integration/queryManager.ts
Moro-Code: Considering you only have one function now you don't need this. Also to mock individual functions you don't need to create a separate file. You can just `jest.mock()` the module and you can define the mock behaviour in the individual tests. See the following section https://jestjs.io/docs/mock-functions#mocking-modules

lib/integration/queryManager.ts
```diff
@@ -0,0 +1,21 @@
...
+const queryManager = {
+  async executeQuery(
+    client: Client,
+    sql: string,
+    params: [string]
+  ): Promise<QueryArrayResult<unknown[]>> {
+    try {
+      if (!sql) throw new Error("Empty or invalid sql passed as a parameter");
+      //Establish a connexion
+      await client.connect();
```
Moro-Code, line 13: I would assume that at this point a connection as been established so you can run just queries. Leave the handling of the client outside this function

lib/middleware/httpRequestAllowed.ts
Moro-Code: In that case you can do this 

```JS
const requestBody = req.body ? JSON.parse(req.body): undefined;
const method =  requestBody?.method && methods.includes(requestBody.method) ? requestBody.method : req.method
switch(method){
  ...
}
```

Accounts for no req.body and still way fewer lines :)

lib/tests/dbconnector.test.js
```diff
@@ -0,0 +1,44 @@
...
+    jest.clearAllMocks();
+  });
+
+  it("Should create and return a valid client", async () => {
+    const client = dbConnector("postgres://test:test@localhost:5432/test");
+    // Try to connect
+    client.connect();
+    expect(client.connect).toBeCalledTimes(1);
+    // Verify query method
+    client.query("SELECT bearer_token FROM templates;", []);
```
Moro-Code, line 23: Hmmm don't see a point in this considering at this point you're just validating your mock object. I think the connect was enough to do that but this is minor

lib/tests/dbconnector.test.js
```diff
@@ -0,0 +1,44 @@
...
+
+  it("Should throw an error reason: there's no db configuration string found.", async () => {
+    try {
+      dbConnector();
+    } catch (error) {
+      expect(error).toBeDefined();
+    }
+  });
+
+  it("Should successfully return a connexion client with Env configuration settings. ", async () => {
```
Moro-Code, line 35: Hahahaha I think connection is spelled connexion in french or something. With that in mind maybe its not misspelled

lib/tests/httpSessionExist.test.js
```diff
@@ -0,0 +1,45 @@
+import { createMocks } from "node-mocks-http";
+import client from "next-auth/client";
+import isUserSessionExist from "../middleware/HttpSessionExist";
+
+jest.mock("next-auth/client");
+
+const handler = async (req, res) => {
```
Moro-Code, line 7: This should be a mock function. That way you can test if its actually called ! But in any case this works for this purpose just for future reference

lib/tests/httpSessionExist.test.js
```diff
@@ -0,0 +1,45 @@
+import { createMocks } from "node-mocks-http";
```
Moro-Code, line 1: Good tests well done !

pages/api/id/[form]/bearer.ts
```diff
@@ -0,0 +1,41 @@
...
+    res.status(500).json({ error: `Error on Server Side` });
+  }
+};
+/**
+ * Will return a bearer token if there is one associated with a given formID
+ * It might return a Not found (404) if token doest exist ( undefined or null)
+ * otherwise 400 as status code.
+ * @param formID A form id
+ */
+export async function getToken(res: NextApiResponse, formID: string) {
```
Comment (Moro-Code) [inline pages/api/id/[form]/bearer.ts:20]: Need a return type on this function

tsconfig.json
```diff
@@ -21,6 +21,6 @@
       "@components/*": ["components/*"]
     }
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx",],
```
Moro-Code, line 24: Not valid JSON remove the last comma

Moro-Code (changes requested):
Almost there !

lib/integration/queryManager.ts
falila: that means bringing back  `await client.connect()` inside dbConnector.ts

Moro-Code (approved):
Think we're good to merge this now. Good work !

PR #542, opened 2021-11-26
Create secure API to deactivate a form owner associated with a form

### LINKED ISSUE #501: Create secure API to deactivate a form owner associated with a form
As an admin I would like to be able to deactivate but not delete a form owner so that they may not have access to the form data but also be able to be referenced from log data should a need arise. 

Please see [design doc 003](https://docs.google.com/document/d/10cM-GocFqGW8TIjohzSNpz9m9r-CxDh6b7dtvB7dya0/edit?usp=sharing) for additional details

Acceptance Criteria:

- [x] An API route in the NextJS application corresponding to the following file path `pages/api/id/[form]/owners.js`
- [ ] This route must accept a `PUT` request. 
- [ ] The method can only be accessed if there is a valid session in the request from next-auth. If the session is not valid then a 403 status code should be returned 
- [ ] The payload of the request should be validated to have the following keys. If validation fails a 400 status code should be returned 
```JSON
{
   "email": "The email of the user associated to the form"
   "active": "Boolean value indicating whether the form owner is active or not"
}
```
- [ ] If the request is valid the relevant record should be updated with the value for the active field in the payload
- [ ] If the record could not be found a 404 status code should be returned

falila:
# Summary | Résumé

As an admin I would like to be able to deactivate but not delete a form owner so that they may not have access to the form data but also be able to be referenced from log data should a need arise.
close #501 

Insert Image showing after a change was made 
Screen Shot 2021-11-26 at 1 39 40 PM

1. Open postman UI
2. restart and run your migration
3. Add emails  and template_id to form_users table
`INSERT INTO form_users(
	template_id, email)
	VALUES (?, ?);`
4. Paste http://localhost:3000/api/id/{firmID}/owners in Postman .
5. Get the cookie object from Chrome devTools and add it to Postman header.
6. Hit the url and it should return : `[ {  id: 1 }, {  id: 2 } ]`.

# Pull Request Checklist

- [x] Have you removed unnecessary lines such as `console.log` statements?


pages/api/id/[form]/owners.ts
```diff
@@ -38,5 +40,40 @@ export async function getEmailListByFormID(
...
+ * @param req
+ * @param res
+ */
+export async function activateOrDeactivateFormOwners(
+  req: NextApiRequest,
+  res: NextApiResponse
+): Promise<void> {
+  //Extracting req body
+  const requestBody =
+    req.body && Object.keys(req.body).length > 0 ? JSON.parse(req.body) : undefined;
```
Comment (Moro-Code) [inline pages/api/id/[form]/owners.ts:58]: Think checking if req.body is truthy is enough of a check. No need for Object.keys

tests/api/id/[form]/owners.test.js
```diff
@@ -172,3 +172,118 @@ describe("Test Owners : retrieve list of emails API endpoint", () => {
     expect(JSON.parse(res._getData()).error).toEqual("Error on Server Side");
   });
 });
+
```
Comment (Moro-Code) [inline tests/api/id/[form]/owners.test.js:175]: A much clearer way of organizing these tests would be the following. But minor point 

```js
describe("/id/[form]/owners", () => {
    describe("GET", () => {...} )
    describe("PUT", () => {...} )
})
```

pages/api/id/[form]/owners.ts
```diff
@@ -38,5 +40,40 @@ export async function getEmailListByFormID(
...
+    //Invalid payload
+    return res.status(400).json({ error: "Invalid payload fields are not define" });
+  }
 
-export default isRequestAllowed(["GET", "POST"], isUserSessionExist(owners));
+  const formID = req.query.form as string;
+  const { email, active } = requestBody;
+  if (!formID) return res.status(400).json({ error: "Malformed API Request Invalid formID" });
+  //Update form_users's records
+  const resultObject = await executeQuery(
```
Comment (Moro-Code) [inline pages/api/id/[form]/owners.ts:69]: How are you sure here that active is not a value of "sdbdbfwbfwe" for example. Why are we using string "1" or "0" instead of true or false

Moro-Code (changes requested):
Minor points but a larger question on why the value for active is a string as opposed to a real boolean

pages/api/id/[form]/owners.ts
Comment (falila) [reply pages/api/id/[form]/owners.ts:69]: Will support both format  1/0 and true/false. I've also added a test case and a validation point.

Comment (falila) [reply pages/api/id/[form]/owners.ts]: sample payload format : 
`{       
        "email": "forms@cds.ca",
        "active":  "true"
}`

Comment (Moro-Code) [reply pages/api/id/[form]/owners.ts]: Why is true or false a string here. It should just be a boolean true or false

pages/api/id/[form]/owners.ts
```diff
@@ -38,5 +40,47 @@ export async function getEmailListByFormID(
...
+): Promise<void> {
+  //Extracting req body
+  const requestBody = req.body ? JSON.parse(req.body) : undefined;
+  //Payload validation
+  if (!requestBody?.email || !requestBody?.active) {
+    //Invalid payload
+    return res.status(400).json({ error: "Invalid payload fields are not define" });
+  } else {
+    const { email } = requestBody; // TODO email should also be checked i.e (valid Gov email). See #491
+    let { active } = requestBody;
```
Comment (Moro-Code) [inline pages/api/id/[form]/owners.ts:64]: active should just be a boolean in JSON. JSON just supports true or false booleans

Moro-Code (changes requested):

pages/api/id/[form]/owners.ts
Comment (falila) [reply pages/api/id/[form]/owners.ts]: Yeah , this was just a sample. But you can find more examples in our tests.  Like `{ id: "2", email: "forms@cds.ca", active: false },`

Moro-Code (approved):
Looks good to merge

PR #557, opened 2021-12-13
feat: form response retrieval authentication api

### LINKED ISSUE #502: Create Form Response Retrieval Authentication API
As a form owner I would like to retrieve a temporary token so that I may authenticate against GC Form services such as the Form Response Retrieval API

Please see [design doc 003](https://docs.google.com/document/d/10cM-GocFqGW8TIjohzSNpz9m9r-CxDh6b7dtvB7dya0/edit?usp=sharing) for specific details about implementation

Acceptance Criteria:
- [ ] An API route `/api/token/temporary` which accepts only POST requests over HTTPS
- [ ] The handler should validate that the request includes a proper bearer token ( see design doc )
- [ ] If the request does not include a proper bearer token a 403 response should be returned 
- [ ] The function should validate that a JSON payload is included with a single key value pair where the key is `email` and the value is the email of the form owner to generate the temporary token for. 
- [ ] Should the payload not be valid a 400 response should be returned 
- [ ] The relevant record for the form ( form ID in claims of bearer token ) and the user should be retrieved. Should the user not exist a 200 response status code should be returned ( to prevent sniffing out form owner emails ) 
- [ ] If the record for the form owner exists a temporary token should be generated and signed using the jsonwebtoken package. An `exp` claim should be set on the token with a value corresponding to a unix timestamp 7 days from when the token was generated. 
- [ ] The token should be saved to the database.
- [ ] Every attempt to generate a temporary token should be logged with the form id and the email of the form owner.

jeberhardt:
# Summary | Résumé
As a form owner I would like to retrieve a temporary token so that I may authenticate against GC Form services such as the Form Response Retrieval API

closes #502 

To test this API specifically:
- Open Postman UI
- The path for the API is: **/api/token/temporary**
- Payload sample (body / raw): **{ "email": "gc-forms@canada.ca" }**
- Authorization: Select type **Bearer Token**
- Enter the bearer token for the authorization
- Submit the request; the response should be: **{"message": "success"}**

Entering an incorrect payload, or incorrect bearer token should respond with error messages.

- Will need to send out a message via Notify upon temporary token generation. (#558)

# Pull Request Checklist


lib/types-database.ts
```diff
@@ -0,0 +1,9 @@
+export interface formUser {
```
Moro-Code, line 1: We have a types folder. Would probably put that there instead of a separate file

pages/api/token/temporary.ts
```diff
@@ -0,0 +1,75 @@
...
+    req: NextApiRequest,
+    res: NextApiResponse,
+    bearerTokenPayload: BearerTokenPayload,
+    email: string
+  ) => void
+) => {
+  return async (req: NextApiRequest, res: NextApiResponse, options?: unknown): Promise<unknown> => {
+    const requestBody = JSON.parse(req.body);
+    try {
+      if (Object.prototype.hasOwnProperty.call(requestBody, "email")) {
```
Moro-Code, line 24: This is very verbose. Any reason you're not just calling this directly from request body ? Or better yet doing something like `if(requestBody?.email)`

pages/api/token/temporary.ts
```diff
@@ -0,0 +1,75 @@
...
+};
+
+const handler = async (
+  req: NextApiRequest,
+  res: NextApiResponse,
+  bearerTokenPayload: BearerTokenPayload,
+  email: string
+) => {
+  try {
+    // TODO: create temporary token
```
Moro-Code, line 42: remove TODO line here

pages/api/token/temporary.ts
```diff
@@ -0,0 +1,75 @@
...
+    const formID = bearerTokenPayload.formID;
+    if (formID == undefined) {
+      res.status(403).json({ error: "Invalid form request." });
+      return;
+    }
+    const form = await getForm(formID, email);
```
Moro-Code, line 48: You're doing an extra query for no reason. If the form doesn't exist the UPDATE query will just return a count of 0.

lib/middleware/bearerToken.ts
```diff
@@ -0,0 +1,60 @@
...
+  formID?: string;
+}
+
+const validate = (
+  handler: (req: NextApiRequest, res: NextApiResponse, options?: unknown) => void
+) => {
+  return async (req: NextApiRequest, res: NextApiResponse): Promise<unknown> => {
+    try {
+      const token = getBearerToken(req);
+      const bearerTokenPayload = jwt.verify(token, process.env.TOKEN_SECRET || "");
```
Moro-Code, line 17: Bearer token also needs to be validated to be the same value in the database. If a new bearer token is generated and you do not check this then there is no way to invalidate old tokens

pages/api/token/temporary.ts
```diff
@@ -0,0 +1,75 @@
...
+      } else {
+        res.status(400).json({ error: "Invalid payload" });
+      }
+    } catch (err) {
+      res.status(400).json({ error: "Invalid payload" });
+    }
+  };
+};
+
+const handler = async (
```
Moro-Code, line 35: Missing sending to Notify ?

Moro-Code (changes requested):
Somethings to be resolved here. Is there any reason we separated out sending the token through Notify from this task ? Aside from that you're missing other areas of the acceptance criteria mainly `The handler should validate that the request includes a proper bearer token ( see design doc )` 

The design doc specifies the following 
1. The bearer token is valid according to its signature ( it has not been modified )
2. The bearer token has not expired
3. The bearer token matches that for the specific template in the database 
4. The email provided is associated with the template specified
5. The active field for the record has a value of True

Mainly 3 and 5

lib/middleware/bearerToken.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+export const updateTemporaryToken = async (
+  temporary_token: string,
+  email: string,
+  template_id: string
+): Promise<QueryResult> => {
+  return executeQuery(
+    await dbConnector(),
+    "UPDATE form_users SET temporary_token = ($1), updated_at = current_timestamp WHERE email = ($2) and template_id = ($3)",
```
Moro-Code, line 55: You also need to make sure the user is active. You can add that to the where

Moro-Code (changes requested):

tests/api/token/temporary.test.js
```diff
@@ -0,0 +1,86 @@
...
+  it("works as expected", async () => {
+    const token = jwt.sign({ formID: "1" }, process.env.TOKEN_SECRET, { expiresIn: "1y" });
+    const { req, res } = createMocks({
+      headers: {
+        "Content-Type": "application/json",
+        Origin: "http://localhost:3000",
+        authorization: `Bearer ${token}`,
+      },
+      body: JSON.stringify({
+        method: "POST",
```
falila, line 37: Why `method = POST` in the body ?

tests/api/token/temporary.test.js
```diff
@@ -0,0 +1,86 @@
...
+  beforeAll(() => {
+    process.env.TOKEN_SECRET = "some_secret_some_secret_some_secret_some_secret";
+    process.env.TOKEN_SECRET_WRONG = "wrong_secret_wrong_secret_wrong_secret_wrong_secret";
+  });
+  afterAll(() => {
+    delete process.env.TOKEN_SECRET;
+    delete process.env.TOKEN_SECRET_WRONG;
+  });
+
+  it("works as expected", async () => {
```
falila, line 28: The test description should be explicite. I can't tell what is expected only by reading its description.

lib/tests/bearerToken.test.js
```diff
@@ -0,0 +1,67 @@
...
+
+  it("rejects an expired bearer token", async () => {
+    const token = jwt.sign({ formID: "1", exp: 1636501665 }, process.env.TOKEN_SECRET);
+    const { req, res } = createMocks({
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        authorization: `Bearer ${token}`,
+      },
+      body: JSON.stringify({
```
falila, line 55: We shouldn't pass method in the body.

falila (changes requested):

pages/api/token/temporary.ts
jeberhardt: This was split out into Issue #558.

lib/middleware/bearerToken.ts
jeberhardt: The "active" check was originally done with the `getForm` call, but now that that call has been removed, I'll add it to this query.

lib/middleware/bearerToken.ts
```diff
@@ -0,0 +1,60 @@
...
+      const bearerTokenPayload = jwt.verify(token, process.env.TOKEN_SECRET || "");
+      return handler(req, res, bearerTokenPayload);
+    } catch (err) {
+      res.status(403).json({ error: "Missing or invalid bearer token." });
+    }
+  };
+};
+
+const getBearerToken = (req: NextApiRequest) => {
+  const authHeader = String(req.headers["authorization"] || "");
```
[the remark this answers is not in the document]
jeberhardt: This is just an extra check to ensure that the string "Bearer " exists in the authHeader. With your approach, if "Bearer " doesn't exist, it'll just return the original string.

pages/api/token/temporary.ts
```diff
@@ -0,0 +1,75 @@
...
+      res.status(403).json({ error: "Invalid form request." });
+      return;
+    }
+    const form = await getForm(formID, email);
+    if (form) {
+      const temporaryToken = createTemporaryToken(email);
+      await updateTemporaryToken(temporaryToken, email, formID);
+      logMessage.info(`Temporary Token Generated: Form ID: ${formID} Email: ${email}`);
+      res.status(200).json({ message: "success" });
+    }
```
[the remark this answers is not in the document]
jeberhardt: As per @Moro-Code's comment, I've removed `getForm` altogether.

lib/middleware/bearerToken.ts
```diff
@@ -0,0 +1,60 @@
...
+const validate = (
+  handler: (req: NextApiRequest, res: NextApiResponse, options?: unknown) => void
+) => {
+  return async (req: NextApiRequest, res: NextApiResponse): Promise<unknown> => {
+    try {
+      const token = getBearerToken(req);
+      const bearerTokenPayload = jwt.verify(token, process.env.TOKEN_SECRET || "");
+      return handler(req, res, bearerTokenPayload);
+    } catch (err) {
+      res.status(403).json({ error: "Missing or invalid bearer token." });
```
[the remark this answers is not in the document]
jeberhardt: While we aren't enforcing it, the `no-useless-return` lint rule should apply here.

lib/middleware/bearerToken.ts
```diff
@@ -0,0 +1,60 @@
...
+const getBearerToken = (req: NextApiRequest) => {
+  const authHeader = String(req.headers["authorization"] || "");
+  if (authHeader.startsWith("Bearer ")) {
+    return authHeader.substring(7, authHeader.length);
+  } else {
+    throw new Error("Missing bearer token.");
+  }
+};
+
+export const createTemporaryToken = (email: string): string => {
```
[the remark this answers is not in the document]
jeberhardt: I started putting it in there, but there aren't any other database functions in `crud.ts`, so it didn't really make much sense.

lib/middleware/bearerToken.ts
jeberhardt: Updated `bearerToken.ts` with a function to check the database.

lib/middleware/bearerToken.ts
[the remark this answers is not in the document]
Moro-Code: Hmmm doesn't really make sense here as well. Considering this is a crud operation thought it would make more sense in that file. I think you can move this then to the handler file itself. If it needs to be used by something else we can figure out a place for it

lib/middleware/bearerToken.ts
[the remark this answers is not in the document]
Moro-Code: It's a convention we're using across the code base to show that a definitive API response is returned. Think its good to be consistent

lib/types-database.ts
Moro-Code: Sorry I meant the types.tsx. It should probably be in there

jeberhardt: Removed the `getForm` function, which removed the need for this interface.

falila (approved):
Awesome work.

lib/middleware/bearerToken.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Checks the existance of the bearer token in the database
+ *
+ * @param bearerToken - the token to look for in the `templates` table in the database
+ * @returns boolean result of existance of bearerToken in the database
+ */
+const checkBearerToken = async (bearerToken: string): Promise<boolean> => {
+  const queryResults = await executeQuery(
+    await dbConnector(),
+    "SELECT bearer_token from templates WHERE bearer_token = ($1)",
```
Moro-Code, line 57: This would mean that I can use the bearer token for any other form and pass this check. Also there is no index on the bearer token and this is a much slower query

Moro-Code (changes requested):
Looks to go but one hurdle which I commented on. Good work!

lib/middleware/bearerToken.ts
```diff
@@ -0,0 +1,53 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import jwt from "jsonwebtoken";
+import { getTokenById } from "../../pages/api/id/[form]/bearer";
```
Moro-Code, line 3: I think we could move this into the crud file in the lib folder rather than importing from another api file. Something for another PR though

pages/api/token/temporary.ts
```diff
@@ -0,0 +1,111 @@
...
+const handler = async (
+  req: NextApiRequest,
+  res: NextApiResponse,
+  bearerTokenPayload: BearerTokenPayload,
+  email: string
+) => {
+  try {
+    const formID = bearerTokenPayload.formID;
+    if (formID == undefined) {
+      res.status(403).json({ error: "Invalid form request." });
```
Moro-Code, line 53: Really hate to nit here. Think I mentioned it before in a previous review you can just do the return on the same line as the res.status().json(). We do this across our code base so might as well keep it consistent. Also looks cleaner than having an empty return line.

https://github.com/cds-snc/platform-forms-client/pull/557#discussion_r770649977

Moro-Code (commented):
Think we're good to go except for some code style changes that need to be made.

Moro-Code (approved):
Good to go !

PR #573, opened 2022-01-06
Feature/refresh bearer token

### LINKED ISSUE #486: Create UI tab on form settings page to view and refresh bearer tokens
As an admin user I would like a way to visually retrieve and regenerate bearer tokens for a specific form so that I may use them to authenticate against GC Form services

Acceptance Criteria 
- [x] A tab in the settings page for a specific form which contains a textarea containing the current bearer token and a button underneath to refresh the bearer token.
- [x] On pressing the refresh bearer token button a request should be sent to the refresh bearer token API ( see #483 ) and the textarea should be updated with the latest bearer token 
- [x] If there is an error. An error text should be displayed above the textarea field. 
- [x] Adequate unit testing on all levels 
- [ ] Adequate cypress end to end testing on all levels 
- [x] All should be WCAG 2.0 accessible and screen reader friendly.

jeberhardt:
# Summary | Résumé

Closes #486 

| 
Screen Shot 2022-01-18 at 10 37 33 AM
 | 
Screen Shot 2022-01-18 at 11 14 05 AM
 |

- Go to the Settings page for a form.
- Click on the Tab labelled: "**Bearer Token**"
- Click the button "Refresh Token". The token in the textarea should update. 

- It'd be useful to have a "Copy Token" button.
- There are no tests for Cypress for this PR, since the page requires authentication, and how that happens with Cypress hasn't been figured out.

# Pull Request Checklist


components/admin/BearerRefresh/BearerRefresh.tsx
```diff
@@ -0,0 +1,77 @@
...
+        className="gc-textarea full-height font-mono"
+        data-testid="bearerToken"
+        defaultValue={bearerTokenState}
+        readOnly
+      ></textarea>
+      <br />
+      <Button
+        type="button"
+        onClick={async () => {
+          const { bearerToken } = (await handleRefreshBearerToken(
```
craigzour, line 65: Not sure how feasible it is but it would be cleaner if we could extract that logic outside of the `onClick` parameter into a small function.

public/static/locales/fr/admin-templates.json
```diff
@@ -15,6 +15,9 @@
     "delete": "Supprimer le formulaire",
     "delete-check": "Êtes-vous sûr de vouloir supprimer ce formulaire ?",
     "confirm-delete": "Confirmer Supprimer",
-    "new": "Le formulaire a été créé avec succès"
+    "new": "Le formulaire a été créé avec succès",
+    "formID": "Form ID:",
+    "bearerToken": "Bearer Token:",
+    "edit": "Edit Form Config File:"
```
craigzour, line 21: Some of the new keys you have added to the english json file are missing here

craigzour (changes requested):
I added two comments which I will let you resolve :)

components/admin/BearerRefresh/BearerRefresh.tsx
jeberhardt: Updated, thanks for the comment.

public/static/locales/fr/admin-templates.json
```diff
@@ -16,8 +16,14 @@
     "delete-check": "Êtes-vous sûr de vouloir supprimer ce formulaire ?",
     "confirm-delete": "Confirmer Supprimer",
     "new": "Le formulaire a été créé avec succès",
-    "formID": "Form ID:",
-    "bearerToken": "Bearer Token:",
-    "edit": "Edit Form Config File:"
+    "formID": "ID de formulaire :",
```
craigzour, line 19: Could be "Identifiant du formulaire" in French

public/static/locales/fr/admin-templates.json
```diff
@@ -16,8 +16,14 @@
...
+    "bearerToken": "Bearer Token :",
```
craigzour, line 20: Could be "Jeton d'accès" in French

public/static/locales/fr/admin-templates.json
```diff
@@ -16,8 +16,14 @@
...
+    "refreshButton": "Refresh Token",
```
craigzour, line 21: Could be "Renouveler le jeton d'accès" in French

public/static/locales/fr/admin-templates.json
```diff
@@ -16,8 +16,14 @@
...
+    "bearerTokenRefreshError": "Unable to refresh bearer token.",
```
craigzour, line 22: Could be "Le renouvellement du jeton d'accès n'a pas fonctionné" in French

public/static/locales/fr/admin-templates.json
```diff
@@ -16,8 +16,14 @@
...
+    "edit": "Edit Form Config File:",
```
craigzour, line 23: Could be "Modifier le formulaire" in French

public/static/locales/fr/admin-templates.json
```diff
@@ -16,8 +16,14 @@
...
+    "tabLabels": {
+      "jsonUpload": "JSON Upload",
```
craigzour, line 25: Could be "Téléversement JSON" in French

public/static/locales/fr/admin-templates.json
```diff
@@ -16,8 +16,14 @@
...
+      "bearerToken": "Bearer Token"
```
craigzour, line 26: Could be "Jeton d'accès" in French

craigzour (commented):
Here are my suggested translations for the `public/static/locales/fr/admin-templates.json` file.

public/static/locales/fr/admin-templates.json
jeberhardt: I copied that from the same file... I'll change it to your suggestion in both places.

craigzour (approved):
LGTM!

components/admin/BearerRefresh/BearerRefresh.tsx
```diff
@@ -0,0 +1,69 @@
...
+        id="bearerToken"
+        rows={3}
+        name="bearerToken"
+        className="gc-textarea full-height font-mono"
+        data-testid="bearerToken"
+        defaultValue={bearerTokenState}
+        readOnly
+      ></textarea>
+      <br />
+      <Button type="button" onClick={() => handleRefreshBearerToken(form?.formID)}>
```
bryan-robitaille, line 62: If `form` is a required prop we can remove the `?` as `formID` is a required property on the Interface.

components/admin/BearerRefresh/BearerRefresh.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Connects to the database and refreshes the bearer token for the specified form ID.
+   *
+   * @param formID
+   */
+  const handleRefreshBearerToken = async (formID: number | undefined) => {
+    try {
+      setErrorState({ message: "" });
+      const serverResponse = await axios({
```
bryan-robitaille, line 31: When using a network call there should always be a Loading state so that the user knows that there is an action happening.  With an axios timeout set to `0` the component will wait indefinitely for a response and the user could continue to click the button repeatedly as no action is being demonstrated on the screen.

components/admin/BearerRefresh/BearerRefresh.tsx
```diff
@@ -0,0 +1,69 @@
...
+        url: `/api/id/${formID}/bearer`,
+        method: "POST",
+        timeout: 0,
+      });
+      const { bearerToken } = serverResponse.data as unknown as BearerResponse;
+      setBearerTokenState(bearerToken);
```
bryan-robitaille, line 37: This can also go back to our conversation the other day where the `form.bearerToken` prop has essentially been mutated but the parent components are not aware of the change.  Will work for now but ideally the `useRefesh` hook could be used here to ensure state remains constant across all components that use the 'form' prop.

bryan-robitaille (changes requested):
I think the most important addition to this PR would be a Loading state of some sort to let the user know that an action has been taken and we're awaiting a response with the refreshed bearer token.

components/admin/BearerRefresh/BearerRefresh.tsx
jeberhardt: Added a "Loader...".

components/admin/BearerRefresh/BearerRefresh.tsx
jeberhardt: Changed it so that the props contain just the "formID", and is never updated.

craigzour (approved):
LGTM :)

components/admin/BearerRefresh/BearerRefresh.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+const BearerRefresh = (props: BearerRefreshProps): React.ReactElement => {
+  const { formID } = props;
+  const [bearerTokenState, setBearerTokenState] = useState("");
+  const { t } = useTranslation("admin-templates");
+  const [errorState, setErrorState] = useState({ message: "" });
+  const [submitting, setSubmitting] = useState(false);
+
+  useEffect(() => {
+    setBearerTokenState("");
```
bryan-robitaille, line 21: No need to call `setBearerTokenState("")` here as the state is already initialized when the useState hook is created using `useState("")`

components/admin/BearerRefresh/BearerRefresh.tsx
```diff
@@ -0,0 +1,98 @@
...
+  }, []);
+
+  const getBearerToken = async (formID: number) => {
+    try {
+      setSubmitting(true);
+      setErrorState({ message: "" });
+      const serverResponse = await axios({
+        url: `/api/id/${formID}/bearer`,
+        method: "GET",
+        timeout: 0,
```
bryan-robitaille, line 32: Similiar to the Submit API I'm wondering if we should timeout the call after 60 seconds but we can leave infinite for the dev environment.  
[Submit API Example](https://github.com/cds-snc/platform-forms-client/blob/906a662a5630e9c7ad91080743fff2adb2f4aa05/lib/integration/helpers.ts#L270)

components/admin/BearerRefresh/BearerRefresh.tsx
```diff
@@ -0,0 +1,98 @@
...
+   * @param formID
+   */
+  const handleRefreshBearerToken = async (formID: number) => {
+    try {
+      setSubmitting(true);
+      setErrorState({ message: "" });
+      const serverResponse = await axios({
+        url: `/api/id/${formID}/bearer`,
+        method: "POST",
+        timeout: 0,
```
bryan-robitaille, line 56: Same comment regarding timeout as the `GET` bearer token call.

components/containers/DynamicForm/Settings.tsx
```diff
@@ -51,20 +53,32 @@ export const FormSettings = (props: FormSettingsProps): React.ReactElement => {
...
+      <Tabs>
+        <TabList>
+          <Tab>{t("settings.tabLabels.jsonUpload")}</Tab>
+          <Tab>{t("settings.tabLabels.bearerToken")}</Tab>
+        </TabList>
+
+        <TabPanel>
+          <div>{newText}</div>
+          <h2>{t("settings.edit")}</h2>
+          <JSONUpload form={form}></JSONUpload>
```
bryan-robitaille, line 68: We should align with the new props requirement and only be passing in the `formID` instead of the complete `form` object.

bryan-robitaille (changes requested):
Few small things around best practices but after that should be ready to merge.

components/containers/DynamicForm/Settings.tsx
jeberhardt: Great idea; I think that's outside the scope of this PR though, as it's changing the `JSONUpload` component, which is otherwise unchanged.

bryan-robitaille: Agreed.  This is my bad, when I was going through the components tree I mistook JSONUpload for the new component that is introduced in this PR.

bryan-robitaille (approved):

PR #579, opened 2022-01-11
Replace Retrieval Api lambda implementation by an App backEnd Api

### LINKED ISSUE #481: Modify Retrieval API
Update code that was originally developed during a prototyping sprint in JS to TypeScript and verify/add required functionality + tests.

Current API endpoint found at 'localhost:3000/api/retrieval'
Associated file(s):
- `./pages/api/retrieval.js`
- [Lambda Function holding previous logic](https://github.com/cds-snc/forms-staging-terraform/blob/main/forms/aws/lambda/retrieval/retrieval.js)

Hints:
- To aid in coding this feature remove the `const session = await getSession({ req });` check

In Scope:
- Authorization / Verification of Bearer Token

The existing Retrieval Lambda functionality can be moved to the Next JS application and modified to meet the new feature requirements.

Acceptance Criteria:
- [x] Receives 'formID' and 'maxRecords' parameters
  - formID is the ID of the form 
  - maxRecords is the maximum number of form responses to return.
- [x] Returns form responses up to the maxRecords parameter
- [x] Marks form responses retrieved with the 'Retrieved' attribute
- [x] Endpoint returns form responses in the shape of a JSON object containing an Array of form response objects.

```json
{
  "responses":  [
    {
       ...form response
    },
    {
       ...form response
    },
  ]
}
```
- [x] Error scopes are properly identified and tested.
- [x] Error on retrieval is sanitized before being logged on the server for potentially sensitive information and a generic error is passed back to the client (HTML code 500ish)
- [x] Error in API after retrieval from Vault and setting of the 'Retrieved' attribute logs the Primary Keys returned with a specialized error message ("Possible Data Loss: ${[...Primary Keys]}")

falila:
# Summary | Résumé

Updating the existing Retrieval Api. In this PR we dropped lambda implementation version for a nextJs backend APi.
close #481

1. command `curl http://localhost:3000/api/retrieval?maxRecords=10&formID=1
   -H "Accept: application/json"
   -H "Authorization: Bearer {token}"`
2. Hit the url (GET) and it should return form's responses.

AWS documentation : 
"A Query operation can retrieve a maximum of 1 MB of data. This limit applies before the filter expression is evaluated."
Therefore, using  filterExpression like `FilterExpression: "FormID = :formID",` to select from responses might rise  some questions from the performance stand point.  As state in AWS documentation  " A filter expression is applied after a Query finishes, but before the results are returned. Therefore, a Query consumes the same amount of read capacity, regardless of whether a filter expression is present".

# Pull Request Checklist


pages/api/retrieval.ts
```diff
@@ -0,0 +1,112 @@
...
+    } catch (error) {
+      //logging submissionIDs
+      logMessage.warn(`Possible data loss for these submissionIDs`);
+      for (const subID in submissionIDlist) logMessage.warn(subID);
+      logMessage.error(error);
+      return res.status(500).json({ responses: "Error on Server Side" });
+    }
+  }
+}
+export default isRequestAllowed(["GET"], validate(retrieval));
```
bryan-robitaille, line 112: Unfortunately the `validate()` wrapper will not correctly evaluate the token on this route.  The `validate()` function checks against the bearer token however this route receives the temporary token which contains the `email

bryan-robitaille: A new function will need to be written that takes the temporary token and checks against the database to see if:
1. The token can be verified: `jwt.verify(()` and contains an email address
2. The token has not expired
2. The token exists in the Database
3. The token record retrieved from the database includes the email, active, and template_id fields
4. The email on the token record matches the email in the jwt.verify() payload.
5. The active flag boolean is true.

pages/api/retrieval.ts
```diff
@@ -0,0 +1,112 @@
...
+ * @param req
+ * @param res
+ */
+export async function getFormResponses(
+  req: NextApiRequest,
+  res: NextApiResponse,
+  bearerTokenPayload: BearerTokenPayload
+): Promise<void> {
+  const { maxRecords } = req.query;
+  if (!maxRecords || parseInt(maxRecords as string) < 0 || parseInt(maxRecords as string) >= 11) {
```
bryan-robitaille, line 43: For readability the parseInt could be extracted into a line above so the type cast and parsing only has to be done once.
Just quickly in my head I was thinking something along the lines of:
```js
// Set a fallback default value if undefined outside of the accepted scope and rename
const { maxRecords: maxRecordsStr="-1"} = req.query;
const maxRecords = parseInt(maxRecordsStr as string);
if(maxRecords > 0 && maxRecords <= 11){
  ...
}
```

pages/api/retrieval.ts
```diff
@@ -0,0 +1,112 @@
...
+    return res.status(400).json({ error: "Invalid paylaod value found maxRecords" });
+  } else {
```
bryan-robitaille, line 45: If you'd like the `else {}` block could be removed as the `return` in the if statement above short circuits the function.

bryan-robitaille (changes requested):
Nice work!  Well commented code :)
Once the logic for the bearer token is swapped out for the temporary token acceptance logic then we should be pretty close.

pages/api/retrieval.ts
```diff
@@ -0,0 +1,112 @@
...
+              ":submission": { S: submissionID },
+            },
+            ReturnValues: "NONE",
+          };
+          //Update one item at the time
+          await db.send(new UpdateItemCommand(updateItem));
+        }
+      }
+      //Return form data
+      return res.status(200).json({ responses: formResponses.Items });
```
Moro-Code, line 102: It would probably be better to flatten our JSON structure as opposed to having 

```JSON
 {
  "someKey": {
      "S": "value"
}
}
```

You can probably do this via a map function

tests/api/retrieval.test.js
```diff
@@ -0,0 +1,262 @@
...
+      if (sql.includes("SELECT bearer_token as bearerToken FROM templates")) {
+        return {
+          rows: [{ bearerToken: token }],
+          rowCount: 1,
+        };
+      }
+    });
+    const dynamodbExpectedReponses = {
+      Items: [
+        { FormID: "1", SubmissionID: "12", FormSubmission: true },
```
Moro-Code, line 157: Is that how Dynamo returns responses or is there an extra layer of "S" and "N" to denote types. The unit test should probably reflect this so its as close to real life as possible

Moro-Code (changes requested):
Hey good work. I just had some additional comments to what Bryan already stated

pages/api/retrieval.ts
jeberhardt: What are the magic numbers `0` and `11`?

pages/api/retrieval.ts
```diff
@@ -0,0 +1,112 @@
...
+    } catch (error) {
+      //logging submissionIDs
+      logMessage.warn(`Possible data loss for these submissionIDs`);
+      for (const subID in submissionIDlist) logMessage.warn(subID);
```
jeberhardt, line 106: It is possible that other log message could get into the log during this `for` loop. Maybe construct a string in the loop, and then log out the entire string at once.

tests/api/retrieval.test.js
```diff
@@ -0,0 +1,262 @@
...
+        "Content-Type": "application/json",
+        Origin: "http://localhost:3000/api/retrieval?maxRecords=10",
+        authorization: `Bearer ${token}`,
+      },
+      query: {
+        maxRecords: "10",
+      },
+    });
+    executeQuery.mockImplementation((client, sql) => {
+      if (sql.includes("SELECT bearer_token as bearerToken FROM templates")) {
```
jeberhardt, line 246: This will need to be changed to `'SELECT bearer_token as "bearerToken" FROM templates'` when merged with the PR that I currently have up.

falila: I've written a new validator where i no longer use this query.

pages/api/retrieval.ts
falila: Your remarque is correct. Now using the toString method instead.

pages/api/retrieval.ts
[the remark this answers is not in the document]
falila: I'll assume the default is : `10`

pages/api/retrieval.ts
falila: To keep this PR short , I'll address this improvement in a separate card.

lib/types.ts
```diff
@@ -219,4 +219,12 @@ export type UploadResult = {
 };
 export interface BearerTokenPayload {
   formID?: string;
+  email?: string;
+}
+
+export interface FormUserDBConfigProperties {
```
jeberhardt, line 225: Typescript variables should be in camelCase. I know this doesn't line up with the naming convention in the database, so there will need to be a way to reconcile the differences, but I think the interface should be camelCase.

pages/api/retrieval.ts
```diff
@@ -0,0 +1,170 @@
...
+ * This method fetchs form responses up to the maximum number (maxRecords) that was specified.
+ * It required to have a valid bearer token which must carry a valid formID. When all these conditons are met,
+ * Not only it should return form reponses but also mark those responses as retrieved by updating the attribute Retrieved
+ * from 0 to 1.
+ * @param res 
+ * @param formID 
+ * @param maxRecords 
+ * @returns 
+ */
+export async function getFormResponses(
```
jeberhardt, line 28: Since this is exported as `default` on Line 170, it doesn't need to be exported here.

pages/api/retrieval.ts
```diff
@@ -0,0 +1,170 @@
...
+ * @param handler - A method to retrieve a form responses
+ * @returns
+ */
+export const formResponsesReqValidator = (
+  handler: (res: NextApiResponse, formID: string, maxRecords: number) => void
+) => {
+  return async function (req: NextApiRequest, res: NextApiResponse): Promise<unknown> {
+    try {
+      //Default value to 10 if it's undefined
+      const { maxRecords = 10, formID } = req.query;
```
jeberhardt, line 119: Why is this default set as a number, and then on line 123 `as string` is being called? Seems unnecessary.

lib/types.ts
falila: I see but the entire project's interface follow this naming convention.  Maybe are you suggesting `FormUserDdConfigProperties` instead. Or am i missing something ?

tests/api/retrieval.test.js
```diff
@@ -0,0 +1,324 @@
...
+
+describe("/api/retrieval", () => {
+  beforeAll(() => {
+    process.env.TOKEN_SECRET = "gc-form-super-secret-code";
+  });
+  afterAll(() => {
+    delete process.env.TOKEN_SECRET;
+  });
+
+  it("Should return 400 status code b/c formID is undefined", async () => {
```
jeberhardt, line 33: `formID`, for this request as defined on Line 59, is an empty string, not `undefined`.

pages/api/retrieval.ts
falila: The reasoning behind it is that the parameter `maxRecord` may not exist in query string. In that case, we have to pass the default value as string since `parsInt` accepts string. 
On the other hand, if maxRecords exists but the value is undefined its default value will be 10. And keep in mind that the type of a query parameter is `string | string []`.

tests/api/retrieval.test.js
```diff
@@ -0,0 +1,324 @@
...
+      template_id: 1,
+      email: "forms@cds.ca",
+      temporary_token: token,
+      active: true,
+    };
+
+    executeQuery.mockImplementation((client, sql) => {
+      if (
+        sql.includes(
+          "SELECT * FROM form_users WHERE template_id = ($1) and email = ($2) and active = true"
```
jeberhardt, line 140: Better to specify the fields to return than to request *

tests/api/retrieval.test.js
falila: Correct. Will update the comment and a new test to cover `undefined` case as well.

tests/api/retrieval.test.js
falila: `active` field.

pages/api/retrieval.ts
falila: By the i could have set the default as string as well. but it doesn't remove the fact that we need to cast it as string.

falila:
@jeberhardt  Thank  you for your feedbacks . I've integrated must of them. If you feel like i've miss something please let me known.

lib/types.ts
bryan-robitaille: James in referring to `template_id` instead of `templateID` and `temporary_token` instead of `temporaryToken`

pages/api/retrieval.ts
bryan-robitaille: The query paramerter is typed as `string | string[]` because if a parameter is repeated in the query they are lumped into an array.  Since in our case we only want formID and maxRecords to show up once in the query we can typecheck using something like:
```
      if (Array.isArray(formID) || Array.isArray(maxRecords))
        return res.status(400).json({ error: "Bad Request" });
```
And since the type of maxRecords has been checked and is not `string[]` it is no longer necessary to typecast.
``` 
const{ maxRecords = "10", formID} = req.query;
//...
// maxRecords will either be a default "10" or be the string value from the req.query
const expectedMaxRecords = parseInt(maxRecords)
```

pages/api/retrieval.ts
```diff
@@ -0,0 +1,171 @@
...
+      const bearerTokenPayload = jwt.verify(
+        token,
+        process.env.TOKEN_SECRET || ""
+      ) as BearerTokenPayload;
+      const { email } = bearerTokenPayload;
+      //Check if a formUserRecord exists for the given bearerToken.
+      const formUserRecord = (
+        await getFormUserRecordByFormIDAndEmail(formID as string, email as string)
+      ).rows[0] as FormUserDBConfigProperties;
+      if (formUserRecord?.active) {
```
bryan-robitaille, line 141: If a record is returned `formUserRecord?.active` will always be true because of the SQL query on line 168 `"SELECT active FROM form_users WHERE template_id = ($1) and email = ($2) and active = true"`.  There might be a way to refactor this so that formUserRecord is checked to see if a result is returned instead.  This would also enable the FormUserDBConfigProperties to not have each property as optional but instead required.

pages/api/retrieval.ts
```diff
@@ -0,0 +1,171 @@
...
+        //Moving forward to fetch form's responses
+        return handler(res, formID as string, expectedMaxRecords);
```
bryan-robitaille, line 143: After adding the typecheck for `string[]` earlier in this function it won't be necessary to typecase formID as `string`.

pages/api/retrieval.ts
```diff
@@ -0,0 +1,171 @@
...
+ * @param maxRecords 
+ * @returns 
+ */
+async function getFormResponses(
+  res: NextApiResponse,
+  formID: string,
+  maxRecords: number
+): Promise<void> {
+  //A list to store submissionIDs
+  let submissionIDlist: (string | undefined)[] | undefined = [];
```
bryan-robitaille, line 35: We can add a type check later on in the code to remove the `| undefined` which helps with readability.

pages/api/retrieval.ts
```diff
@@ -0,0 +1,171 @@
...
+      //You need to specify those attributes in the key condition expression, not the filter expression.
+      FilterExpression: "FormID = :formID",
+      ProjectionExpression: "FormID,SubmissionID,FormSubmission,Retrieved",
+    };
+    //Get form's responses for formID
+    const formResponses = await db.send(new QueryCommand(getItemsDbParams));
+    //Collecting items submissionIDs for logging and updating items.
+    submissionIDlist = formResponses?.Items?.map((response) => {
+      return response.SubmissionID.S;
+    });
```
bryan-robitaille, line 59: We can add a coalescing operator to check if null/undefined and return an empty array instead of `undefined`.
```
    submissionIDlist =
      formResponses.Items?.map((response) => {
        return response.SubmissionID.S;
      }) ?? [];
```

This basically does the type check and simplifies the variable typing.

pages/api/retrieval.ts
falila: @bryan-robitaille  
Yeah the returned will always be active. I tried to avoid returning the hole row (unnecessary data)  but just  enough data.
Now i'm leaning towards this query  `SELECT EXISTS(SELECT 1 FROM form_users WHERE template_id = ($1) and email = ($2) and active = true)` which will return true or false so we can get rid of the entire formUSerRecords type.
OR 
`"SELECT  *  FROM form_users WHERE template_id = ($1) and email = ($2) and active = true"`  and make all formUserRecord's properties required. Then check if it is defined. 

What is your take on it ?

bryan-robitaille: I like the `SELECT EXISTS` and I see you've implemented it 👍

bryan-robitaille (dismissed):

pages/api/retrieval.ts
```diff
@@ -0,0 +1,166 @@
...
+ * It returns an active record from form_users table base upon a formID/template_id and an email.
+ * @param formID - The id of the form
+ * @param email - The email that is associated to the formID
+ * @returns A form_user's record with active field i.e [{ active: true }]
+ */
+const checkActiveFormUserRecord = async (formID: string, email: string): Promise<QueryResult> => {
+  //Retrieving a tokenRecord or return an empty array
+  return executeQuery(
+    await dbConnector(),
+    "SELECT EXISTS(SELECT 1 FROM form_users WHERE template_id = ($1) and email = ($2) and active = true)",
```
Moro-Code, line 161: Rather than doing a nested query like this . Just do the actual query and if there is a row return true otherwise false

Moro-Code (changes requested):
A little bit confused here. You're still not validating the temporary token at all just the bearer token. Please have a look at previous review comments

pages/api/retrieval.ts
```diff
@@ -0,0 +1,169 @@
...
+ *  - Must contains an email address
+ *  - It has not expired
+ *  - It exists in our Database
+ *  - The record found in the database includes the email, active, and template_id fields
+ *  - The email on the token record matches the email in the payload.
+ *  - The active flag boolean is true.
+ * @param handler - A method to retrieve form's responses
+ * @returns
+ */
+export const formResponsesReqValidator = (
```
Moro-Code, line 114: This should have been in a separate file as with mostly all middleware functions for reusability

lib/types.ts
```diff
@@ -219,4 +219,5 @@ export type UploadResult = {
 };
 export interface BearerTokenPayload {
```
Moro-Code, line 220: Why are we using the same Interface for the bearer token and temporary token. We should ideally have separate interfaces as opposed to making all fields within the interface optional. This makes it clear what we expect from each

lib/middleware/bearerToken.ts
```diff
@@ -41,7 +41,7 @@ const validate = (
  * @throws
  * This exception is thrown if the bearer token is not found
  */
-const getBearerToken = (req: NextApiRequest) => {
+export const getBearerToken = (req: NextApiRequest): string => {
```
Moro-Code, line 44: This could probably be better names from getBearerToken to `extractTokenFromAuth` or something along those lines to make it clear what exactly it is doing.

pages/api/retrieval.ts
```diff
@@ -0,0 +1,169 @@
...
+    //Get form's responses for formID
+    const formResponses = await db.send(new QueryCommand(getItemsDbParams));
+    //Collecting items submissionIDs for logging and updating items.
+    submissionIDlist =
+      formResponses?.Items?.map((response) => {
+        return response.SubmissionID.S;
+      }) ?? [];
+    if (formResponses.Count && formResponses.Count > 0) {
+      //Preparing a query to set records's attribute Retrieved to 1.
+      for (const submissionID of submissionIDlist as string[]) {
```
Moro-Code, line 62: I would have probably separated this out in a separate try catch block so its clear exactly where the error happened either when retrieving the rows or updating the retrieved status. At the moment it wouldn't be so clear from the error messages

pages/api/retrieval.ts
```diff
@@ -0,0 +1,169 @@
...
+ */
+export const formResponsesReqValidator = (
+  handler: (res: NextApiResponse, formID: string, maxRecords: number) => void
+) => {
+  return async function (req: NextApiRequest, res: NextApiResponse): Promise<unknown> {
+    try {
+      //Default value to 10 if it's undefined
+      const { maxRecords = "10", formID } = req.query;
+      //Check that formID and maxRecords aren't repeated
+      if (Array.isArray(formID) || Array.isArray(maxRecords))
```
Moro-Code, line 122: If we wanted to be as modular as possible this function should just have been validating whether or not the temporary token is valid

pages/api/retrieval.ts
```diff
@@ -0,0 +1,169 @@
...
+      //Get the token from request object
+      const token = getBearerToken(req);
+      //Verify the token
+      const bearerTokenPayload = jwt.verify(
+        token,
+        process.env.TOKEN_SECRET || ""
+      ) as BearerTokenPayload;
+      const { email } = bearerTokenPayload;
+      //Check if an active formUserRecord exists for the given bearerToken.
+      if (await isTokenExists(formID, email as string, token))
```
Moro-Code, line 141: If it's going to be wrapped around two lines anyways then this doesn't really help readability by not using brackets

tests/api/retrieval.test.js
```diff
@@ -0,0 +1,339 @@
...
+describe("/api/retrieval", () => {
+  beforeAll(() => {
+    process.env.TOKEN_SECRET = "gc-form-super-secret-code";
+  });
+  afterAll(() => {
+    delete process.env.TOKEN_SECRET;
+  });
+
+  it("Should return 400 status code b/c formID is an empty string", async () => {
+    const mockSession = {
```
Moro-Code, line 34: Do we need to mock a session for a get request ?

tests/api/retrieval.test.js
```diff
@@ -0,0 +1,339 @@
...
+        maxRecords: "10",
+        formID: undefined,
+      },
+    });
+    await retrieval(req, res);
+    expect(res.statusCode).toBe(400);
+    expect(JSON.parse(res._getData())).toEqual(expect.objectContaining({ error: "Bad Request" }));
+  });
+
+  it("Should return 400 status code with Invalid paylaod value found maxRecords ", async () => {
```
Moro-Code, line 101: The test naming here is kind of confusing

Moro-Code (dismissed):
If we had more time there would be more changes that are needed here. However, in the interest of time since this PR has been open for approximately 18 days its better to get this in and fix after the fact. I have however documented some of the things that should be addressed at some point for your reference are future reference should someone be working on this code

tests/api/retrieval.test.js
falila: Done!

tests/api/retrieval.test.js
falila: No. Unnecessary or unused code was removed.

lib/middleware/bearerToken.ts
falila: Renamed  as `extractBearerTokenFromReq`

pages/api/retrieval.ts
falila: Done!

lib/types.ts
falila: I've added `TemporaryTokenPayload` interface.

pages/api/retrieval.ts
falila: Done! It was renamed as  `checkIfValidTemporaryToken`

pages/api/retrieval.ts
falila: fixed

pages/api/retrieval.ts
falila: Done!

pages/api/retrieval.ts
```diff
@@ -0,0 +1,117 @@
...
+        await db?.send(new UpdateItemCommand(updateItem));
+      }
+    }
+    //Return responses data
+    return res.status(200).json({ responses: formResponses.Items });
+  } catch (error) {
+    logMessage.warn(`Possible data loss for these submissionIDs`);
+    //print submission ids as string
+    logMessage.warn(submissionIDlist ? submissionIDlist.toString() : "Empty submissionID list");
+    logMessage.error(error);
```
jeberhardt, line 111: The above logs could be concatenated into one string before logged. With a busy system, it's possible that other `logMessage` commands could be executed by another client, and these messages could be split up.

Moro-Code (approved):
Looks fine to me atm

tests/api/retrieval.test.js
```diff
@@ -0,0 +1,312 @@
...
+        return {
+          rows: [{ column: 1 }],
+          rowCount: 1,
+        };
+      }
+    });
+    await retrieval(req, res);
+    expect(res.statusCode).toBe(400);
+    expect(JSON.parse(res._getData())).toEqual(
+      expect.objectContaining({ error: "Invalid paylaod value found maxRecords" })
```
jeberhardt, line 124: Type-o in `paylaod`. Should be `payload`.

pages/api/retrieval.ts
```diff
@@ -0,0 +1,117 @@
...
+  formID: string
+): Promise<void> {
+  //Default value to 10 if it's undefined
+  const { maxRecords = "10" } = req.query;
+  //Check maxRecords aren't repeated
+  if (Array.isArray(maxRecords) || !formID) return res.status(400).json({ error: "Bad Request" });
+  const expectedMaxRecords = parseInt(maxRecords);
+  //Range is 1- 10
+  if (expectedMaxRecords < 1 || expectedMaxRecords > 10) {
+    return res.status(400).json({ error: "Invalid paylaod value found maxRecords" });
```
jeberhardt, line 41: Type-o in paylaod. Should be payload.

tests/api/retrieval.test.js
```diff
@@ -0,0 +1,312 @@
...
+        "Content-Type": "application/json",
+        Origin: "http://localhost:3000/api/retrieval?maxRecords=1",
+        authorization: `Bearer ${token}`,
+      },
+      query: {
+        maxRecords: "11",
+        formID: "22",
+      },
+    });
+    executeQuery.mockImplementation((client, sql) => {
```
jeberhardt, line 109: This mockImplementation for this query should never be called in this test, because of the early return and 400 error.

tests/api/retrieval.test.js
```diff
@@ -0,0 +1,312 @@
...
+
+    executeQuery.mockImplementation((client, sql) => {
+      if (
+        sql.includes(
+          "SELECT 1 FROM form_users WHERE template_id = ($1) and email = ($2) and temporary_token = ($3) and active = true"
+        )
+      ) {
+        return {
+          rows: [],
+          rowCount: 1,
```
jeberhardt, line 216: Is `rowCount` really **1** in this case?

PR #597, opened 2022-01-20
Fix JSON component refresh after update

bryan-robitaille:
# Summary | Résumé

This PR creates a new react hook that enables the refreshing of the serverSideProps passed into the NextJS page.

Also renamed the other hooks to align with React Hooks naming convention `use___()`

Make a change in the JSON editor through `/en/id/{formID}/settings` and upload the changes.  After upload is successful the JSON editor now displays the new version of the template file instead of the previously fetched version.

# Pull Request Checklist


components/admin/JsonUpload/JsonUpload.test.js
```diff
@@ -38,10 +43,16 @@ describe("JSON Upload Component", () => {
...
+    useRouter.mockImplementation(() => ({ replace: jest.fn(), push: jest.fn() }));
+
     render(<JSONUpload form={form}></JSONUpload>);
     await act(async () => {
-      await fireEvent.click(screen.queryByTestId("upload"));
+      fireEvent.click(screen.queryByTestId("upload"));
+    });
+
+    waitFor(() => {
+      expect(screen.queryByTestId("submitStatus")).toBeInTheDocumen;
```
jeberhardt, line 55: missing a `t` on the end of `toBeInTheDocumen`

jeberhardt: Kind of surprised that the test passed.

jeberhardt:
I'm wondering if we can just update the page with the return value from the api call to `/api/templates`, instead of calling `router.replace`.  The return value might need to be updated to include all the form data.

components/admin/JsonUpload/JsonUpload.test.js
bryan-robitaille: Good catch thanks :). The test was passing because waitFor() is async and wasn't being awaited.  Working on that now.

craigzour (approved):
LGTM!

components/admin/JsonUpload/JsonUpload.test.js
jeberhardt: With a last name like mine, I will always notice if the end of a word is missing a `t`.

bryan-robitaille:
> I'm wondering if we can just update the page with the return value from the api call to `/api/templates`, instead of calling `router.replace`. The return value might need to be updated to include all the form data.

Great suggestion.   Here's the reasoning behind why I went with this solution instead.

bryan-robitaille:
> I'm wondering if we can just update the page with the return value from the api call to `/api/templates`, instead of calling `router.replace`. The return value might need to be updated to include all the form data.

Great Suggestion.  Here's the reasoning behind this refresh strategy.
The form config is passed into this component as a prop.  Best practice in React is that props are immutable.  If we follow the chain of components up to the top the state for the form config is passed in through serverSideProps (creates a JSON that is used as the initial state when react hydrates client side).  The downside of doing the fetch to the `api/templates` endpoint and modifying the form config state directly on the component means that it's now out of sync with the parent component and other potential components on the page  that leverage the form config.
If we really wanted to do a heavy refactor and take a more React approach we could drop the serverSideProps call for the initial form config,  and instead build a hook that creates a React.Context at the page level and manages the form config state.  We'd also need to leverage some sort of <Suspense /> component during api calls.

So I guess the bigger question is: Do we want this project to behave more like a 'React' app or more like a 'server side rendered' app.  NextJS allows for both....

@Moro-Code thoughts?

Moro-Code (approved):
Code looks good. I think this is a good approach in terms of refreshing the page as opposed to using state to populate the field @bryan-robitaille @jeberhardt . We're using NextJS for its SSR capability. I think our next big refactoring will present itself when we finally update to NextJS 12 and are able to use server side components. For now I think this is the simplest solution

PR #658, opened 2022-02-25
Add timer to prevent submission of form.

jeberhardt:
# Summary | Résumé

1-3 sentence description of the changed you're proposing.
Please include the GitHub issue number or relevant acceptance criteria.

To specify that this PR closes a particular issue you can use the following format
to reference and issue `closes #[issue number]`. You can use any number of keywords to link
an issue to a PR. Please see the [following section](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
of GitHub's documentation.

| 
Screen Shot 2022-02-25 at 6 33 08 AM
 |
Screen Shot 2022-02-25 at 6 32 59 AM
  |

links in this section

# Pull Request Checklist

### DIFF
--- a/components/forms/Button/Button.tsx
+++ b/components/forms/Button/Button.tsx
@@ -25,6 +25,7 @@ export const Button = (
       "gc-button--secondary": secondary,
       "gc-button--base": base,
       "gc-button--destructive": destructive,
+      "gc-button--disabled": disabled,
     },
     className
   );
--- a/components/forms/Form/Form.tsx
+++ b/components/forms/Form/Form.tsx
@@ -27,6 +27,7 @@ const InnerForm = (props: InnerFormProps & FormikProps<FormValues>) => {
   const errorId = "gc-form-errors";
   const serverErrorId = `${errorId}-server`;
   const formStatusError = props.status === "Error" ? t("server-error") : null;
+  const [formSubmitTimeout, setFormSubmitTimeout] = useState(false);
 
   //  If there are errors on the page, set focus the first error field
   useEffect(() => {
@@ -44,6 +45,17 @@ const InnerForm = (props: InnerFormProps & FormikProps<FormValues>) => {
       setCanFocusOnError(false);
     }
   }, [formStatusError, errorList, lastSubmitCount, canFocusOnError]);
+
+  useEffect(() => {
+    // timeout to prevent form from being submitted too quickly
+    const millisPerFormElement = 2000;
+    setTimeout(
+      () => {
+        setFormSubmitTimeout(true);
+      },
+      children ? React.Children.count(children) * millisPerFormElement : 0
+    );
+  }, [formSubmitTimeout]);
 
   return (
     <>
@@ -76,7 +88,9 @@ const InnerForm = (props: InnerFormProps & FormikProps<FormValues>) => {
           >
             {children}
             <div className="buttons">
-              <Button type="submit">{t("submitButton")}</Button>
+              <Button type="submit" disabled={!formSubmitTimeout}>
+                {t("submitButton")}
+              </Button>
             </div>
           </form>
         </>
--- a/components/forms/Form/Form.test.js
+++ b/components/forms/Form/Form.test.js
@@ -7,6 +7,8 @@ jest.mock("../../../lib/integration/helpers", () => ({
   submitToAPI: jest.fn(() => {}),
 }));
 
+jest.useFakeTimers();
+
 const formConfig = {
   id: 1,
   version: 1,
@@ -42,12 +44,32 @@ describe("Generate a form component", () => {
 
     it("Form is submitted", async () => {
       await act(async () => {
-        await render(<Form formConfig={formConfig} language="en" t={(key) => key}></Form>);
+        render(<Form formConfig={formConfig} language="en" t={(key) => key}></Form>);
       });
+
+      const submitButton = screen.getByRole("button", { type: "submit" });
+
+      // complete the timeout to allow the form to be submitted
+      jest.runAllTimers();
+
       await act(async () => {
-        await screen.getByRole("button", { type: "submit" }).click();
+        submitButton.click();
       });
+      expect(submitButton).not.toBeDisabled();
       expect(mockedSubmitFunction).toBeCalledTimes(1);
+    });
+
+    it("Form is not submitted because not enough time has passed", async () => {
+      await act(async () => {
+        render(<Form formConfig={formConfig} language="en" t={(key) => key}></Form>);
+      });
+
+      const submitButton = screen.getByRole("button", { type: "submit" });
+      await act(async () => {
+        submitButton.click();
+      });
+      expect(submitButton).toBeDisabled();
+      expect(mockedSubmitFunction).not.toHaveBeenCalled();
     });
   });
 });
--- a/styles/_forms.scss
+++ b/styles/_forms.scss
@@ -210,6 +210,13 @@
   }
 }
 
+.gc-button--disabled {
+  @apply bg-gray-800;
+  &:hover {
+    @apply bg-gray-800;
+  }
+}
+
 .gc-label {
   @apply md:text-small_base block mb-2 text-base font-bold w-3/4 xl:w-full;
   &.required > span {


components/forms/Form/Form.tsx
```diff
@@ -45,6 +46,20 @@ const InnerForm = (props: InnerFormProps & FormikProps<FormValues>) => {
...
+    const formSubmitTimeout = setTimeout(
+      () => {
+        setFormSubmitDelay(true);
+      },
+      children ? React.Children.count(children) * millisPerFormElement : 0
+    );
+    return () => {
+      clearTimeout(formSubmitTimeout);
+    };
+  }, [formSubmitDelay]);
```
bryan-robitaille, line 61: With `formSubmitDelay` added as a useEffect() dependency it will actionally run 1 more time then you might expect.
- Inital form render, useEffect runs and create setTimeout callback, formSubmitDelay state = false
- Timer expires and executes callback, formSubimtDelay state = true
- UseEffect dependency detected and checks if previous value = new value.  false = true so run useEffect again.
- useEffect runs and creates setTimeout callback, formSubmitDelay state = true
- Timer expires and executre callback, formSubmitDelay state = true
- UseEffect dependency detected and checks if previous value = new value.  true = true so do not run useEffect again.

This behavior can be avoided by instead passing an empty dependnecy array `useEffect(()=> (), [] )`
An empty dependency array tells useEffect to only run on the initial render and never again.

For more info: https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects

components/forms/Form/Form.tsx
```diff
@@ -45,6 +46,20 @@ const InnerForm = (props: InnerFormProps & FormikProps<FormValues>) => {
...
+    // timeout to prevent form from being submitted too quickly
+    const millisPerFormElement = 2000;
+    const formSubmitTimeout = setTimeout(
+      () => {
+        setFormSubmitDelay(true);
+      },
+      children ? React.Children.count(children) * millisPerFormElement : 0
+    );
+    return () => {
+      clearTimeout(formSubmitTimeout);
```
bryan-robitaille, line 59: Nice use of a cleanup function.

components/forms/Form/Form.tsx
```diff
@@ -45,6 +46,20 @@ const InnerForm = (props: InnerFormProps & FormikProps<FormValues>) => {
...
   }, [formStatusError, errorList, lastSubmitCount, canFocusOnError]);
 
+  useEffect(() => {
+    // timeout to prevent form from being submitted too quickly
+    const millisPerFormElement = 2000;
+    const formSubmitTimeout = setTimeout(
+      () => {
+        setFormSubmitDelay(true);
+      },
+      children ? React.Children.count(children) * millisPerFormElement : 0
```
bryan-robitaille, line 56: A possible enhancement would be to filter out rich text components and only use question components to set the timer.

bryan-robitaille (changes requested):
Great start!
A few small tweaks and I think this PR is ready in terms of functionality.
Then is just designs and copy on how to provide that context to the user.

components/forms/Form/Form.tsx
jeberhardt: Yeah, that's in my comments for this PR. I wasn't sure how accurate the timer needed to be. I also wasn't sure if there were more elements than `richText` elements that should be filtered out.

bryan-robitaille: Sorry, I didn't see that question earlier in the 'unresolved questions'.
For now I believe the only elements that aren't interactive are richText and heading.  For this PR it would be nice if we could filter them out as on a lot of the newer forms we're seeing lot's of content in those 2 element types.

craigzour:
@jeberhardt I see that you have requested reviews but it is still in a Draft state. Are you still making some adjustments?

jeberhardt:
> @jeberhardt I see that you have requested reviews but it is still in a Draft state. Are you still making some adjustments?

@craigzour - Yes, still making changes. I had it "Ready for Review", but then there were some changes to the requirements, and I marked it as **Draft**. I'll let everyone know when it's ready.

craigzour:
@jeberhardt found a small bug while testing. If I pressed the submit button and the timer is displayed it gets reset to 5 seconds even if there was 1 second left.

craigzour:
> @jeberhardt found a small bug while testing. If I press the submit button and the timer is displayed it gets reset to 5 seconds even if there was 1 second left.

Looks like it is not a bug now that I am looking at the code :)

components/forms/Button/Button.tsx
```diff
@@ -25,6 +25,7 @@ export const Button = (
       "gc-button--secondary": secondary,
       "gc-button--base": base,
       "gc-button--destructive": destructive,
+      "gc-button--disabled": disabled,
```
craigzour, line 28: Not sure if  this is even used with your implementation. Could it be code that you had early in the development but ended up changing strategy?

components/forms/Form/Form.tsx
```diff
@@ -45,6 +49,39 @@ const InnerForm = (props: InnerFormProps & FormikProps<FormValues> & DynamicForm
...
+        );
+      } catch {
+        return 0;
+      }
+    }).length;
+
+    const submitDelaySeconds = secondsBaseDelay + numberOfRequiredElements * secondsPerFormElement;
+    setSubmitDelay(submitDelaySeconds);
+    setSubmitTimer(submitDelaySeconds);
+  }, [submitDelay]);
```
craigzour, line 70: Feel like this `useEffect` should be set up in a run only once mode. `}, []);`

components/forms/Button/Button.tsx
jeberhardt: That's exactly what happened. I'll remove it.

components/forms/Form/Form.tsx
```diff
@@ -45,6 +49,39 @@ const InnerForm = (props: InnerFormProps & FormikProps<FormValues> & DynamicForm
...
+    const secondsBaseDelay = 2;
+    const secondsPerFormElement = 2;
+    const numberOfRequiredElements = React.Children.toArray(children).filter((child) => {
+      try {
+        return (
+          ((child as ReactElement).props as GenerateElementProps).element.properties.validation
+            ?.required == true
+        );
+      } catch {
+        return 0;
```
craigzour, line 63: `return false` would be more suitable in a `filter` statement :)

components/forms/Form/Form.tsx
```diff
@@ -45,6 +49,39 @@ const InnerForm = (props: InnerFormProps & FormikProps<FormValues> & DynamicForm
     }
   }, [formStatusError, errorList, lastSubmitCount, canFocusOnError]);
 
+  useEffect(() => {
+    // calculate initial delay for submit timer
+    const secondsBaseDelay = 2;
+    const secondsPerFormElement = 2;
+    const numberOfRequiredElements = React.Children.toArray(children).filter((child) => {
+      try {
```
craigzour, line 57: I would not know (right now) how to simplify this code but I have this feeling that a cleaner solution could exist 😕 
Maybe someone else in the team will propose something 👀

craigzour (approved):
LGTM! Maybe wait for a second review before merging?

craigzour (approved):

bryan-robitaille (dismissed):
Refactored and modified a few small things.  I was going through the logic on Forms.tsx and I originally though it could be simpler but ran into some roadblocks that I'm sure you encountered as well.  Looking forward to potentially revisiting in the future to refine and refactor a little more.

components/forms/Form/Form.tsx
```diff
@@ -45,6 +49,36 @@ const InnerForm = (props: InnerFormProps & FormikProps<FormValues> & DynamicForm
...
+    const submitDelaySeconds = secondsBaseDelay + numberOfRequiredElements * secondsPerFormElement;
+    setSubmitDelay(submitDelaySeconds);
+    setSubmitTimer(submitDelaySeconds);
+  }, [formConfig]);
+
+  useEffect(() => {
+    // timeout to prevent form from being submitted too quickly
+    let timeoutId = 0;
+    if (submitTimer > 0) {
+      timeoutId = window.setTimeout(() => {
```
bryan-robitaille, line 73: I modified to `window.setTimeout` as it returns a number.  Instead of using `setTimeout()` which returns a type of NodeJS.Timer.  This will never be called server side (aka Node) so we're safe to use `window.setTimeout()` here.  It also saves us from retyping from NodeJS.Timer => unknown => number.

components/forms/Form/Form.tsx
```diff
@@ -45,6 +49,36 @@ const InnerForm = (props: InnerFormProps & FormikProps<FormValues> & DynamicForm
...
 
+  useEffect(() => {
+    if (!formConfig) {
+      return;
+    }
+
+    // calculate initial delay for submit timer
+    const secondsBaseDelay = 2;
+    const secondsPerFormElement = 2;
+    const numberOfRequiredElements = formConfig.elements.filter(
```
bryan-robitaille, line 60: I refactored a bit because `(element)` will never be 'undefined'

components/forms/Form/Form.tsx
```diff
@@ -1,12 +1,13 @@
 import React, { useEffect, useState } from "react";
```
bryan-robitaille, line 1: Removed unused imports.

bryan-robitaille:
Sorry for the follow up with the comments.  They were originally all part of my review but by pushing commits during my review (the refactorings) it somehow put my review into an in-between stale state.

components/forms/Form/Form.tsx
craigzour: That is correct! @jeberhardt  I saw you added this as part of a "Testing fix" commit so I thought you would end up removing it 😆

bryan-robitaille:
> It would have probably been better to add this behind a feature flag as discussed. On top of not being able to disable this for local development our github actions will take longer as parallelism will be disabled and the wait commands.

Agreed for the feature flag.  I can quickly add.
Our Cypress tests on github were never actually running in parallel as it requires a Cypress Dashboard account and a matrix strategy to be configured.

craigzour (approved):

components/forms/Form/Form.tsx
jeberhardt: Yes, I added it, because I thought it might have had something to do with why the Cypress tests were failing (since that's the part of the code that I changed). As it turns out, the Cypress tests failing had nothing to do with that, and the null check on `elements` was totally not needed.

craigzour (approved):

bryan-robitaille (approved):
LGTM

PR #661, opened 2022-02-25
fix: Sort the error summary to match the displayed elements.

### LINKED ISSUE #521: Error Summary links in wrong order.
When a form is submitted and and error validation happens, the Error Summary at the top of the page does not reflect the same order as the elements on the form.

To re-create this issue, change the order of the required items in `layout` in the formConfig.

Hover over the links in the error summary, and they will appear in the order of the items as found in the `elements` of the formConfig, irrespective of the order of the layout.

# ACCEPTANCE CRITERIA

[ ] The order of the links in the Error Summary will be in the same order as the way the items appear on the screen.

jeberhardt:
# Summary | Résumé

Closes #521 

This will loop through the list of errors generated by Formik and match them up with the list of elements in the form itself.

The UI changes to this are imperceivable as a screenshot, since it's only the anchor for the link that changed.

- Create a form config where the `layout` ids are not in sequential order. 
- Submit a blank form, generating lots of errors.
- Verify that the links in the error list match and go to the correct form elements.

# Pull Request Checklist

### DIFF
--- a/lib/validation.tsx
+++ b/lib/validation.tsx
@@ -206,8 +206,21 @@ export const getErrorList = (
 ): JSX.Element | null => {
   let errorList;
   const formElementErrors = Object.entries(props.errors);
-  if (props.touched && formElementErrors.length) {
-    errorList = formElementErrors.map(([formElementKey, formElementErrorValue]) => {
+  if (formElementErrors.length === 0) {
+    return null;
+  }
+  if (props.children === undefined) {
+    return null;
+  }
+  const formElements = Object.entries(props.children as Array<unknown>);
+  if (formElements.length === 0) {
+    return null;
+  }
+
+  const formElementErrorsDisplayOrder = sortElementErrors(formElementErrors, formElements);
+
+  if (props.touched && formElementErrorsDisplayOrder.length) {
+    errorList = formElementErrorsDisplayOrder.map(([formElementKey, formElementErrorValue]) => {
       if (Array.isArray(formElementErrorValue)) {
         return formElementErrorValue.map((dynamicRowErrors, dynamicRowIndex) => {
           return Object.entries(dynamicRowErrors).map(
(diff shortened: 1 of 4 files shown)


lib/validation.tsx
```diff
@@ -206,8 +206,21 @@ export const getErrorList = (
...
   let errorList;
   const formElementErrors = Object.entries(props.errors);
-  if (props.touched && formElementErrors.length) {
-    errorList = formElementErrors.map(([formElementKey, formElementErrorValue]) => {
+  if (formElementErrors.length === 0) {
+    return null;
+  }
+  if (props.children === undefined) {
+    return null;
+  }
```
Moro-Code, line 214: These two if statements could probably be combined

lib/validation.tsx
```diff
@@ -206,8 +206,21 @@ export const getErrorList = (
...
+  const formElements = Object.entries(props.children as Array<unknown>);
```
Moro-Code, line 215: I'm not sure this is the best way of doing this. Is there any reason we are not using the layout array in the form config? The formConfig should be available in the props

Moro-Code: If you get the layout array you could do a map on it and return the relevant error if it exists

Moro-Code (changes requested):
Please see suggestion for simpler implementation of this

lib/validation.tsx
jeberhardt: > I'm not sure this is the best way of doing this. Is there any reason we are not using the layout array in the form config? The formConfig should be available in the props

I was hoping that the formConfig would have been available in the props. However, it doesn't seem to be available as an `InnerFormProps` or `FormikProps`. The only way I'd be able to access it is by refactoring some components higher up in the chain, and then passing in the layout. I figured that since the `children` prop was available, that could just as easily be used.

Moro-Code: We can discuss tomorrow. Im fairly certain its available but I could totally be wrong

lib/validation.tsx
```diff
@@ -202,12 +202,29 @@ export const validateOnSubmit = (values: FormValues, props: DynamicFormProps): R
...
+  props: InnerFormProps & FormikProps<FormValues> & DynamicFormProps
 ): JSX.Element | null => {
+  if (!props.formConfig) {
+    return null;
+  }
   let errorList;
-  const formElementErrors = Object.entries(props.errors);
-  if (props.touched && formElementErrors.length) {
-    errorList = formElementErrors.map(([formElementKey, formElementErrorValue]) => {
+  const formElementErrors: [string, string | undefined][] = Object.entries(props.errors);
```
Moro-Code, line 211: Leave the errors as an object instead of tuning it into an array

lib/validation.tsx
```diff
@@ -202,12 +202,29 @@ export const validateOnSubmit = (values: FormValues, props: DynamicFormProps): R
...
+  const layout: string[] = props.formConfig.layout;
+  if (formElementErrors.length === 0 || layout.length === 0) {
+    return null;
+  }
+
+  const sortedFormElementErrors: [string, string | undefined][] = [];
+  layout.map((element) => {
```
Moro-Code, line 218: This way you just need to do the following 

```js
const sortedFormElementErrors = layout.map((element) => {
    return [ element, props.errors[element] ]
})
```

Moro-Code (changes requested):
Almost there.

lib/validation.tsx
jeberhardt: Had to add in a filter as well, otherwise it returned an array that was the full size of `layout`.

Moro-Code (dismissed):
Looks good to go

Moro-Code (approved):

PR #671, opened 2022-03-04
Fix/http methods

bryan-robitaille:
# Summary | Résumé
Modifies the various Axios calls to use the appropriate REST HTTP method instead of relying on data passed through the body of the request.  Current logic was checking both the request method and the body method which could lead to unexpected results.

Also reverts jest back to v26.6.3 as there is a breaking change with v27 that is not compatible with jest-chain.
- Modified the respective middlewares to only check the req.method
- Modified calls to lambda for breaking change.  Translates REST method to previous body method.

JEST tests pass. Manual checks can be done by creating, modifying, and deleting a form template. 

The Organizations endpoing/lambda integration was not patched due to it currently not being in use as well as changes that will be made to the functionality due to leveraging Next Auth.

# Pull Request Checklist


lib/middleware/httpRequestAllowed.ts
```diff
@@ -8,8 +8,7 @@ const isRequestAllowed = (
   return async (req: NextApiRequest, res: NextApiResponse): Promise<unknown> => {
     try {
       const session = await getSession({ req });
-      const requestBody = req.body && Object.keys(req.body).length > 0 ? req.body : undefined;
-      const method = requestBody?.method ? requestBody.method : req.method;
+      const method = req.method ?? "";
```
jeberhardt, line 11: A little outside the scope of this PR, but your change highlights this.

If `method` ends up being "", then it will check for a session and allow the request to proceed, as long as they are logged in.

If `req.method` ends up being "", I think the default behaviour should either be:
- return a 403, or
- assume a `GET` request, not requiring a `session`

jeberhardt (approved):

lib/middleware/httpRequestAllowed.ts
bryan-robitaille: I don't think it's outside the scope of this PR at all.  I like the idea of adding more error handling to the `req.method` check.  I'd recommend throwing an error which then triggers the 500 response.   If `req.method` is undefined it's not the clients issues but instead the middleware was not implemented properly.
I'll modify and update this branch.

lib/integration/crud.ts
```diff
@@ -54,17 +54,17 @@ async function _crudTemplates(payload: CrudTemplateInput): Promise<CrudTemplateR
     switch (payload.method) {
```
jeberhardt, line 54: Should `switch` on `method`, not `payload.method`, and it's defined on line 52.

jeberhardt:
SOOOO much cleaner, using `req.method`, thanks for making this change.

jeberhardt (approved):

PR #685, opened 2022-03-16
Add arg for recaptcha to docker and build stages

bryan-robitaille:
# Summary | Résumé
Fixes the missing Env variable required for the reCaptcha feature.

The browser side variable is static and is required at build time.  It will not be picked up in the server side process.


falila (approved):
🌮

PR #690, opened 2022-03-18
Align Cypress tests with best practices

bryan-robitaille:
# Summary | Résumé

Our Cypress tests were often 'Flaky' and would pass or fail seemingly without reason.
This PR:
- Ensures that current tests follow the pattern of command => assertion
  - The cllick() command is never retried in Cypress because it can manipulate the DOM.  So when using the click() command in our test we need to ensure it is followed up by an assertion that the DOM has changed before progressing with the test.   
- Each test is it's separate scope and test order is irrelevant in the Test Suite.    
- Create Cypress commands to mock:
  - form templates by modifying the NextJS Data JSON object passed to React for Rehydration.
  - Feature flags by intercepting the API request from the browser. 
- Removes the mockedFormFile and AllowMockForms functionality from the code base
- Removes CYPRESS check logic as much as possible from the application code.
 
To properly test locally please ensure that your `.env` file only contains the AWS account credentials and Nextauth_url values.  This is to ensure that when `yarn start:test` is run that it doesn't connect to your local DB or Redis instance.

`yarn build && yarn start:test` and in a second terminal `yarn cypress:headless`

While mocking out the Flags feature I noticed that Redis instances/connections are created in both the Template and Flags libs.  There should be one central connection handler for all app feature requirements.

A look to refactor / review the Jests tests for Templates API.

# Pull Request Checklist


craigzour (approved):
Nice work! Good riddance to some of the legacy code :)

cypress/support/commands.js
```diff
@@ -23,3 +23,46 @@
...
+      delete req.headers["if-none-match"];
+      return req.continue((res) => {
+        // let's use the same test greeting
+        res.body.pageProps.formConfig = { ...mockedForm.form, formID: "test" };
+      });
+    });
+  });
+});
+
+Cypress.Commands.add("useFlag", (flagName, value) => {
```
falila, line 58: Very cleaver way to implement these depends via interceptors.

falila (approved):
Clean work 🌮

PR #691, opened 2022-03-18
feat: input character limit

jeberhardt:
# Summary | Résumé

1-3 sentence description of the changed you're proposing.
Please include the GitHub issue number or relevant acceptance criteria.

To specify that this PR closes a particular issue you can use the following format
to reference and issue `closes #[issue number]`. You can use any number of keywords to link
an issue to a PR. Please see the [following section](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
of GitHub's documentation.

links in this section

# Pull Request Checklist

### DIFF
--- a/components/forms/TextInput/TextInput.tsx
+++ b/components/forms/TextInput/TextInput.tsx
@@ -2,11 +2,13 @@ import React, { useEffect, useState } from "react";
 import classnames from "classnames";
 import { useField } from "formik";
 import ErrorMessage from "../ErrorMessage/ErrorMessage";
+import { CharacterCountMessages } from "@lib/types";
 
 interface RequiredTextInputProps {
   id: string;
   name: string;
   type: "text" | "email" | "name" | "number" | "password" | "search" | "tel" | "url";
+  characterCountMessages: CharacterCountMessages;
 }
 
 interface CustomTextInputProps {
@@ -21,8 +23,17 @@ export type OptionalTextInputProps = CustomTextInputProps & JSX.IntrinsicElement
 export type TextInputProps = RequiredTextInputProps & OptionalTextInputProps;
 
 export const TextInput = (props: TextInputProps): React.ReactElement => {
-  const { id, type, className, required, ariaDescribedBy, placeholder, autoComplete, maxLength } =
-    props;
+  const {
+    id,
+    type,
+    className,
+    required,
+    ariaDescribedBy,
+    placeholder,
+    autoComplete,
+    maxLength,
+    characterCountMessages,
+  } = props;
   const [field, meta, helpers] = useField(props);
   const classes = classnames("gc-input-text", className);
 
@@ -57,11 +68,14 @@ export const TextInput = (props: TextInputProps): React.ReactElement => {
         onChange={handleTextInputChange}
       />
       {maxLength && remainingCharacters < maxLength * 0.25 && remainingCharacters >= 0 && (
-        <div id="character-count-message">You have {remainingCharacters} characters left.</div>
+        <div id={"character-count-message-" + id}>
+          {characterCountMessages.part1} {remainingCharacters} {characterCountMessages.part2}
+        </div>
       )}
       {maxLength && remainingCharacters < 0 && (
-        <div id="character-count-message" className="gc-error-message">
-          You have {remainingCharacters * -1} characters too many.
+        <div id={"character-count-message-" + id} className="gc-error-message">
+          {characterCountMessages.part1Error} {remainingCharacters * -1}{" "}
+          {characterCountMessages.part2Error}
         </div>
       )}
     </>
--- a/components/forms/TextInput/TextInput.tsx
+++ b/components/forms/TextInput/TextInput.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import classnames from "classnames";
 import { useField } from "formik";
 import ErrorMessage from "../ErrorMessage/ErrorMessage";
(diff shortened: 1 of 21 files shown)


components/containers/Admin/Organizations/CreateOrganization.tsx
```diff
@@ -43,11 +43,33 @@ export const CreateOrganization = (): React.ReactElement => {
           <Label htmlFor="orgNameEn" required={true}>
             Organization Name (English)
           </Label>
-          <TextInput id="orgNameEn" name="orgNameEn" type="text" required={true}></TextInput>
+          <TextInput
+            id="orgNameEn"
```
falila, line 47: IMO: It may make more sens to stick to one style  for consistency accros the code base. 

`id="orgNameEn"'     name="orgNameEn`"  this is more consistent with javascript code.

Further in the code you will encounter a different syntax like so.
These below look more like html4/5-like attributes or in html.

```
id="input-type-text"
name="input-type-text"
```

falila (changes requested):
Minor changes that aren't urgent. 
Great work and well tested.

craigzour:
@jeberhardt Two things I have noticed so far:
- I know we had a discussion about how to pass the translation function to the components but now I am wondering why we did it this way. Could you remind me why you did not just import `import { useTranslation } from "next-i18next";` in the components like `FileInput.tsx` for instance?
- I see that you did not check the accessibility checkmark in the PR checklist, are you still working it? I tested the feature and voice over does not catch the dynamic error message. Since I have recently worked on something similar I would suggest a something like:
```
<input
        data-testid="textInput"
        className={classes}
        id={id}
        type={type}
        required={required}
        autoComplete={autoComplete ? autoComplete : "off"}
        aria-describedby={`${"character-count-message-" + id} ${ariaDescribedBy}`}
        placeholder={placeholder}
        {...field}
        onChange={handleTextInputChange}
      />
```
Pretty sure it would read the error when it comes up. However I am not sure it would be read every time there is an update so it could be something to investigate if we want to make this feature accessible.

components/containers/Admin/Organizations/CreateOrganization.tsx
jeberhardt: Fair enough. This isn't something that I introduced, but PRs will often reveal details that may have slipped though before.

components/forms/TextInput/TextInput.tsx
```diff
@@ -62,7 +62,7 @@ export const TextInput = (props: TextInputProps): React.ReactElement => {
         type={type}
         required={required}
         autoComplete={autoComplete ? autoComplete : "off"}
-        aria-describedby={ariaDescribedBy}
+        aria-describedby={`${meta.error} ${"character-count-message-" + id} ${ariaDescribedBy}`}
```
craigzour, line 65: Nice! Have you been able to test it? I just did and it reads it when it comes up but unfortunately does not react to dynamic updates of the error message. Not sure if there is a solution for now but will be looking into it ;)

craigzour: @jeberhardt Found this solution:
```
return (
    <>
      {meta.error ? <ErrorMessage>{meta.error}</ErrorMessage> : null}
      <input
        data-testid="textInput"
        className={classes}
        id={id}
        type={type}
        required={required}
        autoComplete={autoComplete ? autoComplete : "off"}
        aria-describedby={ariaDescribedBy}
        placeholder={placeholder}
        {...field}
        onChange={handleTextInputChange}
      />
      {maxLength && remainingCharacters < maxLength * 0.25 && remainingCharacters >= 0 && (
        <div id={"character-count-message-" + id} role="alert">
          {characterCountMessages.part1} {remainingCharacters} {characterCountMessages.part2}
        </div>
      )}
      {maxLength && remainingCharacters < 0 && (
        <div id={"character-count-message-" + id} className="gc-error-message" role="alert">
          {characterCountMessages.part1Error} {remainingCharacters * -1}{" "}
          {characterCountMessages.part2Error}
        </div>
      )}
    </>
  );
```

It uses `role="alert"` which acts as an ARIA live region and will read dynamic content when updated. The only remaining issue which I am not sure we can do something about is that depending on what you enter in the textfield, VoiceOver will mix the word you just entered with the alert message ("X characters left"). I will let you see by yourself and tell me if you also have the same behavior.

jeberhardt: I've made an update to it. It will read the updated `describeby` value every time the field gains focus.

jeberhardt: I had inadvertently used the `meta-error` value, when really I needed to use an ID. That may have been causing it to not read out any values.

jeberhardt: The spec that we are working from (https://design-system.service.gov.uk/components/character-count/) does not use `role="alert"`.

craigzour: @jeberhardt Just tested your latest changes but it does not work the same way as the UK version. UK version is close to what I would expect even if it has a small flaw (which is probably not something that can be solved). I guess you could compare the two versions and see what we are missing on our side 😕 

Something I noticed with our version is that VoiceOver will only read the updated number of characters (just the number) instead of the the full sentence. Here is a solution I found:
````
<div id={"characterCountMessage" + id} aria-atomic="true">
````

jeberhardt: Ah, I see now that the UK code uses ` aria-live="polite"` on the character count message. I'll try that, and maybe `aria-atomic` isn't required.

craigzour (approved):

falila (approved):
🍕

Issue #586, opened 2022-01-17
Only delete form responses from Vault after user confirmation. (API and UI)


bryan-robitaille:
**User story**: Persona based
As a form administrator
I need a way to easily retry a download of a form response
So that if there was a problem with the initial download I don't need to open a support issue.

**Story context**
Currently once a download attempt has been made on a form response from the vault the form response is marked for deletion and can no longer be returned as a result from the retrieval API.  If there is a network error the user is required to open a support ticket for form response retrieval from the archive.  This support process also relies heavily upon the user knowing the form response identifier.
In order to alleviate this need for external support and reduce the risk of data loss the retrieval API should require a successful acknowledgement of the form response ID's before returning a new batch of form responses for download.

`GET /api/retrieval` => return form responses
`DELETE /api/retrieval/[formResponseID]` => Mark each response as ready for archive/deletion

**Acceptance criteria**
- [ ] Update design doc 003 on new approach 
- [x] GET method will consistently return the same form responses and does not mark as retrieved
- [x] DELETE takes list of submissionIDs to mark as retrieved
- [x] DELETE responds with a 200 on success, 404 if ID is not found or invalid, 500 on server error.
- [ ] Update required ATO documentation to include archive and delete timelines.

bryan-robitaille:
I was just using the API path that was currently implemented but definitely open to discussion to using:

PR #672, opened 2022-03-04
feat: add new API patch to plug GC Notify callback feature

craigzour:
# Summary | Résumé

closes #242 
linked to https://github.com/cds-snc/forms-terraform/pull/186
based on https://docs.google.com/document/d/1XXtqmSeQXQRdklpz4FmpUdagKgr6new67633OsyHmzg (which will be updated shortly to reflect the final solution that was implemented)

- added new API path for GC Notify to call us back when a notification has failed to be delivered

Note:
- This PR adds a new environment variable `GC_NOTIFY_CALLBACK_BEARER_TOKEN`
- This PR replaces `LOCAL_S3_ENDPOINT` and `LOCAL_DYNAMODB_ENDPOINT` by `LOCAL_AWS_ENDPOINT` in the .env file
- It requires GC Notify callback to be enabled and properly set up using the portal. It has been done for staging using a specific token that has been saved in LastPass (Application Variables - Staging)

Warning:
- It looks like only one callback path can be set up in GC Notify so I am not sure how we can handle both Staging and Production since the URL will be different

# Test instructions | Instructions pour tester la modification

Here is the request that can be used to manually trigger the API

```
curl -X POST http://gc-forms/api/notify-callback
   -H "Content-Type: application/json"
   -H "Authorization: Bearer {token}"
   -d '{"id": "fcc82d18-622b-4cea-8843-ac7e28695696", "reference": "1c19ebc1-908c-4997-9d88-31498cd5cd70", "to": "clement.janin@cds-snc.ca", "status": "delivered", "provider_response": "nil", "created_at": "2017-05-14T12:15:30.000000Z", "completed_at": "2017-05-14T12:15:30.000000Z", "sent_at": "2017-05-14T12:15:30.000000Z", "notification_type": "email"}'
```

# Pull Request Checklist

Please complete the following items in the checklist before you request a review:

- [x] Have you completely tested the functionality of change introduced in this PR? Is the PR solving the problem it's meant to solve within the scope of the related issue?
- [x] The PR does not introduce any new issues such as failed tests, console warnings or new bugs.
- [x] If this PR adds a package have you ensured its licensed correctly and does not add additional security issues?
- [x] Is the code clean, readable and maintainable? Is it easy to understand and comprehend.
- [x] Does your code have adequate comprehensible comments? Do new functions have [docstrings](https://tsdoc.org/)?
- [x] Have you modified the change log and updated any relevant documentation?
[...]

PR #694, opened 2022-03-22
Retrieval API DELETE request

Moro-Code:
# Summary | Résumé

Linked to issue #586 

This PR should not be merged unless a prerequisite infrastructure PR in the forms-terraform repo is merged. ~~Once that PR is open it will be linked here.~~
https://github.com/cds-snc/forms-terraform/pull/191

This PR introduces several changes some major and some minor 

## Major Changes 

1. Moved the retrieval API to be under the path `id/[form]/retrieval`. This is to make the API experience more consistent by having the form ID passed in via the url parameter as opposed to a separate query argument 
2. Added DynamoDocumentDBClient. This allows us to work with DynamoDB with native JS types. This means the response from the API will look like this:
```JSON
{
   "response":[
      {
         "key":"value"
      }
   ]
}
```
as opposed to this
```JSON
{
   "response":[
      {
         "key":{
            "S":"value"
         }
      }
   ]
}
```
3. Removed maxRecord query argument. This presented little value to the user. The retrieval API will now always return a maximum of up to 10 values at a time.
4. Implementation of the DELETE method to take a string array of SubmissionIDs and mark the relevant items as retrieved and return the same list as a response if its successful 

## Minor Changes

1. Setting options parameter for the `jsonValidator` middleware to be an optional argument. This was to accommodate its usage for the DELETE method of the retrieval API and to allow further flexibility 
2.  Modified jsonValidator tests to account for new middleware format 
3. Changing the key for error code responses from `response` to `error` for consistency 
4. Changing some emitted logMessage.warn messages for greater clarity 

1. Set up localstack according to instructions in the forms-terraform repo 
2. Create a local form that is set to save to the vault and create a couple of responses. Invoke the reliability queue for each SubmissionID to put in the vault table ( note: There is a change that is needed in the reliability queue code to set the value of Retrieved to 0 when the record is initially pushed to the vault table. Without this you will not see any responses when doing a GET request See PR https://github.com/cds-snc/forms-terraform/pull/191 )
3. Send a DELETE request to the retrieval API with the body being the array submissionIDs in the vault to delete.
4. The response should be the same submissionIDs array that you sent and a GET request against the retrieval API should yield no responses 

Since this PR is so big and I have other priorities there were a couple of things I could not resolve immediately 

1. No distinct testing file for the `validTemporaryToken.ts` file in the middleware. Since this is net new code this should have had dedicated tests. I will open a ticket for this and link here. 
5. Testing for the retrieval API does not test emitted logMessages. This is actually core business logic as we are using these emitted messages in cloudwatch for traceability purposes. New tests written for the DELETE method test these outputs however for the GET methods I have more pressing things to do. I will open a separate PR for this 

# Pull Request Checklist


pages/api/id/[form]/retrieval.ts
```diff
@@ -0,0 +1,207 @@
...
+  DynamoDBDocumentClient,
+  QueryCommand,
+  UpdateCommand,
+  QueryCommandOutput,
+} from "@aws-sdk/lib-dynamodb";
+import { MiddlewareProps } from "@lib/types";
+
+interface ResponseListResult {
+  formResponses: QueryCommandOutput;
+  documentClient: DynamoDBDocumentClient;
```
Comment (craigzour) [inline pages/api/id/[form]/retrieval.ts:16]: Looks like this is not needed (anymore?)

pages/api/id/[form]/retrieval.ts
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * Helper function to instantiate DynamoDB and Document client.
+ * https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/dynamodb-example-document-client.html
+ */
+function connectToDynamo(): DynamoDBDocumentClient {
+  //Create dynamodb client
+  const db = new DynamoDBClient({
+    region: process.env.AWS_REGION ?? "ca-central-1",
+    endpoint: process.env.LOCAL_DYNAMODB_ENDPOINT,
```
Comment (craigzour) [inline pages/api/id/[form]/retrieval.ts:48]: Great addition! Wondering if we should not be just adding one env variable named `LOCAL_AWS_ENDPOINT` since it is the same address we use for two different services already and https://github.com/cds-snc/platform-forms-client/pull/672 will benefit from it as well. Right now in my implementation I decided to use the existing `LOCAL_S3_ENDPOINT` (with a comment) even though it is for the SNS client.

pages/api/id/[form]/retrieval.ts
```diff
@@ -0,0 +1,207 @@
...
+async function getFormResponses(
+  req: NextApiRequest,
+  res: NextApiResponse,
+  formID: string,
+  email?: string,
+  temporaryToken?: string
+): Promise<void> {
+  const result = await getLatestResponseList(res, formID);
+  // function will return false if there is an error. The response will have
+  // already been set so we do not need to continue with the function execution
```
Comment (craigzour) [inline pages/api/id/[form]/retrieval.ts:119]: Not sure it is a good idea since we could basically get rid of this comment just by moving the response logic from the helper function to here.

pages/api/id/[form]/retrieval.ts
```diff
@@ -0,0 +1,207 @@
...
+            (submissionID) => !submissionIDlist.includes(submissionID)
+          )}]`
+        : `[${requestBody.toString()}]`
+    );
+    logMessage.error(error as Error);
+    return res.status(500).json({ error: "Error on Server Side" });
+  }
+}
+
+// Only a GET request is allowed
```
Comment (craigzour) [inline pages/api/id/[form]/retrieval.ts:199]: Not up to date but thinking we could get rid of this since the middleware function is pretty transparent on what is allowed ;)

craigzour (changes requested):
Great work! Still got a few questions and suggestions for you :)

pages/api/id/[form]/retrieval.ts
Comment (Moro-Code) [reply pages/api/id/[form]/retrieval.ts:48]: Yes definitely could probably be a single environment variable. I would have to modify code I'm not currently working on for this PR so I would prefer a separate one to do this

pages/api/id/[form]/retrieval.ts
Comment (Moro-Code) [reply pages/api/id/[form]/retrieval.ts:119]: Think you're right here. The helper function was a remnant of my previous implementation where it was reused. I must have forgotten to move back the code

pages/api/id/[form]/retrieval.ts
Comment (Moro-Code) [reply pages/api/id/[form]/retrieval.ts:199]: Yup I agree thanks for the catch

Moro-Code:
> Great work! Still got a few questions and suggestions for you :)

Thanks for the great comments ! Will implement changes and suggesting we address in a separate PR for some

pages/api/id/[form]/retrieval.ts
Comment (craigzour) [reply pages/api/id/[form]/retrieval.ts:38]: I can definitely tackle this as part of my https://github.com/cds-snc/platform-forms-client/pull/672 once this is merged.

craigzour (approved):
LGTM!

PR #706, opened 2022-03-28
fix: remove redundant `required` copy.

jeberhardt:
# Summary | Résumé

There is no issue for this PR.

The removed code is redundant for non-group form elements. If a element is already marked as `required`, then the screen reader will read out: "Full name required required", since it is reading out the visually-hidden copy of `required`, as well as the `required` attribute for the form elements (ie: `input`, `textarea`, etc..).

- load a form with required elements
- tabbing through the form elements should only read out `required` only once per element, regardless if it is a group element or not

# Pull Request Checklist

### DIFF
--- a/components/forms/Label/Label.tsx
+++ b/components/forms/Label/Label.tsx
@@ -37,7 +37,6 @@ export const Label = (props: LabelProps): React.ReactElement => {
           ({t("required")})
         </span>
       )}
-      {required && <i className="visually-hidden">{t("required-field")}</i>}
       {hint && <span className="gc-hint">{hint}</span>}
     </>
   );
--- a/components/forms/Label/Label.tsx
+++ b/components/forms/Label/Label.tsx
@@ -37,6 +37,7 @@ export const Label = (props: LabelProps): React.ReactElement => {
           ({t("required")})
         </span>
       )}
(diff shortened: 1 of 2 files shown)


Moro-Code (approved):

components/forms/Label/Label.tsx
```diff
@@ -37,7 +37,7 @@ export const Label = (props: LabelProps): React.ReactElement => {
           ({t("required")})
         </span>
       )}
-      {required && <i className="visually-hidden">{t("required-field")}</i>}
+      {group && required && <i className="visually-hidden">{t("required-field")}</i>}
```
Moro-Code, line 40: What is this prop used for exactly ? If it's not used anywhere else should we remove it from the destructuring assignment from line 18.

Moro-Code (commented):
Had a question on this `group` prop

components/forms/Label/Label.tsx
jeberhardt: `group` is used to determine if the current element is a "single" element, like "input", or if it represents multiple elements, like checkboxes. It is already used on **Line 45**.

Moro-Code (approved):

PR #718, opened 2022-04-11
Test storybook actions

falila:
# Summary | Résumé
Testing story-book-actions


craigzour:
@falila The action should only be triggered when there is a push on the `main` branch.

PR #720, opened 2022-04-12
Enable GC notify's callback  support per environment by adding a new service

### LINKED ISSUE #696: Create a second GC Notify service account (one for Staging and one for Production)
**User story**

As a developer
I need to be able to set up a GC Notify service for one specific environment (production or staging)
So that I can leverage the entirety of GC Notify features

**Context**

When developing https://github.com/cds-snc/platform-forms-client/pull/672 I found out that GC Notify callback feature can not handle multiple environments as it can only take one URL on which it will send POST requests. In order to have both our Staging and Production environments working similarly we need to solve this limitation.

**Acceptance criteria**

- [x] Create new GC Notify service account and define which is for Staging and Production
- [x] Move resources from one account to the other depending on the targeted environment
- [x] Clean resources that have been moved to the second account
- [x] Set up GC Notify callback integration on both account

falila:
closed #696 
Depends on [PR](https://github.com/cds-snc/forms-terraform/pull/196)


.env.example
```diff
@@ -16,3 +16,14 @@ GC_NOTIFY_CALLBACK_BEARER_TOKEN=
 
 # Used when running against LocalStack
 LOCAL_AWS_ENDPOINT=http://localhost:4566
+
+# Local dev 
+# Temporary token templateID
+TEMPORAY_TOKEN_TEMPLATE_ID=7199f6c9-41de-4941-85be-6fa6919b05d9
+
+# A link to GC Nofify service like staging or production
+GC_NOTIFY_URL=https://staging.notification.cdssandbox.xyz
```
Moro-Code, line 25: We're not using notify staging for this. We just want a separate service in the production instance of Notify

falila: After  a chat with @Moro-Code  it seems that Notify staging is not guaranteed to up all time and free of bugs. 
For those reasons, i will be creating a production instance of Notify for staging.

.env.example
```diff
@@ -16,3 +16,14 @@ GC_NOTIFY_CALLBACK_BEARER_TOKEN=
 
 # Used when running against LocalStack
 LOCAL_AWS_ENDPOINT=http://localhost:4566
+
+# Local dev 
+# Temporary token templateID
+TEMPORAY_TOKEN_TEMPLATE_ID=b6885d06-d10a-422a-973f-05e274d9aa86
+
+# A link to GC Nofify service like staging or production
+GC_NOTIFY_URL=https://notification.canada.ca
```
Moro-Code, line 25: This is the incorrect url. It should be https://api.notification.canada.ca . We should probably go back to hardcoding this

pages/api/submit.ts
```diff
@@ -97,9 +97,9 @@ const callLambda = async (formID: string, fields: Responses, language: string) =
 };
 
 const previewNotify = async (form: PublicFormSchemaProperties, fields: Responses) => {
-  const templateID = "92096ac6-1cc5-40ae-9052-fffdb8439a90";
+  const templateID = process.env.TEMPLATE_ID;
   const notify = new NotifyClient(
-    "https://api.notification.canada.ca",
+    process.env.GC_NOTIFY_URL,
```
Moro-Code, line 102: This is not present in the ecs form_viewer task JSON which means it wont be present in staging and production. See comment on your PR in forms-terraform. I think going back to a hardcoded value is preferred since this is static and unchanging.

Moro-Code (changes requested):

pages/api/token/temporary.ts
```diff
@@ -95,14 +95,15 @@ const updateTemporaryToken = async (
 };
 
 const sendTemporaryTokenByEmail = async (email: string, temporaryToken: string) => {
+  const sendTempTokenTemplateID = process.env.TEMPORAY_TOKEN_TEMPLATE_ID;
```
craigzour, line 98: `TEMPORARY_TOKEN_TEMPLATE_ID` instead of `TEMPORAY_TOKEN_TEMPLATE_ID`

craigzour (changes requested):

craigzour (approved):

.env.example
```diff
@@ -17,8 +17,15 @@ GC_NOTIFY_CALLBACK_BEARER_TOKEN=
 # Used when running against LocalStack
 LOCAL_AWS_ENDPOINT=http://localhost:4566
 
+# Local dev 
+# Temporary token templateID
+TEMPORAY_TOKEN_TEMPLATE_ID=b6885d06-d10a-422a-973f-05e274d9aa86
+
```
bryan-robitaille, line 23: Are these the actual Notify values?
Although not exactly secret values they should probably not be listed in our `.env.example` file.  Always use fake strings when providing examples.

bryan-robitaille: I'm also aware these were hardcoded before but if we're going with a dynamic value it might be best to not use the real one in case it causes confusion.

pages/api/token/temporary.ts
```diff
@@ -95,14 +95,15 @@ const updateTemporaryToken = async (
 };
 
 const sendTemporaryTokenByEmail = async (email: string, temporaryToken: string) => {
+  const sendTempTokenTemplateID = process.env.TEMPORARY_TOKEN_TEMPLATE_ID;
   const notifyClient = new NotifyClient(
     "https://api.notification.canada.ca",
-    process.env.NOTIFY_API_KEY
+    process.env.NOTIFY_API_KEY ?? "thisIsATestKey"
```
bryan-robitaille, line 101: I'm curious as to why a default value was added for the 'NOTIFY_API_KEY'.  Would we not want the call to error out if the env var does not exist in the system?  With the default value it might throw a 403 error (forbidden) instead of a conneciton error.  The differece between the two errors could be important when debugging errors in Staging or Production.

bryan-robitaille (changes requested):
A few quick questions but will be quick to approve once responded to.

pages/api/token/temporary.ts
falila: You're absolutely correct, the default value here won't make it obvious from  the standpoint of an error handling.

Moro-Code (approved):

PR #725, opened 2022-04-19
NextJS v12 upgrade + others

### LINKED ISSUE #518: Migrate from Next JS 11 to 12
With the release of Next JS 12 we should migrate the application to use it. Next JS 12 comes with some breaking changes so we want to ensure we do not break any functionality or accessibility. These are

- After making webpack 5 the default in Next.js 11, we've now officially removed webpack 4. We've worked closely with the community to ensure a smooth transition to webpack 5.
- target in next.config.js is no longer needed.
- next/image now uses a span as the wrapping element instead of a div.
- The minimum Node.js version has been bumped from 12.0.0 to 12.22.0, which is the first version of Node.js with native ES modules support.

Acceptance Criteria:

- [ ] Ensure all functionality works as expected and there are no regressions introduced 
- [ ] Ensure there are no a11y regressions introduced by the upgrade

bryan-robitaille:
# Summary | Résumé

This PR includes the following major package upgrades and their new implementation for production dependencies:
- Nextjs 11 => 12
- Tailwind 2 => 3
- IOredis 4 => 5
- Next-18next 8 => 11
- Pino 6 => 7
- React 17 => 18
- react-app-polyfill 2 => 3
- react-dom 17 => 18

The following packages were removed:
- raw-loader: No longer required as webpack 5 can import assets directly.
- react-spinners: No longer needed and Tailwind spinner using 'animate' has been implemented
- ioredis-mock: No longer needed, mocking call in Jest and Cypress and during NestJS build

The following packages were added:
 - cssnano: To minify CSS in postprocessing
 
Refactored the following:
- Google Fonts are pulled in at build time instead of async on page render
- Removed jest-chain as usage want not significant and the package is not longer compatible with Jest > 26.
- Implemented best practice in using useTranslation like a hook and not prop drilling down through components.
- Set polyfills package to use browserlist instead of a separate options list.
- Form Timer functionality into it's own custom react hook.  Separating logic from component.  Refactored associated Jest tests for functional scope vs end-to-end.
- Remove cypress `wait` and replace with time manipulation utilities `clock` and `tick`

Closed #518 

- `yarn install` to refresh package
- `yarn dev` to test current app functionality 
- `yarn build && yarn start` to build production version of app

- NextJS 12 now has a new middleware option however the current middleware functions have not been ported to this new pattern.
- Babel packages have been left in even though they are no longer used by Next.  Concerned removal may impact storybook compilation.
- Should storybook share the same package.json as the NextJS app?

# Pull Request Checklist


CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed maxRecord query argument. This presented little value to the user. The retrieval API will now always return a maximum of up to 10 values at a time. [#694](https://github.com/cds-snc/platform-forms-client/pull/694)
 - Design of DynamicRow component buttons [#528](https://github.com/cds-snc/platform-forms-client/issues/528)
 - Logging strategy. Now sending info, warn and error log types to AWS [#699](https://github.com/cds-snc/platform-forms-client/issues/699)
+- Upgraded NextJS and other associated GCForms dependencies to next major version. [#725](https://github.com/cds-snc/platform-forms-client/pull/725)
```
craigzour, line 36: This should be part of a new `unreleased` section :) I did add it in my [PR](https://github.com/cds-snc/platform-forms-client/pull/727/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR8) in case it gets merged before yours

craigzour:
Did review it but will not approve it alone as I do not feel comfortable doing so.

lib/flags.ts
```diff
@@ -12,10 +11,13 @@ const loadMockedValues = async () => {
     const { default: initialSettings }: { default: Record<string, boolean> } = await import(
       "../flag_initialization/default_flag_settings.json"
     );
-    for (const [key, value] of Object.entries(initialSettings)) {
-      logMessage.debug(`Creating flag: ${key} with value ${value} (Mocked Redis)`);
+    for (const key in initialSettings) {
+      logMessage.info(
```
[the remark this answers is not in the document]
bryan-robitaille: Good catch.  Reverting back to info.  Must have missed it during the rebase.

.github/workflows/eslint.yml
```diff
@@ -13,10 +13,10 @@ jobs:
     if: github.event_name == 'push'
     steps:
       - uses: actions/checkout@v1
-      - name: Node.JS 12
+      - name: Node.JS 16
```
Moro-Code, line 16: If we're updating to Node 16 for eslint then we should also do the same for our image running on ECS

lib/hooks/useFormTimer.tsx
```diff
@@ -0,0 +1,91 @@
...
+  disableTimer: () => void;
+};
+export const useFormTimer = (): [formTimerStateType, formTimerDispatchType] => {
+  enum actions {
+    START = "START",
+    CHECK = "CHECK",
+    DISABLE = "DISABLE",
+  }
+
+  function formTimerReducer(
```
Moro-Code, line 27: Very smart of using a reducer here. I think this is a great implementation

components/forms/TextArea/TextArea.test.js
```diff
@@ -92,36 +91,35 @@ describe("Verfify character count restrictions", () => {
 
   it("does not display any message when not enough characters have been typed in", () => {
     const textInput = screen.getByRole("textbox");
-    userEvent.type(textInput, "This is 21 characters");
+    fireEvent.change(textInput, { target: { value: "This is 21 characters" } });
```
Moro-Code, line 94: Why this switch from userEvent to fireEvent? 

According to the documentation 

>Most projects have a few use cases for fireEvent, but the majority of the time you should probably use [@testing-library/user-event](https://testing-library.com/docs/user-event/intro).

components/forms/TextInput/TextInput.test.js
```diff
@@ -87,27 +88,27 @@ describe("Verfify character count restrictions", () => {
 
   it("does not display any message when not enough characters have been typed in", () => {
     const textInput = screen.getByRole("textbox");
-    userEvent.type(textInput, "This is 21 characters");
+    fireEvent.change(textInput, { target: { value: "This is 21 characters" } });
```
Moro-Code, line 91: See comment on switching from userEvent to fireEvent

components/forms/TextPage/TextPage.tsx
```diff
@@ -15,12 +14,19 @@ interface TextPageProps {
...
+  urlQuery: string | null;
+}
+
+const PageContent = ({ pageText, urlQuery }: PageContextProps) => {
   // Check if there's a custom text for the end page specified in the form's JSON config
   if (pageText && pageText !== undefined) {
     return <RichText className="confirmation">{pageText}</RichText>;
   }
 
+  const { t } = useTranslation("confirmation");
```
Moro-Code, line 28: If the components are on the same file. Does it not make sense to just pass the t function from the parent TextPage to the child PageContent. Seems to be a duplication otherwise

components/globals/Loader.tsx
```diff
@@ -1,28 +1,18 @@
...
-  loading = true,
-  message,
-}: CustomLoaderProps): React.ReactElement => {
+export const Loader = ({ message }: LoaderProps): React.ReactElement => {
   return (
     <div role="status" className="text-center">
       <p className="pb-8">{message}</p>
-      <Loader loading={loading} color="#00703C" size="25px" css={override} />
+      <div className="flex items-center justify-center">
+        <div className="w-40 h-40 border-t-4 border-b-4 border-green-default rounded-full animate-spin"></div>
```
Moro-Code, line 12: Nice use of tailwind instead of a package

pages/_app.tsx
```diff
@@ -1,24 +1,22 @@
...
 import Base from "../components/globals/Base";
 import "../styles/app.scss";
-import { useHMR } from "../lib/hooks/useHMR";
 import i18nextConfig from "../next-i18next.config";
 
 const SafeHydrate = ({ children }: { children: React.ReactNode }) => {
   return (
-    <div suppressHydrationWarning={Boolean(process.env.CYPRESS)}>
-      {typeof window === "undefined" && process.env.CYPRESS ? null : children}
+    <div suppressHydrationWarning={Boolean(process.env.ISOLATED_INSTANCE)}>
```
Moro-Code, line 13: Not sure what this is doing here. I know it existed before and just changing the variable name but would be helpful to include a comment

pages/admin/users.tsx
```diff
@@ -47,18 +47,20 @@ const AdminEnable = ({
...
 };
 
 interface UserRowProps {
   user: AuthenticatedUser;
-  t: TFunction;
 }
 
-const UserRow = ({ user, t }: UserRowProps) => {
+const UserRow = ({ user }: UserRowProps) => {
+  const { t } = useTranslation();
```
Moro-Code, line 63: Is this really more efficient then passing it in via the parent ?

Moro-Code (commented):
This is a heroic effort. Well done @bryan-robitaille . I'm have a couple of comments that would be good to have a discussion on. Pretty minor things overall. Otherwise I think this is good to go

lib/integration/redisConnector.ts
```diff
@@ -0,0 +1,22 @@
...
+  if (!redisConnection) {
+    const redis = await createRedisInstance();
+    redisConnection = redis;
+    return redisConnection;
+  } else {
+    return redisConnection;
+  }
+};
+
+if (!process.env.ISOLATED_INSTANCE) {
```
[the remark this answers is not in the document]
bryan-robitaille: Added description to `.env.example`.
The var is used when no Redis instance is available and to identify testing environments.  When set the application returns all default flags values found in the flag initialization settings.
This variable is automatically set when running tests (Jest + Cypress) and during the NextJS build process when pre-rendering pages

lib/integration/redisConnector.ts
```diff
@@ -0,0 +1,22 @@
+import Redis from "ioredis";
```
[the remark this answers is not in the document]
bryan-robitaille: A new file yes, but existing logic.  A test or two can be written to test it's extremely simply logic.

components/forms/TextPage/TextPage.tsx
bryan-robitaille: My thought process was the separation of scope and self containment of components.  Efficiency doesn't really come into play in the sense that each useTranslation hook is coupled with Suspense.  The hook simply accesses the instance that already exists on the Page context.  By including it on each functional component that requires translation it simplifies the props and TypeScript typing needed for each component declaration.
Just my 2 cents.

pages/_app.tsx
bryan-robitaille: I'll add a comment in the code.
Essentially this block disables SSR when in testing mode.  This is because in Cypress we're manipulating and mocking the API response calls and the SSR pages were not matching the React rendered pages generating a warning in the browser console.

components/forms/TextInput/TextInput.test.js
bryan-robitaille: See earlier response

components/forms/TextPage/TextPage.tsx
Moro-Code: Think it's a fairly minor detail then. If the useTranslation hook doesn't add any time to the rendering or causes a re-render it can go either way. But in general this is something to think about when deciding whether to pass down or use a hook. I find often times devs use one or the other without thinking about renders or performance ( not saying thats you :p  just mean in general )

Moro-Code (dismissed):
Looking good to go !

PR #727, opened 2022-04-20
Feature/system audit privileged functions

### LINKED ISSUE #700: system audits the execution of privileged functions

craigzour:
# Summary | Résumé

closes #700 

- Logging admin activity in our PostgreSQL database.

Screen Shot 2022-04-20 at 10 28 13 AM

- Run the branch locally and play with the admin panel in order to generate new logs in the database (table name = `admin_logs`)

- A new issue has been created to handle `INSERT` performance issue in the newly create PostgreSQL table: https://github.com/cds-snc/platform-forms-client/issues/740

# Pull Request Checklist

### DIFF
--- a/lib/adminLogs.ts
+++ b/lib/adminLogs.ts
@@ -1,0 +1,28 @@
+import dbConnector from "@lib/integration/dbConnector";
+import executeQuery from "@lib/integration/queryManager";
+import { logMessage } from "@lib/logger";
+import { AdminLogAction, AdminLogEvent } from "./types";
+
+/**
+ * Logs usage of privileged functions by admins
+ * @param adminSession
+ * @param action
+ * @param event
+ * @param description
+ */
+export const logAdminActivity = async (
+  userId: string,
+  action: AdminLogAction,
+  event: AdminLogEvent,
+  description: string
+): Promise<void> => {
+  try {
+    await executeQuery(
+      await dbConnector(),
+      "INSERT INTO admin_logs (user_id, action, event, description) VALUES ($1, $2, $3, $4)",
+      [userId, action, event, description]
+    );
+  } catch (error) {
+    logMessage.error(error as Error);
+  }
+};
--- a/pages/api/templates.ts
+++ b/pages/api/templates.ts
@@ -4,6 +4,8 @@ import { middleware, jsonValidator, cors, sessionExists } from "@lib/middleware"
 import templatesSchema from "@lib/middleware/schemas/templates.schema.json";
 import { NextApiRequest, NextApiResponse } from "next";
 import { isAdmin } from "@lib/auth";
+import { logAdminActivity } from "@lib/adminLogs";
+import { AdminLogAction, AdminLogEvent } from "@lib/types";
 
 const allowedMethods = ["GET", "POST", "PUT", "DELETE"];
 const authenticatedMethods = ["POST", "PUT", "DELETE"];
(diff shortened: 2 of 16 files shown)


lib/adminLogs.ts
```diff
@@ -0,0 +1,126 @@
...
+export const logFormAccessAddEmail = async (
+  userAdminId: string,
+  email: string,
+  formId: string
+): Promise<void> => {
+  return logAdminActivity(
+    userAdminId,
+    Action.Create,
+    "GrantInitialFormAccess",
+    `Email ${email} has been given access to form id:${formId}`
```
jeberhardt, line 97: Consistency in the format would be ideal, with colons and spacing.

`Email: ${email} has been given access to form id: ${formId}`

pages/api/templates.ts
```diff
@@ -13,6 +14,13 @@ const templates = async (req: NextApiRequest, res: NextApiResponse) => {
     const session = await isAdmin({ req });
     const response = await crudTemplates({ ...req.body, method: req.method, session });
     if (response) {
+      if (session && session.user.id && (req.method === "POST" || "PUT" || "DELETE")) {
```
jeberhardt, line 17: I haven't executed this line of code, but I think you're actually missing the comparison against `req.method`.

```(req.method === "POST" || req.method === "PUT" || req.method === "DELETE")```

File: lib/adminLogs.ts [hunk 3]
```diff
@@ -0,0 +1,126 @@
...
+      await dbConnector(),
+      "INSERT INTO admin_logs (user_id, action, event, description) VALUES ($1, $2, $3, $4)",
+      [userId, action.toString(), event, description]
+    );
+  } catch (error) {
+    logMessage.error(error as Error);
+  }
+};
+
+export const logAdminRoleUpdate = async (
```
Moro-Code, line 36: I'm not sure this is a good way to go in terms of calling logAdminActivity. All these functions are really doing is calling another function. This means that a new function is going to be added every time a new event is to be added. If there is no other logic it would be better to just create a string enum with the events and have logAdminActivity called directly from the main code. Further evidence of this is that these functions are `async` when all they are doing is returning the result of another function. These functions don't need to be async at all.

tests/api/flags.test.js
```diff
@@ -0,0 +1,59 @@
+import { createMocks } from "node-mocks-http";
```
Moro-Code, line 1: Thanks for increasing coverage on forms :)

tests/api/templates.test.js
```diff
@@ -105,4 +112,76 @@ describe("Test JSON validation scenarios", () => {
     expect(res.statusCode).toBe(400);
     expect(JSON.parse(res._getData()).error).toContain('instance requires property "form"');
   });
+
+  it("Should log admin activity if POST API call completed successfully", async () => {
```
Moro-Code, line 116: I think this test can be removed by simply adding `expect(logFormUpload).toHaveBeenCalled()` in the first test. Perhaps you can rename the first test to `Should successfully handle a POST request to create a template`

tests/api/templates.test.js
```diff
@@ -105,4 +112,76 @@ describe("Test JSON validation scenarios", () => {
...
+      },
+    });
+
+    await templates(req, res);
+
+    expect(res.statusCode).toBe(200);
+    expect(logFormUpload).toHaveBeenCalled();
+  });
+
+  it("Should log admin activity if PUT API call completed successfully", async () => {
```
Moro-Code, line 140: Minor nit but this test is not only testing logging admin activity. Perhaps it would be better suited to rename to `Should successfully handle PUT request`

tests/api/templates.test.js
```diff
@@ -105,4 +112,76 @@ describe("Test JSON validation scenarios", () => {
...
+      },
+    });
+
+    await templates(req, res);
+
+    expect(res.statusCode).toBe(200);
+    expect(logFormUpdate).toHaveBeenCalled();
+  });
+
+  it("Should log admin activity if DELETE API call completed successfully", async () => {
```
Moro-Code, line 164: See comment above

tests/api/users.test.js
```diff
@@ -230,5 +235,34 @@ describe("Users API endpoint", () => {
         expect.objectContaining({ error: "User not found" })
       );
     });
+
+    it("Should log admin activity if API call completed successfully", async () => {
```
Moro-Code, line 239: I suspect this test can be wrapped up by including `expect(logAdminRoleUpdate).toHaveBeenCalled();` in one of the previous tests

Moro-Code (commented):
This is really great work @craigzour well done. I especially liked that the route calling these logging functions will not fail if the record is not going to be inserted in the database. With that said I have two overall comments 

1. A lot of these functions in `adminLog.ts` are simply passthrough functions to set the event action. This gets messy real quick because that means a new function must be added in order to add a new event . This means you have to read all the other functions to make sure you're not duplicating. The code becomes obfuscated as well as you'd have to follow the function declaration to see what event you are logging. I would suggest you go the string enum route like you did for the Action and have the code directly call `logAdminActivity`. This insures you don't have wild strings as I think this was your intention to prevent but at the same time removes these unnecessary functions 

2. At some point this  table will be a problem. It will get very large and cause inserts to be extremely slow which could cause our admin to run like molasses or in fact cause the whole app to crash every so often under high loads since the database will be running VACUUM to free up memory and disk. We experienced this with Notify . I suggest we create a new issue for this and potentially looking at ways of mitigating this and the level of effort before merging this PR. Once we have identified that we can merge this PR and it can be handled as a separate task. Potential solutions include for example creating a cron lambda to archive records longer than three months and reindex the table to prune the tree to free up memory and make inserts faster.

tests/api/users.test.js
craigzour: There was none so I will rename to `Should successfully handle PUT request`

Moro-Code (approved):

PR #728, opened 2022-04-20
Add security attributes

### LINKED ISSUE #701: Data Classification Attributes
In support of GCForms's SRTM control AC-16.
>Information is represented internally within information systems using abstractions known as data structures. Internal data structures can represent different types of entities, both active and passive. Active entities, also known as subjects, are typically associated with individuals, devices, or processes acting on behalf of individuals. Passive entities, also known as objects, are typically associated with data structures such as records, buffers, tables, files, inter-process pipes, and communications ports. Security attributes, a form of metadata, are abstractions representing the basic properties or characteristics of active and passive entities with respect to safeguarding information. These attributes may be associated with active entities (i.e., subjects) that have the potential to send or receive information, to cause information to flow among objects, or to change the information system state. These attributes may also be associated with passive entities (i.e., objects) that contain or receive information. The association of security attributes to subjects and objects is referred to as binding and is typically inclusive of setting the attribute value and the attribute type. Security attributes when bound to data/information enable the enforcement of information security policies for access control and information flow control either through organizational processes or information system functions or mechanisms. The content or assigned values of security attributes can directly affect the ability of individuals to access organizational information.

Scope: 
Define and associate a list of Security Attributes to various Forms and associated form submissions.
For example a form response that is Protected B must be able to be identified vs a form response that is unclassified.

Acceptance Criteria:
- [x] Security Attributes are clearly defined and aligned with government standards
- [x] As an administrator I can verify the security attribute of a response without viewing the data.  For example the response had an error during processing and have been moved to the Dead Letter Queue.  An administrator needs to be aware of the level of content of the information contained in the erred response so specific protocols regarding the treatment of the contained information can be put into place.
- [x] The security attributes are set at the form level and are propagated to it's form submissions.

falila:
# Summary | Résumé
Set a security attribute  that will allow an administrator to take the appropriate actions when retrieving erred data.
closed #701 
[deps](https://github.com/cds-snc/forms-terraform/pull/200) 

Security Attribute Badge: 

Screen Shot 2022-05-03 at 3 23 07 PM


lib/middleware/schemas/templates.schema.json
```diff
@@ -7,6 +7,11 @@
   "default": {},
   "examples": [],
   "properties": {
+    "securityAttribute":{
```
jeberhardt, line 10: It would be great to add an enum here, listing the allowed values.

ie:
```"enum": [
            "textField",
            "textArea",
            "dropdown",
            "radio",
            "checkbox",
            "fileInput",
            "dynamicRow",
            "richText"
          ]
          ```

File: lib/tests/crud.test.js [hunk 2]
```diff
@@ -32,6 +32,7 @@ jest.mock("@aws-sdk/client-lambda", () => {
                             some: "form",
                           },
                           publishingStatus: true,
+                          securityAttribute: "Unclassified",
```
jeberhardt, line 35: Is there any reason this has an uppercase "U"? As more values are added, it might be easier if everything was lowercase.

tests/api/id/[form]/retrieval.test.js
```diff
@@ -508,4 +508,67 @@ describe("/api/retrieval", () => {
...
+        ) {
+          return {
+            rows: [{ column: 1 }],
+            rowCount: 1,
+          };
+        }
+      });
+      const dynamodbExpectedReponses = {
+        Items: [
+          { FormID: "01", SubmissionID: "51", SecurityAttribute: "Protectd B" },
```
Comment (jeberhardt) [inline tests/api/id/[form]/retrieval.test.js:549]: Type-o in `Proctectd`. Missing the last `e`. Should be `Protected`.

lib/middleware/schemas/templates.schema.json
falila: Absolutely ! it will improve the readability as well.

lib/integration/helpers.ts
```diff
@@ -162,6 +162,8 @@ function _buildFormDataObject(form: PublicFormSchemaProperties, values: Response
   }
 
   formData["formID"] = `${form.formID}`;
+  //add security attribtue
```
jeberhardt, line 165: Is this comment necessary?

lib/tests/crud.test.js
falila: [Security attribute list](https://github.com/cds-snc/platform-forms-client/issues/701#issuecomment-1098166716)

falila: [Security attribute list](https://github.com/cds-snc/platform-forms-client/issues/701#issuecomment-1098166716)

lib/integration/crud.ts
```diff
@@ -259,6 +260,7 @@ const _onlyIncludePublicProperties = async ({
       return {
         formID: template.formID,
         publishingStatus: template.formConfig.publishingStatus,
+        securityAttribute: template.formConfig?.securityAttribute ?? null,
```
jeberhardt, line 263: I think the null check (?) after `formConfig` is unnecessary.

lib/integration/crud.ts
```diff
@@ -232,6 +232,7 @@ async function _getFormByStatus(
             ...record.formConfig.form,
             publishingStatus: record.formConfig.publishingStatus,
             displayAlphaBanner: record.formConfig.displayAlphaBanner ?? true,
+            securityAttribute: record.formConfig?.securityAttribute ?? null,
```
jeberhardt, line 235: I think the null check (?) after formConfig is unnecessary.

documentation/Forms/FormViewer.stories.mdx
```diff
@@ -56,6 +57,9 @@ There are 2 main parts to the structure:
    `publishingStatus`: If set to false, the link to the form will not appear on the main page, rather under "sandbox"
 
    `displayAlphaBanner`: If set to false, the ALPHA tag will not appear in the header of the page. If not set if will be defaulted to true.
+  
+   `securityAttribute`: It shows an administrator what proceduce should be taken when accessing a form's responses, possible value Protected B, A, Unclassied
```
jeberhardt, line 61: Type-o: `proceduce`.

jeberhardt, line 61: As a piece of documentation, the full string of "Protected A" should be used here, not just "A".

tests/api/id/[form]/retrieval.test.js
Comment (jeberhardt) [inline tests/api/id/[form]/retrieval.test.js:549]: Are there FormIDs that actually have leading zeros?  ie: `FormID : "01"`.

Comment (falila) [reply tests/api/id/[form]/retrieval.test.js:549]: There aren't yet. In the future we may completely drop a numerical order so i don't think it makes any difference in this scenario.

Comment (falila) [reply tests/api/id/[form]/retrieval.test.js:549]: There aren't yet. In the future we may completely drop a numerical order so i don't think it makes any difference in this scenario.

lib/integration/crud.ts
```diff
@@ -259,6 +260,7 @@ const _onlyIncludePublicProperties = async ({
       return {
         formID: template.formID,
         publishingStatus: template.formConfig.publishingStatus,
+        securityAttribute: template.formConfig?.securityAttribute ?? "",
```
jeberhardt, line 263: The `?` here is not required, since `formConfig` will always be defined in this case.

lib/integration/helpers.ts
```diff
@@ -162,6 +162,7 @@ function _buildFormDataObject(form: PublicFormSchemaProperties, values: Response
   }
 
   formData["formID"] = `${form.formID}`;
+  formData.securityAttribute = form?.securityAttribute ?? "";
```
jeberhardt, line 165: The ? here is not required, since form will always be defined in this case.

lib/tests/helpers.test.js
```diff
@@ -170,6 +171,7 @@ describe("Submit and Template helpers", () => {
 
     expect(form.formID).toEqual("1");
     expect(form.publishingStatus).toBe(true);
+    expect(form?.securityAttribute).toEqual("Unclassified");
```
jeberhardt, line 174: `?` is not necessary here.

lib/integration/crud.ts
falila: same as [here](https://github.com/cds-snc/platform-forms-client/pull/728#discussion_r856516608)

falila: same as [here](https://github.com/cds-snc/platform-forms-client/pull/728#discussion_r856516608)

jeberhardt (dismissed):

craigzour:
@falila It is almost good to go but @bryan-robitaille added a comment to the ticket https://github.com/cds-snc/platform-forms-client/issues/701#issuecomment-1112551569

components/globals/Base.js
```diff
@@ -38,7 +39,15 @@ const Base = ({ children }) => {
             {isAdmin && <AdminNav user={children.props.user} />}
           </header>
         )}
-        <main id="content">{children}</main>
+
+        <main id="content">
+          {formConfig?.securityAttribute && (
+            <SecurityAttributeBadge
```
craigzour, line 45: I wonder if this is the right place to add the security attribute badge. Should not this be part of the Form component directly? The `Base` component is also used in non form related pages so it should probably stay clean and only includes basic things. What do you think?

craigzour:
@falila  Did you plan on adding integration tests to make sure the badge is displayed depending on form definitions?

components/globals/Base.js
falila: One reason i put it there was that i thought it make much more sens to have the badge above the form's title instead of . Do you see 
Screen Shot 2022-05-03 at 2 51 25 PM

craigzour: Gotcha! Then it can still be done through the `DynamicForm.tsx` (which is closer to the form component than `Base`) if I am correct ;)

falila: That's right 👍🏿  I've forgotten that a form is wrapped with DynamicFrom. Ty

components/globals/Base.js
```diff
@@ -39,15 +38,7 @@ const Base = ({ children }) => {
...
-
-        <main id="content">
-          {formConfig?.securityAttribute && (
-            <SecurityAttributeBadge
-              securityLevel={formConfig.securityAttribute}
-            ></SecurityAttributeBadge>
-          )}
-          {children}
-        </main>
+        <main id="content">{children}</main>
```
craigzour, line 41: Missing a `>`

tests/data/tsbContactTestFormProtectedB.json
```diff
@@ -0,0 +1,189 @@
+{
```
craigzour, line 1: Just a comment here to say that you did not need to create such a big form config for that specific test ;) It could have been a very minimal one.

components/globals/SecurityBadge.tsx
```diff
@@ -9,7 +9,11 @@ const SecurityAttributeBadge = (props: SecurityAttributeBadgeProps) => {
     : props.securityLevel + t("securityAttributeBadge.desc");
   return (
     <div className={props.className}>
-      <span className="gc-security-badge" aria-describedby="security-attribute-desc">
+      <span
+        id="security-attribute-content"
```
craigzour, line 13: If the `id` here is for testing only then I think you can use `data-testid` and then use it in the Cypress test.
(example: `cy.get("[data-testid='add-row-button-3']").click();`)

craigzour (changes requested):

craigzour (approved):

PR #739, opened 2022-04-25
DRAFT - Next Auth v4 / Prisma Migration

bryan-robitaille:
# Summary | Résumé
This PR updates Next-Auth to v4 and includes the following breaking changes:
- Prisma ORM must be used to manage data model migrations
- postgres-migrations will be a legacy feature and eventually removed.

Current base of this branch is `feature/nextjs_v12_upgrade` but will be rebased onto main once PR #725 is merged.
This PR will remain in draft until the last PR containing Database schema changes is merged.  At that point Prisma migration files will be generated.

Next-Auth v4 no longer supports direct management of the database and requires the project to leverage one of its 'Adapters'.  Prisma as an ORM is light weight and includes such features as connection and request pooling by default.  The existing queries to the Users table has been modified to now leverage Prisma.  All associated tests have also been modified to fit the new Prisma mocked instance.

 
# Pull Request Checklist


pages/api/token/temporary.ts
```diff
@@ -82,16 +82,43 @@ const createTemporaryToken = (email: string, tokenSecret: string): string => {
...
+            email: user.email,
+            templateId: user.templateId,
+          },
+        },
+        data: {
+          temporaryToken: temporaryToken,
+        },
+      });
+    } else {
+      throw new Error("User is not found or not active");
```
jeberhardt, line 114: Do these type of API errors need to also be in FR?

bryan-robitaille: Not that I'm aware of.  API errors are usually not internationalized which is why HTTP status codes are predominantly used.

prisma/schema.prisma
```diff
@@ -0,0 +1,85 @@
...
+}
+
+model FormUser {
+  id              String       @id @default(cuid())
+  templateId     String
+  email           String    @db.VarChar(80)
+  temporaryToken String?
+  active          Boolean   @default(true)
+  created_at      DateTime  @default(now()) @db.Timestamptz(6)
+  updated_at      DateTime  @default(now()) @db.Timestamptz(6)
```
falila, line 46: I wonder why you choose to set  `updated_at` field  default value to `now()` and make it required?  
Since  `updated_at` is not required when creating the entity , I think it should be marked as optional and only updatable. Like  `updated_at    DateTime? @updatedAt`

falila (changes requested):
That was the first class work. Terrific!

prisma/schema.prisma
bryan-robitaille: Good catch!  It was an oversight after the introspection of the existing Postgres Database Scheme.
I'll modify it to use the built-in `@updatedAt` Prisma attribute for the field.  I don't think the field needs to be marked as optional when set with the `@updatedAt` attribute but I'll definitely look into it.

jeberhardt (approved):

falila (approved):

falila (approved):

PR #416, opened 2021-09-07
Feature/ircc list integration

Moro-Code:
# Summary | Résumé

closes #407  

IRCC List Management API Integration. Please see README for instructions on how to set this up locally. The repo for the list management API can be found [here](https://github.com/cds-snc/list-manager).

# Test instructions | Instructions pour tester la modification

1. Run list manager locally 
2. Set up lists in list manager 
3. Test integration

# Help requested | Aide requise

Look at code ensure its sound :)

# Unresolved questions / Out of scope | Questions non résolues ou hors sujet
 
N/A

# Reviewer checklist | Liste de vérification du réviseur

This is a suggested checklist of questions reviewers might ask during their
review | Voici une suggestion de liste de vérification comprenant des questions
que les réviseurs pourraient poser pendant leur examen :

- [ ] Does this meet a user need? | Est-ce que ça répond à un besoin utilisateur?
- [ ] Is it accessible? | Est-ce que c’est accessible?
- [ ] Is it translated between both offical languages? | Est-ce dans les deux
      langues officielles?
- [ ] Is the code maintainable? | Est-ce que le code peut être maintenu?
- [ ] Have you tested it? | L’avez-vous testé?
- [ ] Are there automated tests? | Y a-t-il des tests automatisés?
- [ ] Does this cause automated test coverage to drop? | Est-ce que ça entraîne
      une baisse de la quantité de code couvert par les tests automatisés?
- [ ] Does this break existing functionality? | Est-ce que ça brise une
      fonctionnalité existante?
- [ ] Should this be split into smaller PRs to decrease change risk? | Est-ce
      que ça devrait être divisé en de plus petites demandes de tirage (« pull
      requests ») afin de réduire le risque lié aux modifications?
- [ ] Does this change the privacy policy? | Est-ce que ça entraîne une
      modification de la politique de confidentialité?
- [ ] Does this introduce any security concerns? | Est-ce que ça introduit des
      préoccupations liées à la sécurité?
- [ ] Does this significantly alter performance? | Est-ce que ça modifie de
      façon importante la performance?
- [ ] What is the risk level of using added dependencies? | Quel est le degré de
      risque d’utiliser des dépendances ajoutées?
- [ ] Should any documentation be updated as a result of this? (i.e. README
      setup, etc.) | Faudra-t-il mettre à jour la documentation à la suite de ce
      changement (fichier README, etc.)?

PR #754, opened 2022-04-29
Add CSRF protection

### LINKED ISSUE #716: Add CSRF to pages that include forms

falila:
Add Cross-Site Request Forgery (CSRF)
closed #716 


lib/middleware/middleware.ts
```diff
@@ -10,9 +10,13 @@ import { NextApiRequest, NextApiResponse } from "next";
...
   middlewareArray: Array<MiddlewareRequest>,
-  handler: (req: NextApiRequest, res: NextApiResponse, props: MiddlewareProps) => Promise<void>
+  handler: (
+    req: NextApiRequest,
+    res: NextApiResponse,
+    props: MiddlewareProps
+  ) => Promise<void | NodeJS.Timeout>
 ) => {
-  return async (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
+  return async (req: NextApiRequest, res: NextApiResponse): Promise<void | NodeJS.Timeout> => {
```
craigzour, line 19: I know the reason why you changed the signature of the `middleware` function but I am wondering if it is relevant because the `submit` function in `submit.ts` does not look like it could return such `Timeout` object.

That code was aded by @Moro-Code through https://github.com/cds-snc/platform-forms-client/pull/416/files#diff-09973ff12369d6c2b43bd018d2b6f4c9869ec4da7d37f1d7d1efe840dbd9abecR39 . Do you remember why @Moro-Code ?

We might be able to refactor it so that we just remove the `Timeout` type in both places?

falila: Gotcha!

falila: Gotcha!

lib/tests/helpers.test.js
```diff
@@ -216,4 +218,50 @@ describe("Submit and Template helpers", () => {
     expect(mockedFormikBag.setStatus).toBeCalledTimes(1);
     expect(mockedFormikBag.setStatus).toBeCalledWith("Error");
   });
+
+  test("Should submit with the correct csrf token", async () => {
```
bryan-robitaille, line 222: When going through the current code base I noticed Jest Tests is something that we really need to work on as a team.  I'll use this as an example but by no means is it the only one in our codebase.
The way the current test is written goes beyond the scope of what the client side function should test for.  The additional scope also doesn't add much value since all of the server side logic and responses are mocked out.
For example:
- the test mocks the axios server response as succeeded so `expect(mockedRouter.push).toBeCalledTimes(1)` will always evaluate to true.
- testing the payload `buildFormDataObject(TestForm, mockResponses)` is redundant and will always evaluate to true as it's relying on `buildFormDataObject` to check it's generated output against itself.

The test could be written to test solely that the Submit API function is correctly adding the token to the request as so:
```js
  test("Should submit with the correct csrf token", async () => {
    getCsrfToken.mockReturnValueOnce("csrfTokenSecretValue");

    mockedAxios.mockResolvedValue({
      status: 200,
      data: { received: true },
    });

    await submitToAPI(mockResponses, {
      props: {
        formConfig: TestForm,
        language: "en",
        router: { push: jest.fn() },
      },
      setStatus: jest.fn(),
    });
    expect(mockedAxios.mock.calls[0][0].headers["X-CSRF-Token"]).toBe("csrfTokenSecretValue");
  });

```

lib/tests/helpers.test.js
```diff
@@ -216,4 +218,50 @@ describe("Submit and Template helpers", () => {
...
+        language: "en",
+        router: mockedRouter,
+      },
+    });
+    expect(mockedAxios.mock.calls[0][0].headers["X-CSRF-Token"]).toBe("csrfTokenSecretValue");
+    expect(mockedAxios.mock.calls[0][0].data).toEqual(buildFormDataObject(TestForm, mockResponses));
+    expect(mockedRouter.push).toBeCalledTimes(1);
+  });
+
+  test("Should fail on submit", async () => {
```
bryan-robitaille, line 245: This type of test should probably be extracted to it's own separate test file for the csrfProtected middleware.
The csrfProtected middleware tests should cover:
- Middleware passes (correct token and allowed methods)
- Middleware fails (incorrect token and unallowed methods)
- Middleware returns correct errors (can be incorporated in Middleware Fails tests to check for correct errors being thrown)

bryan-robitaille (changes requested):
Nice use of leveraging existing next-auth CSRF functions and incorporating it into middleware.  Some use cases that we'll want to cover with this PR are:
- Allowed methods that we expect CSRF token to be on
- a separate testing file for csrfProtected that covers passing, failing, and error checking

lib/middleware/csrfProtected.ts
```diff
@@ -0,0 +1,28 @@
...
+      if (allowMethods && req.method && allowMethods.includes(req.method)) {
+        const csrfToken = await getCsrfToken({ req });
+        if (!csrfToken) throw Error("Invalid Csrf Token found");
+        if (csrfToken === req.headers["x-csrf-token"]) {
+          return { next: true };
+        } else {
+          res.status(403).json({ error: "Access Denied" });
+          return { next: false };
+        }
+      } else {
```
bryan-robitaille, line 18: If the method is not CSRF protected then it should pass through and not throw a "Method is not Allowed".  The CORS middleware already checks for which methods the API endpoint will accept.  Similar to the 'sessionExists' middleware it should only apply it's logic to specific declared methods.

lib/tests/csrfProtected.test.js
```diff
@@ -0,0 +1,152 @@
...
+      0: "Seat",
+      1: "Love",
+      2: "CHMC",
+      3: ["Yes", "Other"],
+    },
+  ],
+  8: "",
+};
+
+describe("Csrf Protection middleware", () => {
```
bryan-robitaille, line 35: Just a note that CSRF is usually only applied to state changing API calls (POST, PUT, DELETE)

lib/tests/csrfProtected.test.js
```diff
@@ -0,0 +1,152 @@
...
+  beforeEach(() => {
+    mockedConsoleError = jest.spyOn(logMessage, "error");
+    mockedConsoleError.mockImplementation(() => {});
+  });
+  afterEach(() => {
+    mockedConsoleError.mockRestore();
+    getCsrfToken.mockRestore();
+  });
+
+  test("Should fail with code 405 Method Not Allowed", async () => {
```
bryan-robitaille, line 46: For example if the mock request is "GET" and only the "POST" path is CSRF protected then the status code should be 200.

bryan-robitaille (changes requested):
A logic change to ensure that unprotected CSRF paths are passed through and aligning the associated tests to that logic should be all that's left.

lib/middleware/csrfProtected.ts
```diff
@@ -0,0 +1,28 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import { MiddlewareRequest, MiddlewareReturn } from "@lib/types";
+import { getCsrfToken } from "next-auth/client";
+import { logMessage } from "@lib/logger";
+
+export const csrfProtected = (allowMethods: string[]): MiddlewareRequest => {
+  return async (req: NextApiRequest, res: NextApiResponse): Promise<MiddlewareReturn> => {
+    try {
+      if (allowMethods && req.method && allowMethods.includes(req.method)) {
```
bryan-robitaille, line 9: Just a quick note that we don't need to check if 'allowedMethods' is 'falsey' because in the typescript declaration it's specified that 'allowedMethods' must be a String Array.

lib/middleware/csrfProtected.ts
```diff
@@ -3,21 +3,22 @@ import { MiddlewareRequest, MiddlewareReturn } from "@lib/types";
 import { getCsrfToken } from "next-auth/client";
 import { logMessage } from "@lib/logger";
 
-export const csrfProtected = (allowMethods: string[]): MiddlewareRequest => {
+export const csrfProtected = (protectedMethods: string[]): MiddlewareRequest => {
   return async (req: NextApiRequest, res: NextApiResponse): Promise<MiddlewareReturn> => {
     try {
-      if (allowMethods && req.method && allowMethods.includes(req.method)) {
+      if (isProtected(req, protectedMethods)) {
```
bryan-robitaille, line 9: This logic can be simplified a bit.  If a CSRF header is not present it should be treated as unauthorized and not throw a 500 server error.  If as Cross site attack vector were to happen and a CSRF token was not provided the current logic would flood our logs with Errors and trigger Application Alarms that something was wrong with the site.  In this case the site is behaving normally by blocking non csrf and invalid csrf headers.
```js
    try {
      if (isProtected(req, protectedMethods)) {
        const csrfToken = await getCsrfToken({ req });

        if (csrfToken && csrfToken === req.headers["x-csrf-token"]) {
          return { next: true };
        } else {
          res.status(403).json({ error: "Access Denied" });
          return { next: false };
        }
      } else {
        //allow unrestricted method
        return { next: true };
      }
    } catch (error) {
      logMessage.error(error);
      res.status(500).json({ error: "Malformed API Request" });
      return { next: false };
    }
```

bryan-robitaille (changes requested):
A quick change to ensure that Unauthorized attempts are not logged as server errors and we should be good to approve :)

lib/middleware/csrfProtected.ts
```diff
@@ -0,0 +1,31 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import { MiddlewareRequest, MiddlewareReturn } from "@lib/types";
+import { getCsrfToken } from "next-auth/client";
+
+export const csrfProtected = (protectedMethods: string[]): MiddlewareRequest => {
+  return async (req: NextApiRequest, res: NextApiResponse): Promise<MiddlewareReturn> => {
+    try {
+      if (isProtected(req, protectedMethods)) {
+        const csrfToken = await getCsrfToken({ req });
+        if (!csrfToken || csrfToken !== req.headers["x-csrf-token"]) {
```
bryan-robitaille, line 10: I'm curious as to why you decided to inverse the logic of the token check

from:  if(nextAuth is funtional and token is good) pass else block everything

to:  if(nextAuth is not functional or token is not good) block else pass everything

From a readability standpoint the inverse negation logic made the statement difficult for me to follow.  The default to pass everything that doesn't meet the statement criteria also seems like an anti-pattern for access control type logic.

lib/tests/csrfProtected.test.js
```diff
@@ -0,0 +1,149 @@
...
+      1: "Love",
+      2: "CHMC",
+      3: ["Yes", "Other"],
+    },
+  ],
+  8: "",
+};
+
+describe("Csrf Protection middleware", () => {
+  let mockedConsoleError;
```
bryan-robitaille, line 36: Since logMessage.error was removed from the csrfProtected function we should also remove the jest spy on the function from the tests.

lib/tests/csrfProtected.test.js
```diff
@@ -0,0 +1,149 @@
...
+    });
+    jest.spyOn(logMessage, "error");
+
+    await csrfProtected(["POST"])(req, res);
+    expect(res.statusCode).toBe(200);
+  });
+});
+
+describe("Submit a form with csrf protection enable", () => {
+  let mockedConsoleError;
```
bryan-robitaille, line 105: Since logMessage.error was removed from the csrfProtected function we should also remove the jest spy on the function from the tests.

bryan-robitaille (changes requested):
A little cleanup required in the test file to align with the changes made to csrfProtected middleware.

bryan-robitaille (approved):
Thank you for the changes.
LGTM 😃

PR #758, opened 2022-05-04
build: fix staging-deploy Github action

craigzour:
# Summary | Résumé

- Trying to fix a recurring issue we have with our staging deployment script. @patheard suggested a new way of waiting for an action to complete before triggering another one.

### DIFF
--- a/.github/workflows/staging-build-push-container.yml
+++ b/.github/workflows/staging-build-push-container.yml
@@ -1,4 +1,4 @@
-name: Staging — Push container to ECR
+name: "Staging — Push container to ECR"
 
 on:
   push:
--- a/.github/workflows/staging-deploy.yml
+++ b/.github/workflows/staging-deploy.yml
@@ -1,24 +1,27 @@
 name: Deploy to Staging
 
 on:
-  push:
-    branches: [main]
+  workflow_run:
+    workflows: ["Staging — Push container to ECR"]
+    types:
+      - completed
 
 jobs:
-  deploy-form-viewer-service:
+  unable-to-deploy:
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
     runs-on: ubuntu-latest
     steps:
-      - name: Wait for container to be built and pushed
-        # v1.0.0 of this action as of Jan 28 2021
-        uses: fountainhead/action-wait-for-check@4699210ccc66e2a13260803fadbb77085421b891
-        id: wait-for-build
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          checkName: push-staging
-          ref: ${{ github.sha }}
+      - name: Post to Slack
+        run: |
+          json='{"blocks":[{"type":"section","text":{"type":"mrkdwn","text":":red: Unable to deploy Staging form viewer: <https://github.com/cds-snc/platform-forms-client/actions/workflows/staging-build-push-container.yml|ECR push failed>"}}]}'
+          curl -X POST -H 'Content-type: application/json' --data "$json" ${{ secrets.FORMS_DEV_SLACK_WEBHOOK }}
+          exit 1
 
+  deploy-form-viewer-service:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-latest
+    steps:
       - name: Checkout
-        if: steps.wait-for-build.outputs.conclusion == 'success'
         uses: actions/checkout@v2
 
       - name: Configure AWS credentials
@@ -38,11 +41,13 @@ jobs:
         id: cluster
         run: |
           echo "##[set-output name=name;]$(aws ecs list-clusters | jq -r '.clusterArns[0]' | cut -f 2 -d "/" )"
+
       - name: Download Form Viewer task definition
         id: download-taskdef-form-viewer
         run: |
           aws ecs describe-task-definition --task-definition form-viewer --query taskDefinition > form_viewer.json
           echo "##[set-output name=container_name;]$(jq -r '.containerDefinitions[0].name' form_viewer.json)"
+
       - name: Render image for form viewer service
         id: taskdef-form-viewer
         # v1.0.10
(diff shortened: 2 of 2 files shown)


patheard (changes requested):
Just two minor tweaks from my end.  

1. There are a few spots in this workflow where `${{ github.sha }}` is referenced.  Since this workflow is now being triggered by another workflow being complete, I think it would be safer to use the [`workflow_run` payload](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_run) to get that SHA: 
```
${{ github.event.workflow_run.head_sha }}
```

2. If you wanted, you could also add a check to fail early if the `push-staging` workflow fails, rather than trying to deploy the failed image push.  A simple example would be just to add a `on-failure` job to the workflow:
```yaml
jobs:
  on-success:
    runs-on: ubuntu-latest
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
    steps:
      - run: echo 'The triggering workflow passed'
   
  on-failure:
    runs-on: ubuntu-latest
    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
    steps:
      - run: echo 'The triggering workflow failed'
```

craigzour:
> ```yaml
> on-failure
> ```

Gotcha! Just a question about your second point. How would the `staging-deploy` action not be fully executed just because we have this `on-failure` step? How does it work?

patheard:
> Gotcha! Just a question about your second point. How would the `staging-deploy` action not be fully executed just because we have this `on-failure` step? How does it work? cc @patheard

This would be to catch cases like what happened yesterday where the `push-staging` job timed out and failed.  In that case, you wouldn't want to try to run `staging-deploy` to deploy a new image (since it wouldn't exist).

In the `on-failure` job, you could do something like [use a Slack webhook](https://github.com/cds-snc/notification-terraform/blob/main/.github/workflows/merge_to_main_staging.yml#L127-L131) to post back to the `#forms-dev` channel to alert people that the deploy job didn't work as expected.

craigzour:
> > Gotcha! Just a question about your second point. How would the `staging-deploy` action not be fully executed just because we have this `on-failure` step? How does it work? cc @patheard
> 
> This would be to catch cases like what happened yesterday where the `push-staging` job timed out and failed. In that case, you wouldn't want to try to run `staging-deploy` to deploy a new image (since it wouldn't exist).
> 
> In the `on-failure` job, you could do something like [use a Slack webhook](https://github.com/cds-snc/notification-terraform/blob/main/.github/workflows/merge_to_main_staging.yml#L127-L131) to post back to the `#forms-dev` channel to alert people that the deploy job didn't work as expected.

Oh ok so we would add those two "steps" in the `staging-build-push-container` action then? The run command  with the `echo` will be what triggers or not the execution of the `staging-deploy` action?

patheard:
I'd been thinking more that you could add them to the `.github/workflows/staging-deploy.yml` workflow like so:

```yaml
name: Deploy to Staging

on:
  workflow_run:
    workflows: ["Staging — Push container to ECR"]
    types:
      - completed

jobs:
  # This job will only trigger if the staging-build-push-container workflow is unsuccessful.
  # It will post a message to Slack to notify the team there was a problem. 
  # You'll need a `FORMS_DEV_SLACK_WEBHOOK` GitHub secret that can post to the #forms-dev channel to set this up.
  unable-to-deploy:
    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
    runs-on: ubuntu-latest
    steps:
      - name: Post to Slack
        run: |
          json='{"blocks":[{"type":"section","text":{"type":"mrkdwn","text":":red: Unable to deploy Staging form viewer: <https://github.com/cds-snc/platform-forms-clients/actions/workflows/staging-build-push-container.yml|ECR push failed>"}}]}'
          curl -X POST -H 'Content-type: application/json' --data "$json" ${{ secrets.FORMS_DEV_SLACK_WEBHOOK }} 
          exit 1

  # If staging-build-push-container was successful, proceed with the deployment.
  deploy-form-viewer-service:
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v2

    # remaining deploy steps

```
Happy to chat more tomorrow if you've still got questions.

craigzour:
> I'd been thinking more that you could add them to the `.github/workflows/staging-deploy.yml` workflow like so:
> 
> ```yaml
> name: Deploy to Staging
> 
> on:
>   workflow_run:
>     workflows: ["Staging — Push container to ECR"]
>     types:
>       - completed
> 
> jobs:
>   # This job will only trigger if the staging-build-push-container workflow is unsuccessful.
>   # It will post a message to Slack to notify the team there was a problem. 
>   # You'll need a `FORMS_DEV_SLACK_WEBHOOK` GitHub secret that can post to the #forms-dev channel to set this up.
>   unable-to-deploy:
>     if: ${{ github.event.workflow_run.conclusion == 'failure' }}
>     runs-on: ubuntu-latest
>     steps:
>       - name: Post to Slack
>         run: |
>           json='{"blocks":[{"type":"section","text":{"type":"mrkdwn","text":":red: Unable to deploy Staging form viewer: <https://github.com/cds-snc/platform-forms-clients/actions/workflows/staging-build-push-container.yml|ECR push failed>"}}]}'
>           curl -X POST -H 'Content-type: application/json' --data "$json" ${{ secrets.FORMS_DEV_SLACK_WEBHOOK }} 
>           exit 1
> 
>   # If staging-build-push-container was successful, proceed with the deployment.
>   deploy-form-viewer-service:
>     if: ${{ github.event.workflow_run.conclusion == 'success' }}
>     runs-on: ubuntu-latest
>     steps:
>       - name: Checkout
>         uses: actions/checkout@v2
> 
>     # remaining deploy steps
> ```
> 
> Happy to chat more tomorrow if you've still got questions.

Got it now! Stil need to Github actions and I got mixed up by the `job` aspect vs the `step` one. Going to update the PR soon. Thanks :)

craigzour:
@patheard  New Slack secret has been added here https://github.com/cds-snc/platform-forms-client/settings/secrets/actions

PR #790, opened 2022-05-18
docs: updated env example when testing using localstack

mohamed-cds:
Adding more examples on whats needed to run locally using localstack


Moro-Code (approved):
Thanks Mohamed this is great !

PR #791, opened 2022-05-18
fix: add `'limit` and `offset` the the crudTemplates call, retrieving the templates in batches

### LINKED ISSUE #786: Template GET All limit reached
The Template Lambda is erroring out when trying to retrieve all template configs.  The resulting set is larger than 1MB which cannot be handled by the RDS Data API.

No PR's for the application will be able to be successfully merged until resolved.  The cypress tests and Sandbox page will continue to throw errors until resolved.

This will be a temporary solution as #347 will move the CRUD functionality from Lambda to NextJS.

Acceptance Criteria:
- [ ] Template Lambda returns all form config (same functionality as current)

Potential Solutions:
 - In Lambda use Data API to get number of records and then use LIMIT and OFFSET to build return object.

jeberhardt:
# Summary | Résumé

Closes #786 

Created a recursive function to retrieve the form templates in batches. This adds parameters for `limit` and `offset`, to request only some form templates in each call to the lambda function, ensuring that each batch is below the 1MB limit. It will merge all the batches together, and return all templates back to the calling function.

- PR 211 on the `forms-terraform` repo will need to be merged in first in order to test.
- The database will need to have more than 50 forms in the `templates` table in order to test this.
- Visit the `/admin/view-templates` page, and observe that all templates are returned.
- All jest and cypress tests should pass too.

What should the default `limit` be? Currently set at 50, but I suspect it could be quite a bit higher than that, and still be under the 1MB limit.

# Pull Request Checklist

### DIFF
--- a/lib/integration/crud.ts
+++ b/lib/integration/crud.ts
@@ -250,11 +250,19 @@ async function _getFormByStatus(
  * Recursively calls lambda function to retrieve the forms (templates) from the database
  * @param templates the array which the records will be added to, then recursively passed to the next `getForms()` call
  * @param limit the number of records to fetch from the database
- * @param offest the record to start from
+ * @param offset the record to start from
  * @returns an array of all the forms (templates) from the database
  */
 
-const _getForms = async (templates: unknown[] = [], limit = 50, offset = 0): Promise<unknown[]> => {
+const _getForms = async (
+  templates: CrudTemplateResponse = {
+    data: {
+      records: [],
+    },
+  },
+  limit = 50,
+  offset = 0
+): Promise<CrudTemplateResponse> => {
   try {
     const lambdaResult = await crudTemplates({ method: "GET", limit: limit, offset: offset });
 
@@ -262,7 +270,15 @@ const _getForms = async (templates: unknown[] = [], limit = 50, offset = 0): Pro
       return templates;
     }
 
-    templates = templates.concat(lambdaResult.data.records);
+    if (templates.data.records) {
+      templates.data.records = templates.data.records.concat(lambdaResult.data.records);
+    } else {
+      templates = {
+        data: {
+          records: [...lambdaResult.data.records],
+        },
+      };
+    }
 
     if (lambdaResult.data.records.length === limit) {
       // There could be more records in the database, so get the next batch
@@ -272,7 +288,7 @@ const _getForms = async (templates: unknown[] = [], limit = 50, offset = 0): Pro
     }
   } catch (e) {
     logMessage.error(e as Error);
-    return [];
+    return templates;
   }
 };
 
--- a/lib/integration/crud.ts
+++ b/lib/integration/crud.ts
@@ -262,7 +262,7 @@ const _getForms = async (templates: unknown[] = [], limit = 50, offset = 0): Pro
       return templates;
     }
 
-    Array.prototype.push.apply(templates, lambdaResult.data.records);
+    templates = templates.concat(lambdaResult.data.records);
 
     if (lambdaResult.data.records.length === limit) {
       // There could be more records in the database, so get the next batch
--- a/lib/integration/crud.ts
+++ b/lib/integration/crud.ts
@@ -246,6 +246,28 @@ async function _getFormByStatus(
   return [];
 }
 
+const _getForms = async (templates: unknown[] = [], limit = 50, offset = 0): Promise<unknown[]> => {
+  try {
+    const lambdaResult = await crudTemplates({ method: "GET", limit: limit, offset: offset });
+
+    if (!lambdaResult?.data?.records) {
+      return templates;
+    }
+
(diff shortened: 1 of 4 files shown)


pages/admin/view-templates.tsx
```diff
@@ -9,14 +9,23 @@ export const getServerSideProps = requireAuthentication(async (context) => {
...
-    const templatesJSON =
-      lambdaResult &&
-      lambdaResult.data &&
-      lambdaResult.data.records &&
-      lambdaResult.data.records.length > 0
-        ? lambdaResult.data.records
-        : [];
+    const templatesJSON: unknown[] = [];
+
+    const getTemplatesFromLambda = async (limit = 13, offset = 0) => {
```
Moro-Code, line 14: This seems like a function thats better defined in the lib/integration/crud.ts file

pages/admin/view-templates.tsx
```diff
@@ -9,14 +9,23 @@ export const getServerSideProps = requireAuthentication(async (context) => {
...
+      const lambdaResult = await crudTemplates({ method: "GET", limit: limit, offset: offset });
+
+      if (!lambdaResult || !lambdaResult.data || !lambdaResult.data.records) {
+        return;
+      }
+
+      Array.prototype.push.apply(templatesJSON, lambdaResult.data.records);
+
+      if (!(lambdaResult.data.records.length < limit)) {
```
Moro-Code, line 23: Inverse logic makes it difficult for readability. Easier as `if(lambdaResult.data.records.length >= limit)`

Moro-Code: I'm not sure I understand the logic here. If the number of records returned from the lambda is greater than the limit we recurse. So if 15 records are returned and the limit is 13 . The offset will be set to 13 on the first recursion meaning that we will have a duplication of two records. Furthermore shouldn't we only recurse if we get the number of records below the limit ?

Moro-Code (changes requested):
Good work. Just a comment on the logic here and some readability items

pages/admin/view-templates.tsx
```diff
@@ -9,14 +9,23 @@ export const getServerSideProps = requireAuthentication(async (context) => {
...
+    const templatesJSON: unknown[] = [];
+
+    const getTemplatesFromLambda = async (limit = 13, offset = 0) => {
+      const lambdaResult = await crudTemplates({ method: "GET", limit: limit, offset: offset });
+
+      if (!lambdaResult || !lambdaResult.data || !lambdaResult.data.records) {
+        return;
+      }
+
+      Array.prototype.push.apply(templatesJSON, lambdaResult.data.records);
```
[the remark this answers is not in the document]
Moro-Code: I think concat should be sufficient. I'm also not exactly sure what you mean by deep copy. Doing a push on the individual items would not do a deep copy either . concat returns the new array which you would set

lib/integration/crud.ts
```diff
@@ -244,6 +246,28 @@ async function _getFormByStatus(
...
 
+const _getForms = async (templates: unknown[] = [], limit = 50, offset = 0): Promise<unknown[]> => {
+  try {
+    const lambdaResult = await crudTemplates({ method: "GET", limit: limit, offset: offset });
+
+    if (!lambdaResult?.data?.records) {
+      return templates;
+    }
+
+    Array.prototype.push.apply(templates, lambdaResult.data.records);
```
Moro-Code, line 257: Since you resolved the convo for visibility I will re-comment here. I'm not sure exactly what you mean by "concat doesn't do a deep copy`. Concat returns the amalgamation of two arrays which is completely new. If you're talking about the individual items I don't think applying the push function on each of the items would result in a deep copy of the individual items.

Moro-Code: This would be much simpler to read 

`templates = templates.concat(lambdaResult.data.records);`

lib/integration/crud.ts
```diff
@@ -244,6 +246,28 @@ async function _getFormByStatus(
   return [];
 }
 
+const _getForms = async (templates: unknown[] = [], limit = 50, offset = 0): Promise<unknown[]> => {
```
Moro-Code, line 249: Although the intention of this function is not to stay long. A doc string would be much appreciated so we can remind ourselves as to the purpose of it

jeberhardt: I think the lack of docstrings in the rest of the file caused me to overlook this.  ;)

lib/integration/crud.ts
jeberhardt: Ack! I mis-remembered that `.concact` does not affect the original array, in the way that `.push` does. 

From the docs:`This method does not change the existing arrays, but instead returns a new array.`

craigzour (dismissed):

Moro-Code (dismissed):

Moro-Code (approved):
Looks good now !

PR #792, opened 2022-05-19
fix: update devcontainer to not use root user

mohamed-cds:
Currently the devcontainer is using the incorrect user, so the shell is defaulting to the root user. This PR fixes that, as well as introducing a devcontainer helper post create script to setup the basics.


bryan-robitaille (approved):
LGTM

PR #802, opened 2022-05-25
feat: generate Software Bill of Materials

patheard:
# Summary | Résumé

Add GitHub workflows to generate SBOMs and upload them
to Dependency-Track. This will allow us to have a view of all
the dependencies used by the org and their potential
vulnerabilities.

# Test instructions | Instructions pour tester la modification

1. Check that a NodeJS SBOM is generated by the creation and merge of this PR.
2. Check that a Docker SBOM is generated when this PR is merged.

# Unresolved questions / Out of scope | Questions non résolues ou hors sujet

n/a

# Pull Request Checklist

Please complete the following items in the checklist before you request a review:

- [x] Have you completely tested the functionality of change introduced in this PR? Is the PR solving the problem it's meant to solve within the scope of the related issue?
- [x] The PR does not introduce any new issues such as failed tests, console warnings or new bugs.
- [ ] ~~If this PR adds a package have you ensured its licensed correctly and does not add additional security issues?~~
- [x] Is the code clean, readable and maintainable? Is it easy to understand and comprehend.
- [x] Does your code have adequate comprehensible comments? Do new functions have [docstrings](https://tsdoc.org/)?
- [ ] ~~Have you modified the change log and updated any relevant documentation?~~
- [ ] ~~Is there adequate test coverage? Both unit tests and end-to-end tests where applicable?~~
- [ ] ~~If your PR is touching any UI is it accessible? Have you tested it with a screen reader? Have you tested it with automated testing tools such as axe?~~

PR #808, opened 2022-05-26
fix: form viewer Docker SBOM generate

patheard:
# Summary
Add the missing `ECR_REGISTRY` environment variable that's used
to identify the Docker image that will be used to generate the SBOM.

# Related
* #802


falila (approved):
🌮

PR #817, opened 2022-06-02
fix: remove service links from devcontainer docker-compose

patheard:
# Summary
Update the devcontainer docker-compose to no longer
explicitly link the services.  

This is because `links` are no longer required as all 
services implicitly share the same Docker network.


mohamed-cds (approved):
LGTM since i added those in 🚀 ✅

mohamed-cds (approved):

PR #822, opened 2022-06-07
fix: hydration errors due to react tree and dom being out of sync

mohamed-cds:
Tracked down two react tree and DOM sync errors that was visible in [staging](https://reactjs.org/docs/error-decoder.html/?invariant=418) and the local environment. Not happening in prod which i suspect is related to using a pre-built version of the code.

According to [Nextjs](https://nextjs.org/docs/messages/react-hydration-error) in order to prevent the first render from being different you can use `useEffect` which is only executed in the browser and is executed during hydration.

I tested locally and the hydration errors are no longer visible after applying the changes to the two components with the error.

image

*Note*: I think the eslint rules came into effect after these files were created so i had to make some formatting changes so that these files comply with the new lint rules.

### DIFF
--- a/components/globals/Base.js
+++ b/components/globals/Base.js
@@ -26,29 +26,31 @@ const Base = ({ children }) => {
   const { t } = useTranslation("common");
 
   return (
-    mounted && (
-      <>
-        <Head>
-          <title>{t("title")}</title>
-          <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
-          <meta charSet="utf-8" />
-          <link rel="shortcut icon" href="/favicon.ico" type="image/x-icon" sizes="32x32" />
-        </Head>
+    <>
+      {!mounted ? null : (
+        <>
+          <Head>
+            <title>{t("title")}</title>
+            <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
+            <meta charSet="utf-8" />
+            <link rel="shortcut icon" href="/favicon.ico" type="image/x-icon" sizes="32x32" />
+          </Head>
 
-        <SkipLink />
-        <div className={classes}>
-          {!isEmbeddable && (
-            <header>
-              {shouldDisplayAlphaBanner && <PhaseBanner />}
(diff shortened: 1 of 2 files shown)


patheard (commented):
Super nice 👍🏻 

I'll let a Forms team member approve just to make sure there's not a React gotcha I don't know about sneaking in with the change.

bryan-robitaille:
I haven't commented on the PR yet because I'm digging a little more into it.  I can't replicate the issue locally so I'm wondering if it's only happening when running the app in a docker image.

bryan-robitaille:
So the hydration errors only appear for static pages when running the application in production mode (yarn build && yarn start).  This is because during the build process the static pages are built without content due to the ISOLATED_INSTANCE env var.  The ISOLATED_INSTANCE environment var was needed in order to prevent server content being generated when no database was available (a la github workflow for Cypress E2E tests).  I'll try to rework the logic to get rid of the hydration errors and ensure that static pages where content is available are properly built.

mohamed-cds:
Thanks @bryan-robitaille. If it helps. I'm running it locally using `yarn dev`

bryan-robitaille:
@mohamed-cds is it possible that in your .env you have ISOLATED_INSTANCE set ?  If so the error would also occur using `yarn dev`

mohamed-cds:
> @mohamed-cds is it possible that in your .env you have ISOLATED_INSTANCE set ? If so the error would also occur using `yarn dev`

@bryan-robitaille i just checked and it sets to true. I deleted the variable (setting to false didn't work) and now the hydration errors are no longer visible when testing locally. Do you want me to raise an issue for staging and close this PR since it's no longer affecting local testing but staging still needs the rework you mentioned ?

bryan-robitaille:
👍 
@mohamed-cds if you could close the PR and raise the issue for Staging it would be greatly appreciated.

mohamed-cds:
Removing the `ISOLATED_INSTANCE` fixes the issue locally. Issue https://github.com/cds-snc/platform-forms-client/issues/824 created to track the issue for the staging environment

PR #845, opened 2022-06-24
Fix/retrieval api

bryan-robitaille:
# Summary | Résumé
This PR creates an example Node script that can be used to retrieve form responses from the Vault and save the output into a CSV file.

This PR also fixes a logic flaw in the retrieval API where `accumulatedResponses` was always being written to because even if no Items matched the requested filter `[]` an empty array is a `truthy` value.

## Terminal 1
`yarn dev`
open `localhost:3000` and submit several responses of a form that uses the Vault.
## Terminal 2
1. `cd ./utils/retrievalAPI`
2. `yarn install`
3. `yarn retrieve` and follow the prompts

# Pull Request Checklist


pages/api/id/[form]/retrieval.ts
```diff
@@ -46,6 +46,34 @@ function connectToDynamo(): DynamoDBDocumentClient {
   return DynamoDBDocumentClient.from(db);
 }
 
+interface dynamoDBItem {
```
Comment (Moro-Code) [inline pages/api/id/[form]/retrieval.ts:49]: This interface is not correct. We're using the DynamoDB document client here which allows us to work with native JS types

pages/api/id/[form]/retrieval.ts
```diff
@@ -46,6 +46,34 @@ function connectToDynamo(): DynamoDBDocumentClient {
...
+  FormSubmission: { S: string };
+  [key: string]: Record<string, string>;
+}
+
+/**
+ * Parses DynamoDB Item into vault response object
+ * @param response DynamoDB Vault Item
+ * @returns Object representative of Item
+ */
+const parseToObject = (response: dynamoDBItem) => {
```
Comment (Moro-Code) [inline pages/api/id/[form]/retrieval.ts:61]: Seems like this could be a one liner in the map function

utils/retrievalAPI/retrieveResponse.ts
```diff
@@ -0,0 +1,138 @@
...
+  } catch (e) {
+    console.log("Error retrieving responses");
+    throw e;
+  }
+}
+
+const main = async () => {
+  try {
+    const email = await getValue("Email address to authenticate with:");
+    const formID = await getValue("Form ID of responses to retrieve:");
```
Moro-Code, line 81: Think it might be worth adding a prompt for the user to accept that in so doing it will remove the responses from our system. Realistically we are the only ones that will use this however, this is open source and so potentially a team working with a form on production could point this script there and use it

Moro-Code (commented):
Like the utility! Three comments. The second one on the function is more of a style comment so its not essential

pages/api/id/[form]/retrieval.ts
Comment (bryan-robitaille) [reply pages/api/id/[form]/retrieval.ts:61]: I had originally seperated it out for easier readability but I've moved it to be directly within the map function.

utils/retrievalAPI/retrieveResponse.ts
bryan-robitaille: Good idea.  Added prompt to confirm.

PR #884, opened 2022-07-19
feat: Login Page - State 1

### LINKED ISSUE #867: Build Login Page - State 1
A new NextJS page will be created that handles a custom implementation of Next Auth v4.  The page will be located at `{base_url}/auth/login` and completes both states of the login process.

The first state of the Login process will reuse existing functionality by leveraging the `{base_url}/api/token/temporary` API endpoint.  A form on the login page will accept a Form Key and email address and POST the information to the endpoint.

This state is part of a larger [Login Flow](https://docs.google.com/document/d/1ua3QfIHAcZwwodgGN2alkVZa5brgHr-P0Mw85gujrz8/edit?lfhs=2#heading=h.6ox7ti5fg93o).  Please review the complete flow before working on this page.  

## Designs for Login State 1:
- [Form](https://www.figma.com/file/ZPt6quJaJ2bVYZboCccWJs/Retrieval-UI?node-id=633%3A944)
- [Error State](https://www.figma.com/file/ZPt6quJaJ2bVYZboCccWJs/Retrieval-UI?node-id=638%3A1016)

## Out of Scope:
- State switching between State 1 and State 2
- 'Request a Login Reset' functionality (link must exist on page)

## Acceptance Criteria:
- [x] State 1 is a separate component from the Login Page itself
- [x] Component follows design as presented in Figma
- [x] Component is accessible
- [x] Component handles error state on incorrect credentials

jeberhardt:
# Summary | Résumé

Closes #867 

| Insert Image showing before the change was made | 
Screen Shot 2022-07-21 at 3 43 41 PM
 |

- load the page `/auth/login`
- enter an email address for a form
- enter the sign-in key for the form
- hit submit
- a successful response from the server will be seen in the developer panel for the network traffic

- try submitting a blank form (an error message should appear)
- try submitting a form with a non-GC email address (an error message should appear)
- try submitting a form with the correct email, but an incorrect sign-in key (an error message should appear)

Note: at this point, entering a successful email/sign-in key does not proceed to the next page

# Pull Request Checklist


bryan-robitaille (changes requested):
I wont' comment on the Errors display/layout at this point because I believe design is looking into changes?  That being said if we're keeping an Error block/list at the top of the page we should incorporate the shift focus and scroll to the Error block.  This is a pattern that can already be found on our Form's page.
There's some HTML sematics to change but looking forward to the next version and potentially new Error states.

falila (approved):
Only thing left is translating login.json into fr and I can help with that.
Great work!  🥐

components/login/SignInKey.test.js
```diff
@@ -0,0 +1,84 @@
...
+    mockedAxios.mockResolvedValue({
+      status: 200,
+    });
+
+    await act(async () => {
+      render(<SignInKey />);
+    });
+
+    const loginEmail = screen.getByTestId("loginEmail");
+    await userEvent.type(loginEmail, "test@cds-snc.ca");
```
craigzour, line 77: Should not the `form sign-in key` also be passed in order to get a successful response from the server (if both are valid of course) ? If the test is valid I guess it means that there is no empty fields validation when the form is submitted.

public/static/locales/fr/login.json
```diff
@@ -0,0 +1,11 @@
+{
+    "title": "Sign in",
```
craigzour, line 2: Are we waiting for a proper translation or should we handle it ourselves?

components/login/SignInKey.tsx
```diff
@@ -0,0 +1,110 @@
...
+          data-testid="signInKey"
+          onChange={handleChange}
+        />
+        <br />
+        <Button type="submit" testid="add-email">
+          Continue
+        </Button>
+      </form>
+      <br />
+      {t("signInKeyForgot")}
```
craigzour, line 103: I have seen a comment from @bryan-robitaille about the fact that it would be great to have links for the `forgot` and `reset` features. Even if they are not working yet.

components/login/SignInKey.tsx
```diff
@@ -0,0 +1,110 @@
...
+    });
+  };
+
+  const handleLoginSubmit = async (e: React.FormEvent) => {
+    e.preventDefault();
+    setErrorState({ message: "" });
+    const loginEmail = formData["loginEmail"];
+    if (isValidGovEmail(loginEmail, emailDomainList.domains)) {
+      try {
+        await axios({
```
craigzour, line 42: If both the email and the form sign-in key (form bearer token) are valid but the request for sending the temporary token failed because the server just did not succeed it does not mean that the credentials are invalid. It may be something to discuss with @connorscarolyns in case we want to display a different error.

components/login/SignInKey.tsx
```diff
@@ -0,0 +1,110 @@
...
+          timeout: process.env.NODE_ENV === "production" ? 60000 : 0,
+        });
+      } catch (err) {
+        setErrorState({
+          message: t("loginSignInErrorMessage"),
+        });
+      }
+    } else {
+      setErrorState({
+        message: t("loginSignInErrorMessage"),
```
craigzour, line 61: Did not you want to use `loginEmailErrorMessage` here? I noticed that even if my email is valid (`clement.janin@cds-snc.ca`) it displays both error messages so it may be why it does that.

craigzour (changes requested):
Added some comments/questions. Let me know what you think. Could be irrelevant because I may be missing some context.

components/login/SignInKey.test.js
jeberhardt: The client-side validation checks to see if the email is a GC email, which is why it is needed to execute this test. There isn't an empty field validation.

public/static/locales/fr/login.json
jeberhardt: Waiting on translation.

components/login/SignInKey.tsx
jeberhardt: Well, we don't actually know if your email is valid. We aren't checking if that user exists in the DB, we are only checking if both the email and token (together) exist. I don't think it's good practice to say: "Your email is correct, but that's a wrong password", giving potential hackers a piece of information.

components/login/SignInKey.tsx
craigzour: @jeberhardt  I think you missed my comment right above ^

components/login/SignInKey.tsx
```diff
@@ -0,0 +1,116 @@
...
+          data-testid="signInKey"
+          onChange={handleChange}
+        />
+        <br />
+        <Button type="submit" testid="add-email">
+          Continue
+        </Button>
+      </form>
+      <br />
+      {t("signInKeyForgot")}
```
craigzour, line 109: @bryan-robitaille also wanted a link on the `forgot` option here

craigzour (changes requested):

components/login/SignInKey.tsx
jeberhardt: I think we'll need a different way to present that type of error, because it doesn't fit with the model of "Problems on this page". It's not a problem that the user can fix by filling in the form differently.

components/login/SignInKey.tsx
jeberhardt: I don't think so. It's not that way in the design doc. I just think the link is supposed to be on the text that is the call-to-action.

components/login/SignInKey.tsx
craigzour: Definitely and this is why I tagged @connorscarolyns :) I think it is important to not tell the user that the credentials are wrong in such case because then they could be trying to fix a problem that does not exist.

craigzour (approved):

falila (approved):

PR #903, opened 2022-08-09
temporary fix: Do not check environment variables in NextAuth at build time

falila:
# Summary | Résumé
issue [here](https://github.com/cds-snc/platform-forms-client/actions/workflows/staging-build-push-container.yml)
issue thread [here](https://gcdigital.slack.com/archives/C01GFD150MC/p1660059542548469)


jeberhardt (approved):
Let's make sure we create an issue to fix this later on, so we don't forget about it!  ;)

PR #915, opened 2022-08-15
Feature/lock access after failed login attempts

### LINKED ISSUE #872: Lockout / Failed Attempts
The system will track failed login attempts of Stage-1 of the Login flow by tracking the number of attempts by email address and storing the value in a cache.  The failed attempts logging and blocking logic will apply to the `{base_url}/api/token/temporary` API and is not specific to NextJS or the Login UI.

 

The existing cache configured in the solution can be leveraged to store a new key based on the email address that increments on every failed login.  Once 10 failed logins are reached the value is no longer incremented however an expiration is set on the key so that it is automatically removed after 1 hour.  The lockout event will also generate an entry in the access_logs table.  If the number of failed login attempts has not reached 10 and a successful login in registered the the key based on the email address in the cache will be deleted.  This ensures that it requires 10 consecutive failed attempts before an account is locked out.

## Cache Key
Format: `auth:failed:${email}`
Value: string (Redis does not have defined integer type. The string stored at the key is interpreted as a base-10 64 bit signed integer to execute the operation.)

## Acceptance Criteria
- [x] 10 consecutive failed attempts to create a temporary token lockout further attempts with that email address for 1 hour.
- [x] A successful temporary token creation removes any failed key in the cache associated to that email address.
- [x] A failed attempt to create a temporary token increments the existing cache key or creates a new one.
- [x] A lockout creation event modifies the cache key to expire after 1 hour.
- [x] A lockout event does not continue processing to check the bearer/email credentials.

craigzour:
# Summary | Résumé

closes #872 

This PR adds a new login lockout mechanism. It is directly plugged on the existing temporary token API.

- Try to get a temporary token using the existing `/api/token/temporary` API but providing either a wrong email or form access token. You should see a new HTTP 401 response with information such as `remainingNumberOfAttemptsBeforeLockout` and `numberOfSecondsBeforeLockoutExpires` (depending on lockout state).

- This PR does not include any UI changes. The updated API will have to be plugged on the login page.
- A new Github issue has been created to plug the updated API to the Login UI https://github.com/cds-snc/platform-forms-client/issues/924
- A new Github issue has been created to add an alarm in Cloudwatch based on the error log message being published when someone accesses the temporary token API with an email address that has no access to any form https://github.com/cds-snc/platform-forms-client/issues/918

# Pull Request Checklist


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Make GC Branding in Footer configurable [#847](https://github.com/cds-snc/platform-forms-client/pull/847)
 - Logout Page [#847] (https://github.com/cds-snc/platform-forms-client/issues/870)
+- New login lockout mechanism plugged on existing temporary token API [#872](https://github.com/cds-snc/platform-forms-client/issues/872)
```
bryan-robitaille, line 27: Just an oversight but this should be moved up to the 'unreleased' section

pages/api/token/temporary.ts
```diff
@@ -18,37 +31,41 @@ import { MiddlewareProps } from "@lib/types";
...
-    const email = req.body["email"];
-    if (!email) {
-      res.status(400).json({ error: "Invalid payload" });
-    }
+const handler = async (
+  req: NextApiRequest,
+  res: NextApiResponse,
+  { formID, email }: MiddlewareProps
+): Promise<void> => {
+  assert(email); // Middleware function `requestContainsEmail` ensures that email is defined
```
bryan-robitaille, line 39: There are good points on both sides of the argument on whether `assert` should be used in production code.  I noticed however that these `asserts` replaced previous code that verified prop existance and either short circuited, by returning early, and returned the appropriate HTTP response code depending on the Error.
The way the current `assert`s are implemented they are outside of the `try` block and are unhandld in the current function.  There is also no custom message in the `assert`s so when the Assertion Error is thrown it may simply show `Assertion Error: undefined == true`

pages/api/token/temporary.ts
```diff
@@ -18,37 +31,41 @@ import { MiddlewareProps } from "@lib/types";
...
+    return res.status(200).json({ message: "success" });
   } catch (error) {
     logMessage.error("Failed to generate temporary token.");
     logMessage.error(error);
 
-    let errorMessage = "Malformed API Request";
-
-    if ((error as Error).message.includes("Could not send temporary token by email")) {
-      errorMessage = "GC Notify service failed to send temporary token";
+    if ((error as Error).message.includes("User is not found or not active")) {
```
bryan-robitaille, line 56: We may want to move this check up in the logic sequence so that if the error is only `User is not found or not active` it doesn't create an application error log.

pages/api/token/temporary.ts
```diff
@@ -143,4 +160,114 @@ const sendTemporaryTokenByEmail = async (email: string, temporaryToken: string)
     });
 };
 
-export default middleware([cors({ allowedMethods: ["POST"] }), validBearerToken()], handler);
+const requestContainsEmail = (): MiddlewareRequest => {
```
bryan-robitaille, line 163: Did this really needed to be seperated out into it's own middleware function?
It was previously in the scope of the existing function with:
```js
    const email = req.body["email"];
    if (!email) {
      return res.status(400).json({ error: "Invalid payload" });
    }
```

pages/api/token/temporary.ts
```diff
@@ -143,4 +160,114 @@ const sendTemporaryTokenByEmail = async (email: string, temporaryToken: string)
...
+    if (!email) {
+      res.status(400).json({ error: "Invalid payload" });
+      return { next: false };
+    }
+
+    return { next: true, props: { email } };
+  };
+};
+
+const canLogIn = (): MiddlewareRequest => {
```
bryan-robitaille, line 176: This middleware function should probably be moved to it's own middleware file.  We can call it something like 'lockedOut.ts`?

pages/api/token/temporary.ts
```diff
@@ -143,4 +160,114 @@ const sendTemporaryTokenByEmail = async (email: string, temporaryToken: string)
...
+  };
+};
+
+/**
+ * This is a middleware function that will validate the bearer token in the authorization header
+ *
+ * @param handler - the function to be executed next in the API call
+ * @returns either the handler to be executed next in the API call, or updates the res status and returns void
+ */
+export const hasValidFormAccessToken = (): MiddlewareRequest => {
```
bryan-robitaille, line 207: As with the other middleware comments this should probably be moved back to `validBearerToken` middleware.

pages/api/token/temporary.ts
```diff
@@ -143,4 +160,114 @@ const sendTemporaryTokenByEmail = async (email: string, temporaryToken: string)
...
+
+/**
+ * API outcomes:
+ * 200: `success` if everything goes well
+ * 400: `Invalid payload` if no email in request payload
+ * 401: `User is not allowed to log in` if email and/or form access token is not valid. Response paylaod also optionally contains `remainingNumberOfAttemptsBeforeLockout` (if user can still try to log in) and `numberOfSecondsBeforeLockoutExpires` (if user is locked out)
+ * 500: `Internal error` if something goes wrong while sending temporary token ; or ; `Server cannot respond at this time` if TOKEN_SECRET environment variable is not defined
+ */
+
+export default middleware(
```
bryan-robitaille, line 265: By consolidating some of the functionality above we could shorten the middleware list to something like:
```js
export default middleware(
  [
    cors({ allowedMethods: ["POST"] }),
    isUserLockedOut(),
    validBearerToken(),
  ],
  handler
);
```

bryan-robitaille (changes requested):
The logic of the feature seems really soundly implemented.  I just think we need to move some of the functionality around in the code base to align with how the project is currently structured.  The reasoning behind that is when we do decide to refactor and leverage NextJS's new middleware platform it would be easier knowing where all the middleware functions reside.
I understand why using `assert` in Typescript can seem like a shortcut to testing is a property exists or meets a condition.  The issue is that when using `assert` we need to ensure proper error handling.  The Node `assert` function is nice but TypeScript often has issues when using it for more  'type-y' assertions.  https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

lib/loginLockout.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    if (formUser) {
+      await prisma.accessLog.create({
+        data: {
+          action: LoggingAction.LOCKED,
+          userId: formUser.id,
+        },
+      });
+    } else {
+      logMessage.error("An email address with no access to any form has been locked out");
```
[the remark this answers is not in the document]
craigzour: Correct I will change to if `warn` level

pages/api/token/temporary.ts
craigzour: The previous code has just been moved in different places which are then ran in a specific order that insures the variables will be defined. Adding custom messages is definitely a good idea though. I initially wanted to test this approach as the beginning of each function was just a bunch of checks with different exit paths and it was kind of difficult to read through and understand what was happening.
Let me know if you still prefer the old solution?

craigzour: Also the functions in which I used the asserts are local to the file so the impact is limited if we were to change the logical order in
````
export default middleware(
  [
    cors({ allowedMethods: ["POST"] }),
    requestContainsEmail(),
    canLogIn(),
    hasValidFormAccessToken(),
  ],
  handler
);
````

File: pages/api/token/temporary.ts (hunk 4 above)
Comment (craigzour) [reply pages/api/token/temporary.ts:163]: The reason I did that is because the next middleware function I created (`canLogIn`) needs an email address to run its check. The `handler` function is the last piece of code being executed in the temporary token API logic.

File: __tests__/api/token/temporary.test.ts [hunk 9]
```diff
@@ -86,10 +94,10 @@ describe("TemporaryBearerToken tests", () => {
     });
 
     await temporary(req, res);
-    expect(res.statusCode).toEqual(403);
+    expect(res.statusCode).toEqual(400);
   });
 
-  it("throws error with invalid bearer token", async () => {
+  it("throws error with invalid form access token", async () => {
```
jeberhardt, line 100: "form access" and "bearer" tokens are the same thing. The term "form access" was introduced from a UI perspective, with the assumption that the average user wouldn't be familiar with the term "bearer token".

No need to change it back, just a comment.

pages/api/token/temporary.ts
craigzour: I guess it will depend on our little discussion in your first comment on this PR but I feel like middleware functions being used in one specific place the code should not be placed outside of that place.

craigzour:
@bryan-robitaille Following our team discussion the PR has been updated accordingly.

pages/api/token/temporary.ts
```diff
@@ -18,39 +25,91 @@ import { MiddlewareProps } from "@lib/types";
...
-    if (formID == undefined) {
-      return res.status(403).json({ error: "Invalid form request." });
-    }
-    if (!process.env.TOKEN_SECRET) {
-      return res.status(500).json({ error: "Server cannot respond at this time" });
+    const lockoutResponse = await isUserLockedOut(email);
+
+    if (lockoutResponse.isLockedOut) {
+      return res
+        .status(401)
```
bryan-robitaille, line 44: Nice use of 401 instead of 403.  Could catch.

pages/api/token/temporary.ts
```diff
@@ -18,39 +25,91 @@ import { MiddlewareProps } from "@lib/types";
...
+            lockoutResponse.remainingNumberOfAttemptsBeforeLockout,
+            lockoutResponse.numberOfSecondsBeforeLockoutExpires
+          )
+        );
     }
 
-    const email = req.body["email"];
-    if (!email) {
-      res.status(400).json({ error: "Invalid payload" });
+    const formID = await hasValidFormAccessToken(req, process.env.TOKEN_SECRET);
```
bryan-robitaille, line 53: The error "Form Access Token is invalid" can be thrown in two different places which could cause errors in the future.  For example if one part of the code is modified and the Error message changes but the other part of the code does not.  I'll refactor quickly to demonstrate how this can be overcome.

bryan-robitaille (approved):
Looks good and works well!

pages/api/token/temporary.ts
bryan-robitaille: https://github.com/cds-snc/platform-forms-client/pull/915/commits/dcf9645e049ce6e18492478d1935a50180ef0e1b

craigzour: Looks good! I did hesitate between the two solution at some point.

PR #951, opened 2022-08-24
feat: retrieve user lastLogin from DB

### LINKED ISSUE #946: Fetch last user login - Acceptable Use
The highlighted area should display the users last login in the correct browser timezone.

**Existing**
- The date/time showed in the current time of the immediate login.

**Fix**
- Fetch the last logon time from the AccessLogs table for the user

jeberhardt:
# Summary | Résumé

Closes #946 

- Login to `/auth/login`.
- Press the "cancel" button on the "Acceptable Use" page.
- Wait a few minutes (5 or more, for example)
- Login again on `/auth/login`
- The timestamp at the top of the acceptable use page should be time from your first login, a few moments ago

There is another PR to format the timestamp correctly.

# Pull Request Checklist

### DIFF
--- a/lib/users.ts
+++ b/lib/users.ts
@@ -106,6 +106,11 @@ export const adminRole = async (isAdmin: boolean, userId: string): Promise<[bool
   }
 };
 
+/**
+ * Retrieves the accessLog entry for the last login for a user
+ * @param userId
+ * @returns AccessLog object
+ */
 export const userLastLogin = async (userId: string): Promise<AccessLog | null> => {
   try {
     return await prisma.accessLog.findFirst({
--- a/lib/users.ts
+++ b/lib/users.ts
@@ -1,8 +1,9 @@
 import { User } from "next-auth";
 import { prisma, prismaErrors } from "@lib/integration/prismaConnector";
-import { FormUser, Prisma, UserRole } from "@prisma/client";
+import { AccessLog, FormUser, Prisma, UserRole } from "@prisma/client";
 import { JWT } from "next-auth";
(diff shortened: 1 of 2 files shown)


lib/users.ts
```diff
@@ -104,3 +105,24 @@ export const adminRole = async (isAdmin: boolean, userId: string): Promise<[bool
...
+      where: {
+        userId: userId,
+        action: LoggingAction.LOGIN,
+      },
+      orderBy: {
+        timestamp: "desc",
+      },
+    });
+  } catch {
+    return null;
```
falila, line 126: Is there a reason not to use `prismaErrors(e,null)` ? if an error occurs it might be helpful for debugging this feature.

falila (changes requested):

lib/users.ts
craigzour: @falila is correct. That function should be use since it will `return null` in the end but it will also take care of logging any Prisma errors

craigzour (changes requested):

lib/users.ts
jeberhardt: Nice catch. I wasn't aware of that custom function.

craigzour:
@jeberhardt I think it would be good to have one or two tests on that new function you created

jeberhardt:
> @jeberhardt I think it would be good to have one or two tests on that new function you created

I don't think there's anything to actually test here. Everything would be mocked, so it's not much of a test. There is zero logic in the function. Get one row back from the DB, or return null. Nothing else.

craigzour (approved):

PR #997, opened 2022-09-01
Form builder text editor

timarney:
# Summary | Résumé

Adds a rich text editor for the text sections of a form.
Also includes a Markdown serializer for Slate.

| Form builder  | Form |
| ------------- | ------------- |
| image  | image  |


pcraig3 (changes requested):
Still some basic things we should try and fix here:

| Slate editor | Form preview |
|--------|-------|
|   Screen Shot 2022-09-09 at 15 21 33    |   Screen Shot 2022-09-09 at 15 22 00   |

There’s a few things in there:

- newline isn’t respected
- first bullet isn’t in a list
- ordered list vs bullet lists aren’t separated

Also, there are a handful of console errors that show up, mostly around unexpected PropTypes. Should be relatively easy to fix them.

pcraig3 (approved):
👍  Let not perfect be the enemy of the good, as they say.

PR #999, opened 2022-09-02
Reset data store between tests

dsamojlenko:
# Summary | Résumé

- Reset the data store between test runs
- Add initialize method to store to enable above
- Add getSchema method to store for easier retrieval


pcraig3 (approved):
Walked through this together. All good on my end.

PR #1003, opened 2022-09-07
Remove getTemplate hook and retrieve serialized schema directly from store

dsamojlenko:
# Summary | Résumé

Small refactor:
- Remove getTemplate hook
- Return serialized schema from getSchema in templates store
- Retrieve serialized schema from template store in CopyToClipboard and Output
- Renamed utility method getSchema to getSchemaFromState for clarity


components/form-builder/store/useTemplateStore.ts
```diff
@@ -92,7 +92,8 @@ const useTemplateStore = create<ElementStore>()(
         state.form.elements[index].properties.choices = choices;
       });
     },
-    getSchema: () => getSchema(get()),
+    getSchema: () => getSchemaFromState(get()),
+    getSerializedSchema: () => JSON.stringify(getSchemaFromState(get()), null, 2),
```
dsamojlenko, line 96: A case could probably be made to combine these into one. Currently we don't call getSchema directly from anywhere.

pcraig3 (approved):
Looks great!

PR #1010, opened 2022-09-08
Updated our Question Panel to be a little bit more like the Figma mocks

pcraig3:
# Summary | Résumé

This pull request adds styling (including a whole lot of borders) to our questions panel and rearranges the actions in the "ActionsPanel". 

No functional changes, rather the goal is to make the design a bit cleaner.

| before | after |
|--------|-------|
|  grey outlines, no numbered questions, "add button" on the left       |    outlines around everything, rounded borders, re-organized actions panel, numbered questions   |
|   Screen Shot 2022-09-08 at 10 38 18   |  Screen Shot 2022-09-08 at 10 38 27   |

Here's what the mocks we have look like, for reference. The question fields here don't have borders, and the "Require" toggle is not a checkbox largely it's the same.

Screen Shot 2022-09-08 at 10 32 59

There are still more design updates to do here but this gets us closer to what we want.


components/form-builder/panel/ElementPanel.tsx
```diff
@@ -169,6 +179,27 @@ const Form = ({ item }: { item: ElementTypeWithIndex }) => {
   return (
     <>
       <Row>
+        <div>
+          <QuestionNumber>{item.index + 1}</QuestionNumber>
```
dsamojlenko, line 183: Should this use `item.id` instead?

dsamojlenko (approved):
Left a question for future consideration

components/form-builder/panel/ElementPanel.tsx
pcraig3: I think the number should be the consistent index, not the "id" of the question. 
I agree that it's a bit disorienting, but the numbers should represent the order in which someone filling out the form will actually see it.

PR #1023, opened 2022-09-12
feat: Form Builder preview

dsamojlenko:
# Summary | Résumé

Adds the ability to preview a form while building it. 

The previewed form can be filled out and submitted. The validation routines will fire if configured, and validation errors displayed. 

When a completed and valid form is submitted, the Confirmation page (endPage) is displayed, although the form submission does not actually go anywhere.

Also moves the hidden JSON output to a "Tab" in the navigation above the Form Builder. This will eventually be the JSON download/upload functionality.

The UI for this is temporary, as we are working towards moving the Form Builder out of the Admin section.

Screen Recording 2022-09-15 at 11 30 20 AM

## Unresolved questions / Out of scope
Note: the `components/form-builder/preview/Form.tsx` is a fork of `components/forms/Form/Form.tsx`. Ideally, we'll merge our copy back upstream, but keeping separate for now.


pcraig3 (approved):
I think it looks great! Awesome work on this 👍

PR #1039, opened 2022-09-14
Update form builder so it looks more like the Figma mockups

pcraig3:
# Summary | Résumé

This PR is mostly stylistic rather than changing logic. It is mainly concerned with adding elements and changing styling to bring us more in line with [the Figma mocks we've been using](https://www.figma.com/file/dbezuEglWNQ8qyO64eJJiA/v1.-Form-builder?node-id=353%3A9861).

The list of changes, more or less:

- Add the GDS/CDS checkboxes & radios
- Add a "Required" field to the Modal
- Always add one option by default to new question fields
- Keep text input styles in an `Input` file
  - Update the "Question" field to just have an underline and use bold text 
- Focus the Question automatically, not the Option
  - When you press "Enter" in an Option, you get a new Option 
- Add CDS design system-type styling for buttons
- Add CDS design system-type focus styling for inputs and the select that lets you change questions
- Make all the text smaller for everything except the main question title

### Screenshots

Lots of small changes but hopefully it looks a bit more like the Figma mocks and less clutter in general (eg, making the text smaller helps this).

|  Screen Shot 2022-09-15 at 10 19 35 | Screen Shot 2022-09-15 at 10 19 33 |

### gif

⚠️ Note: this gif is a little out of date (doesn't have the new style for the Question Title), but everything else still is the right thing.

You can't really see the focus/hover styles in the screenshot, so here's a gif.

| Screen Recording 2022-09-14 at 18 53 35  |  Screen Recording 2022-09-14 at 18 52 07 |

Adding more fields to the modal still needs to happen, and there's also a few more field types that I am planning to add (eg, email, phone number, date)

### DIFF
--- a/components/form-builder/__tests__/useTemplateStore.test.ts
+++ b/components/form-builder/__tests__/useTemplateStore.test.ts
@@ -54,7 +54,9 @@ describe("TemplateStore", () => {
     expect(result.current.form.titleEn).toBe("My Form");
     expect(result.current.form.elements).toHaveLength(1);
     expect(result.current.form.elements[0].properties.titleEn).toBe("");
-    expect(result.current.form.elements[0].properties.choices).toHaveLength(0);
+    // By default, there is one choice available
+    expect(result.current.form.elements[0].properties.choices).toHaveLength(1);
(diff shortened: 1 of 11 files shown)


dsamojlenko (approved):
Looks great!

PR #1042, opened 2022-09-15
feat: disable first move up and last move down buttons

dsamojlenko:
# Summary | Résumé

Disables the Move up button on the first element, and disables the Move down button on the last element.

| image | image |

Add multiple items, notice the move up/down buttons

# Pull Request Checklist

### DIFF
--- a/components/form-builder/panel/PanelActions.tsx
+++ b/components/form-builder/panel/PanelActions.tsx
@@ -95,21 +95,23 @@ export const PanelActions = ({
 }) => {
   const { t } = useTranslation("form-builder");
   const { remove, moveUp, moveDown, add, duplicateElement, form } = useTemplateStore();
+  const isLastItem = item.index === form.elements.length - 1;
+  const isFirstItem = item.index === 0;
 
   return (
     <Actions className="panel-actions">
       <UpDown>
         <PanelButton
-          className={item.index == 0 ? "disabled" : ""}
+          className={isFirstItem ? "disabled" : ""}
           icon={<ChevronUp />}
-          onClick={() => moveUp(item.index)}
+          onClick={isFirstItem ? undefined : () => moveUp(item.index)}
         >
           <Label>{t("Move up")}</Label>
         </PanelButton>
         <PanelButton
-          className={item.index + 1 == form.elements.length ? "disabled" : ""}
+          className={isLastItem ? "disabled" : ""}
           icon={<ChevronDown />}
-          onClick={() => moveDown(item.index)}
+          onClick={isLastItem ? undefined : () => moveDown(item.index)}
         >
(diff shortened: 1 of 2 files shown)


components/form-builder/panel/PanelActions.tsx
```diff
@@ -24,6 +24,26 @@ const PanelButton = styled(Button)`
...
+    &:focus,
+    &:active {
+      color: #ccc;
+      background-color: #eee;
+
+      svg {
+        fill: #ccc;
+      }
+    }
+  }
```
pcraig3, line 46: There is a very subtle background colour here that I don't think we want. Also the focus ring, we should turf it.

It's a `box-shadow: none` to get rid of the blue focus ring.

Also, it's not going to pass a11y (which, you could argue, is not a huge problem if you make the element non-focusable).

Screen Recording 2022-09-15 at 16 50 23

components/form-builder/panel/PanelActions.tsx
```diff
@@ -74,15 +94,25 @@ export const PanelActions = ({
...
 
   return (
     <Actions className="panel-actions">
       <UpDown>
-        <PanelButton icon={<ChevronUp />} onClick={() => moveUp(item.index)}>
+        <PanelButton
+          className={isFirstItem ? "disabled" : ""}
+          icon={<ChevronUp />}
+          onClick={isFirstItem ? undefined : () => moveUp(item.index)}
+        >
```
pcraig3, line 108: Better to add [the `disabled` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-disabled) rather than a class. Then you can just keep the `onClick` except it shouldn't work.

And your css can be like `button[disabled] { ... }`

pcraig3 (changes requested):
I've left some comments. Basically, the button should be completely non-reactive and we should use the `disabled` attribute.

components/form-builder/panel/PanelActions.tsx
```diff
@@ -74,15 +94,25 @@ export const PanelActions = ({
   children: React.ReactNode;
 }) => {
   const { t } = useTranslation("form-builder");
-  const { remove, moveUp, moveDown, add, duplicateElement } = useTemplateStore();
+  const { remove, moveUp, moveDown, add, duplicateElement, form } = useTemplateStore();
```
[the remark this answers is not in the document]
dsamojlenko: 👍

pcraig3 (dismissed):
yo let's move this "in" to the repo!!! 🤙

pcraig3 (commented):
yo let's (not optionally) move this "in" to the repo!!! 🤙

pcraig3 (approved):

PR #1058, opened 2022-09-21
Add 3 new fields: email, phone, and date

pcraig3:
# Summary | Résumé

This PR adds 3 new form fields: `email`, `phone`, and `date`. All of them are a subtype of the `textField` type, but they need new icons, "sample text" and they generate different schema markup.

| Screen Shot 2022-09-21 at 13 43 24 | Screen Shot 2022-09-21 at 13 43 07 |

Here's a closer look at what each one looks like.

| field | img |
|-------|-----|
|   Phone field: `555-555-0000`    | Screen Shot 2022-09-21 at 13 44 54 |
|  Email field: `example@canada.gc.ca`     |  Screen Shot 2022-09-21 at 13 45 07  |
|   Date field: `mm/dd/yyyy`   |  Screen Shot 2022-09-21 at 13 45 28  |

## Notes

In the JSON schema, these 3 new fields are all basically "textInput" fields with different validation rules ([see them here](https://github.com/cds-snc/platform-forms-client/blob/develop/lib/middleware/schemas/templates.schema.json#L299-L302)).  Previously, changing between fields meant changing the type:

- `form.elements[$index].type`

With these fields, the `type` stays as `inputField`, but instead of updating the element's type, we have to update the validation type:

- `form.elements[$index].properties.validation.type` 

This means that when we change away from `inputField` (to `radio`, for example), we have to reset the validation type. I added "[lodash.unset](https://www.npmjs.com/package/lodash.unset)" to the package.json since it's essentially the inverse of "lodash.set".

I wasn't able to directly test switching between question types, but I wrote unit tests that set and unset the validation types.


components/form-builder/panel/ElementPanel.tsx
```diff
@@ -76,11 +93,12 @@ const getSelectedOption = (item: ElementTypeWithIndex): ElementOption => {
...
     form: { elements },
   } = useTemplateStore();
-
-  const { type } = elements[item.index];
+  let { type } = elements[item.index];
 
   if (!type) {
     return elementOptions[2];
+  } else if (type === "textField") {
+    type = elements[item.index].properties.validation.type || type;
```
timarney, line 101: Maybe add a comment to help explain the "why" here

timarney (approved):
Works great --- just the one comment to maybe add a comment.

components/form-builder/panel/ElementPanel.tsx
pcraig3: Sounds good

timarney (approved):

PR #1059, opened 2022-09-21
feat: Update RichText editor to use Plate components

dsamojlenko:
# Summary | Résumé

Replaces the custom Slate editor that we were using with [Plate](https://plate.udecode.io/), a Slate implementation.

| image | image |

- Deserializing single line breaks (not paragraphs) (`/n` appears to be stripped)
- Multiple editors and targeting (see: links) (note: removed link button for now)
- Nested numbered lists not rendering right
- Had to comment out `Layout.test.ts` as a dependency of Plate is causing issues with Jest
- ~~Outstanding type error in `plate-editor/helpers/markdown.ts`~~


components/form-builder/elements/RichTextLocked.tsx
```diff
@@ -18,48 +19,55 @@ const ElementWrapperDiv = styled.div`
...
+  initialValue: string;
   schemaProperty: string;
 }) => {
   const input = useRef<HTMLInputElement>(null);
   const { localizeField, updateField } = useTemplateStore();
-  const [value, setValue] = useState<Descendant[]>(initialValue);
+  const editorId = `${id}-editor`;
+  const editor = useMyPlateEditorRef()!;
+
+  // const [value, setValue] = useState<Descendant[]>(initialValue);
```
timarney, line 53: Let's remove this

components/form-builder/plate-editor/helpers/markdown.ts
```diff
@@ -1,57 +1,38 @@
...
-        listItem: ELEMENT_LI,
-        heading: {
-          1: ELEMENT_H1,
-          2: ELEMENT_H2,
-          4: ELEMENT_H4,
-          3: ELEMENT_H3,
-          5: ELEMENT_H5,
-          6: ELEMENT_H6,
-        },
+export const serializeMd = (value: (BlockType | LeafType)[]) => {
```
timarney, line 4: @dsamojlenko 

Think I tracked down the TS errors on this.

<hr>

I console logged out the value coming into `serializeMd `

From what I can see it's an array with a single element so value[0] which contains either block nodes or leaf nodes

```
 export const serializeMd = (value: (BlockType | LeafType)[]) => {
```

The code you had 

```
return serialize(
    { children: value },
```

Was kind of forcing it to look like a block as Leafs don't have children... at which point it was complaining that the "block" was missing a type prop.

Value[0]  (if not undefined seems to fixes all that) --- seems to work from my testing

------

After that TS it started complaining that a "p" was not assignable to paragraph for the NodeType opts

TypeScript autocompleted those as just strings so I ended up replacing the node types with the string equivalents

timarney: ... further updates for the onChange now use the Value type  ->TDescendant = TElement | TText;

Which pretty my maps back to BlockType | LeafType

dsamojlenko: Found this issue in the remark-slate repo: https://github.com/hanford/remark-slate/issues/57

...that explains the issue with the nodeTypes. I updated the serialize script to use the custom nodeTypes required by Plate: https://github.com/cds-snc/platform-forms-client/pull/1059/commits/c00f56377d1544ffb926391170d79f37af770f28

pcraig3 (approved):
Let not perfect be the enemy of the good.

Let's merge this.

PR #1078, opened 2022-09-28
Save page & Start page

pcraig3:
# Summary | Résumé

This PR adds 2 new pages to our form builder: 
1. A "start" page with a "design a form" link and an "upload" button. ([On Figma](https://www.figma.com/file/dbezuEglWNQ8qyO64eJJiA/v1.-Form-builder?node-id=2%3A4))
2. A "save" page with a download button and a copy button. ([On Figma](https://www.figma.com/file/dbezuEglWNQ8qyO64eJJiA/v1.-Form-builder?node-id=698%3A10680))

## Notes

- Start page is the new page that loads by default
- I've bolded the nav links so that you can see what page you're on 
- Added translation keys for everything that I could see, so we should be good on that one

### Weird stuff

1. The nested code that parses JSON in our Uploader doesn't respect the `try/catch`. (eg, if you upload a non JSON file it will blow up.) Is there a way to solve this that isn't just adding a new `try/catch`. (I feel this looks really ugly.)

2. Since we are being helpful and focusing the next question field when you add a new question: the focus always shoots you down the page when you upload a form. Is there a way to keep this behaviour just when you do "Add element" and not when you are navigating between tabs? (We are doing something similar for the Options, so we would want to adjust that as well.)

### Start page

| screenshot | 
|-------|
|  Screen Shot 2022-09-29 at 11 49 50   |

### Save page

| screenshot | screenshot (view code open) |
|-------|-----|
|  Screen Shot 2022-09-28 at 17 00 17   | Screen Shot 2022-09-28 at 17 00 27   |

## gif

Screen Recording 2022-09-29 at 11 50 07


public/static/locales/en/form-builder.json
```diff
@@ -18,6 +18,19 @@
...
   "json": "JSON",
-  "preview": "preview",
-  "placeHolderFormTitle": "Form Title"
+  "preview": "Preview",
+  "placeHolderFormTitle": "Form Title",
+  "share": "Share",
+  "save": "Save",
+  "saveH1": "Save your progress",
+  "saveP1": "We don’t save your form on our website, but don’t worry, you can save the file to your computer.",
+  "saveP2": "Save the from file to save your progress, or if you’re done, save it and open it again later to pick-up where you left off.",
```
timarney, line 27: > Save the from file to save your progress

Save the form file ya?

pcraig3: Yes, good point. I rewrote this sentence to be a bit simpler.

timarney (approved):

PR #1082, opened 2022-09-29
feat: Double-click headers to switch language

pcraig3:
# Summary | Résumé

We need to get the French/English language toggle out of the top nav but we also want it to be available to us.

The compromise here is using the Headings as the toggle, even though it is not something we want to do long-term.

Still easy to click and very unlikely someone will activate it by accident.

## gif 

Screen Recording 2022-09-29 at 16 05 55


timarney (approved):
Works great for now -- until we get that part into the proper interface

PR #1091, opened 2022-10-03
Change the ordering of elements in the element selector

pcraig3:
# Summary

One of the things we haven't changed since the beginning was the way the elements are listed in the question type selector, we've just been adding new ones to the bottom. To better align with the mockups, we're going to order them this way instead.

```
  Page text
  —
  Single choice
  Multiple choice
  Dropdown list
  —
  Short answer
  Long answer
  —
  Phone number
  Email
  Date
  Numeric field
```

Also, I renamed the "prepend" property to "append" to more accurately reflect what it's doing.

# Screenshot 

| mockup | before | after |
|--------|--------|-------|
|  image (1)   |    Screen Shot 2022-10-03 at 15 04 29  |  Screen Shot 2022-10-03 at 15 04 18   |


dsamojlenko (approved):
Looks good!

PR #1130, opened 2022-10-11
Feat/signup validation

thiessenp-cds:
# Summary | Résumé

This work adds two new fields to the signup form, First Name and Last Name. Validation work was also done for all signup fields. "Yup" was chosen as the validation lib because of its ease of use (intuitive, well tested, works well with Formik e.g. error object shared). 

Note: the First Name and Last Name fields still need to be added to the Server/DB (separate task created)


components/forms/TextInput/TextInput.tsx
```diff
@@ -6,7 +6,7 @@ import { InputFieldProps, CharacterCountMessages, HTMLTextInputTypeAttribute } f
 
 export interface TextInputProps extends InputFieldProps {
   type: HTMLTextInputTypeAttribute;
-  characterCountMessages: CharacterCountMessages;
+  characterCountMessages?: CharacterCountMessages;
```
thiessenp-cds, line 9: I aimed for the most minimal change -vs- a big refactor or a new component

lib/validation.tsx
```diff
@@ -310,3 +310,72 @@ export const isValidGovEmail = (email: string, domains: string[]): boolean => {
...
+
+/**
+ * This function checks if a given password is a valid email based on:
+ * 1. The policy found here found here: https://github.com/cds-snc/forms-terraform/blob/develop/aws/cognito/user_pool.tf#L5-L11
+ * 2. Additionally a max length of 50 characters
+ * And it returns true if the password is a valid otherwise false.
+ * @param password A valid password
+ * @returns {boolean} The validation result
+ */
+export const isValidPassword = (password: string): boolean => {
```
thiessenp-cds, line 322: This function is not actually used, the below ones are. I left it in encase it would be useful in a future but this is kind of a code smell. I should probably remove it - what do you think?

pages/signup/register.tsx
```diff
@@ -1,60 +1,139 @@
...
+      .required(t("input-validation.required", { ns: "common" }))
+      .email(t("input-validation.email", { ns: "common" }))
+      .test(
+        "username-govEmail",
+        t("signUpRegistration.fields.username.error.validGovEmail"),
+        (value = "") => isValidGovEmail(value, emailDomainList.domains)
+      ),
+    firstName: Yup.string().required(t("input-validation.required", { ns: "common" })),
+    lastName: Yup.string().required(t("input-validation.required", { ns: "common" })),
+    password: Yup.string()
```
thiessenp-cds, line 29: I went with the sort of waterfall/filter approach with Yup of displaying one error at a time -vs- many at once with new lines or something like that

components/forms/TextInput/TextInput.tsx
Moro-Code: This works well for now thanks @thiessenp-cds

lib/validation.tsx
Moro-Code: Would remove it if it's not used. If you wanted to keep the regex as a reference you could add it as a comment before the functions below

pages/signup/register.tsx
Moro-Code: I think thats okay for now. This will probably need changing as it might be super annoying for a user registering to keep failing on one condition at a time. However it fits the bill currently, we can probably work around this with a description for the password field containing the rules

pages/signup/register.tsx
```diff
@@ -1,60 +1,139 @@
...
+                id={"validationAlert"}
+                tabIndex={0}
+              >
+                <ol className="gc-ordered-list">
+                  {errors &&
+                    Object.entries(errors).map(([elementId /*, errorMessage*/]) => {
+                      return (
+                        <ErrorListItem
+                          errorKey={elementId}
+                          value={t("input-validation.required", { ns: "common" })}
```
Moro-Code, line 92: Is there any reason we're not passing in the errorMessage here ?

Moro-Code (changes requested):
Great work Peter ! I had some minor comments if you could take some time to look at it, I would greatly appreciate it.

pages/signup/register.tsx
```diff
@@ -1,60 +1,139 @@
 import React, { useState } from "react";
 import { Formik } from "formik";
-import { Button, TextInput, Label } from "@components/forms";
+import { Button, TextInput, Label, Alert, ErrorListItem } from "@components/forms";
 import { useAuth } from "@lib/hooks";
 import { useTranslation } from "next-i18next";
 import { GetServerSideProps } from "next";
 import { serverSideTranslations } from "next-i18next/serverSideTranslations";
 import { Confirmation } from "@components/auth/Confirmation/Confirmation";
+import * as Yup from "Yup";
```
Moro-Code, line 9: Sorry just caught this. This import seems wrong here ? 

I would expect the import to look like 

```JS
import Yup from "yup"
```

Moro-Code (changes requested):

pages/signup/register.tsx
thiessenp-cds: Completely missed that. Good point, will update

pages/signup/register.tsx
thiessenp-cds: Oh right, yup will change yup :)

Moro-Code (changes requested):
Great work think this is close to being merged in. Just had one comment !

Moro-Code (approved):
Looks good. Thanks for all the work Peter !

PR #1133, opened 2022-10-12
Feature/permission migration

### LINKED ISSUE #1099: Add Migration for base Priveleges
In order to ensure that the application can properly run after the permissions model upgrade we'll need to ensure that a bare minimum of privileges exist in the database.

When doing local development or for a brand new install the `/prisma/seed.ts` file is run and populates the database.  However during an upgrade path the `seed.ts` file is not run and the Privilege model in the database will be empty.

Goal:
- Ensure that after the Prisma migration is done the application has all the Privileges in the database it requires in order to function correctly.
 See [Permissions Ideation document](https://docs.google.com/document/d/1J6CWJrw6BQw0trY7hdY6XmkhugIRI_Vt7MTVIbhXTUc/edit?usp=sharing) for potential list.

Acceptance Criteria:
 - [x] After new install, or a `prisma migrate reset` the app has the necessary Privileges to function.
 - [x] After a prisma deploy, or migration of an existing app instance, the app has the necessary Privileges to function.
- [x] All users should have at least the `base` Privilege associated to their User instance.

bryan-robitaille:
# Summary | Résumé
Adds initial privileges that will be required for the application to function properly immediately after migration.

- Removes duplicate seeding from Prisma seed file.
- Moves seeding to Prisma migration (.sql) file.

Closes #1099

### DIFF
--- a/prisma/migrations/20221006181804_permissions/migration.sql
+++ b/prisma/migrations/20221006181804_permissions/migration.sql
@@ -45,3 +45,40 @@ ALTER TABLE "_PrivilegeToUser" ADD CONSTRAINT "_PrivilegeToUser_A_fkey" FOREIGN 
 
 -- AddForeignKey
 ALTER TABLE "_PrivilegeToUser" ADD CONSTRAINT "_PrivilegeToUser_B_fkey" FOREIGN KEY ("B") REFERENCES "User"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- Populate with Privileges
+
+INSERT INTO
+  "Privilege" ("id","nameEn", "nameFr", "descriptionEn", "descriptionFr", "permissions")
+VALUES
+  (gen_random_uuid(),'Base','Base','Base Permissions','Autorisations de Base','[
+    {"action":"create","subject":"FormRecord"},
+    {"action":["view","update","delete"],"subject":"FormRecord","conditions":{"users":{"$elemMatch":{"id":"${user.id}"}}}},
+    {"action":"update","subject":"FormRecord","fields":["publishingStatus"],"inverted":true},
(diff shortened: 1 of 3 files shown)


prisma/migrations/20221006181804_permissions/migration.sql
```diff
@@ -45,3 +45,40 @@ ALTER TABLE "_PrivilegeToUser" ADD CONSTRAINT "_PrivilegeToUser_A_fkey" FOREIGN
...
+
+-- Populate with Privileges
+
+INSERT INTO
+  "Privilege" ("id","nameEn", "nameFr", "descriptionEn", "descriptionFr", "permissions")
+VALUES
+  (gen_random_uuid(),'Base','Base','Base Permissions','Autorisations de Base','[
+    {"action":"create","subject":"FormRecord"},
+    {"action":["view","update","delete"],"subject":"FormRecord","conditions":{"users":{"$elemMatch":{"id":"${user.id}"}}}},
+    {"action":"update","subject":"FormRecord","fields":["publishingStatus"],"inverted":true},
```
craigzour, line 57: Conditions will have to be adjusted as we implement and test final permissions. `publishingStatus` is part of `formConfig` in this case so we can use `formConfig.publishingStatus`.

craigzour (approved):
Outside of my comment it does look good to me :)

prisma/migrations/20221006181804_permissions/migration.sql
bryan-robitaille: 100% Looking to refine these as we go.  Good catch on the sub property.  I'll modify.

bryan-robitaille: We'll also want to figure out what the 'subject' really is.  For example, are we really going to be checking a FormRecord the most or a FormConfiguration type?

PR #1135, opened 2022-10-13
feat: add permission condition interpolation feature

craigzour:
# Summary | Résumé

- Added permission condition interpolation feature
- Added tests for new interpolation function

### DIFF
--- a/lib/policyBuilder.ts
+++ b/lib/policyBuilder.ts
@@ -1,6 +1,19 @@
-import { createMongoAbility, MongoAbility, RawRuleOf, InferSubjects } from "@casl/ability";
+import {
+  createMongoAbility,
+  MongoAbility,
+  RawRuleOf,
+  InferSubjects,
+  MongoQuery,
+} from "@casl/ability";
 import { FormRecord } from "@lib/types";
 import { User, Privilege as PrismaPrivilege } from "@prisma/client";
+import get from "lodash/get";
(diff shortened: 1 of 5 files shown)


lib/policyBuilder.ts
```diff
@@ -1,6 +1,14 @@
...
+import {
+  createMongoAbility,
+  MongoAbility,
+  RawRuleOf,
+  InferSubjects,
+  MongoQuery,
+} from "@casl/ability";
 import { FormRecord } from "@lib/types";
 import { User, Privilege as PrismaPrivilege } from "@prisma/client";
+import { AnyObject } from "immer/dist/internal";
```
bryan-robitaille, line 10: Instead of importing a type from a package that is a dependency elsewhere in the application, we could just clone the type directly in the file.

```js
type AnyObject = {
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  [key: string]: any;
};
```

bryan-robitaille (approved):
Just a single comment on how we can remove the import for {AnyObject} type.

lib/policyBuilder.ts
craigzour: Will change

bryan-robitaille:
> 

We can clean them up as part of our test modifications as we wrap the CRUD functions with permissions.

PR #1137, opened 2022-10-13
fix: default privileges definition clashes with the interpolation function

craigzour:
# Summary | Résumé

- Fix default permissions in the SQL migration script
- Added error log when interpolation fails
- Quick fix permission check (while waiting for final checks to be implemented)


pages/api/auth/[...nextauth].ts
```diff
@@ -97,21 +96,7 @@ export const authOptions: NextAuthOptions = {
       session.user.acceptableUse = token.acceptableUse;
       session.user.name = token.name ?? null;
       session.user.image = token.picture ?? null;
-
-      const privileges = await getPrivilegeRulesForUser(token.userId as string);
-      if (!privileges.length) throw new Error("User has no privileges");
```
Comment (craigzour) [inline pages/api/auth/[...nextauth].ts:102]: @bryan-robitaille removed it from here since the `getPrivilegeRulesForUser` already does that check (and also throws an error)

bryan-robitaille (approved):
Looking good!

PR #1151, opened 2022-10-18
Feature/protect crud functions with permissions

craigzour:
# Summary | Résumé

- Check permissions when accessing templates CRUD function
- Check permissions when accessing users CRUD function

### DIFF
--- a/lib/users.ts
+++ b/lib/users.ts
@@ -4,62 +4,15 @@ import { JWT } from "next-auth";
 import { LoggingAction } from "./auth";
 import { Ability } from "./policyBuilder";
 import { checkPrivileges } from "@lib/privileges";
-/**
- * Get all Users
- * @returns An array of all Users
- */
-export const getUsers = async (ability: Ability) => {
-  try {
-    checkPrivileges(ability, [{ action: "view", subject: "User" }]);
-    const users = await prisma.user.findMany({
-      select: {
-        id: true,
-        name: true,
-        email: true,
-        privileges: {
-          select: {
-            id: true,
-            nameEn: true,
-            nameFr: true,
-            descriptionEn: true,
-            descriptionFr: true,
-          },
-        },
-      },
-      orderBy: {
-        id: "asc",
-      },
-    });
-
-    return users;
-  } catch (e) {
-    return prismaErrors(e, []);
-  }
-};
-
-/**
- * Get an APIUser
- * @param userId ApiUser Id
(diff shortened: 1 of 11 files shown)


lib/users.ts
```diff
@@ -4,62 +4,15 @@ import { JWT } from "next-auth";
...
-};
 
 /**
  * Get or Create a user if a record does not exist
  * @returns A User Object
  */
-export const getOrCreateUser = async (userToken: JWT) => {
+export const getOrCreateUser = async (ability: Ability, userToken: JWT) => {
   try {
+    checkPrivileges(ability, [{ action: "view", subject: "User" }]);
```
bryan-robitaille, line 14: This function is only called internally from NextAuth.  We should remove the checkPrivileges for this one.

lib/users.ts
```diff
@@ -129,6 +82,59 @@ export const getOrCreateUser = async (userToken: JWT) => {
...
+    return prismaErrors(e, []);
+  }
+};
+
+/**
+ * Get a APIUser
+ * @param userId ApiUser Id
+ * @returns ApiUser Object
+ */
+export const getApiUser = async (ability: Ability, userId: string): Promise<ApiUser | null> => {
```
bryan-robitaille, line 124: This function isn't currently used in the codebase and could be removed.  There isn't any privileges that currently restrict access to the ApiUser model.  When needed in the codebase the ApiUser object is returned as part of the Template relationship if a user has the ability to view the template object (they are one of the owners of the Template)

pages/api/templates.ts
```diff
@@ -20,14 +19,31 @@ import {
...
+const templates = async (
+  req: NextApiRequest,
+  res: NextApiResponse,
+  { session }: MiddlewareProps
+) => {
   try {
-    const session = await isAdmin({ req, res });
-    const response = await templateCRUD({ user: session?.user, method: req.method, ...req.body });
+    if (!session) {
+      res.status(403);
```
bryan-robitaille, line 35: Let's swap this out to a `401` instead of a `403`.  
`401`  : Unauthorized (do not have a session
`403`  :  Forbidden (not allowed to perform action)

bryan-robitaille (changes requested):
Awesome work!  A few small things I picked up for a quick change and we should be good.

bryan-robitaille (approved):
Awesome!  Let's get this merged!

PR #1154, opened 2022-10-19
Permissions Feature ... One more time.... 🎧  Daft Punk

bryan-robitaille:
# Summary | Résumé
This is one of the last PR's that covers off the permissions feature.
Wrapped:
- Bearer token API and CRUD functions
- Flags aka Application Settings API and CRUD functions
- Auth lib

Fixed:
- Typescript error in CORS Middleware
- Removed Retrieval API from play by denying access to all methods (For now until Vault is a priority)
- Small refactors to remove UserRole from the code base
- Fixed tests that were failing because of a dependence on UserRole.Admin
- Fixed AcceptableUse API to require a valid session (was previously allowing unauthenticated calls)
- some other small refactors along the way.

### DIFF
--- a/__tests__/api/id/form/bearer.test.ts
+++ b/__tests__/api/id/form/bearer.test.ts
@@ -6,18 +6,17 @@
 import { createMocks, RequestMethod } from "node-mocks-http";
 import { unstable_getServerSession } from "next-auth/next";
 import retrieve from "@pages/api/id/[form]/bearer";
+import { Base, ManageForms, getUserPrivileges } from "__utils__/permissions";
 import jwt from "jsonwebtoken";
-import { logMessage } from "@lib/logger";
 import { logAdminActivity } from "@lib/adminLogs";
-import { prismaMock, checkLogs } from "@jestUtils";
-import { Prisma, UserRole } from "@prisma/client";
+import { prismaMock } from "@jestUtils";
+import { Prisma } from "@prisma/client";
 
 jest.mock("next-auth/next");
 jest.mock("@lib/adminLogs");
 
 //Needed in the typescript version of the test so types are inferred correclty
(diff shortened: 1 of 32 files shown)


__tests__/api/id/form/bearer.test.ts
```diff
@@ -32,131 +31,374 @@ describe("/id/[form]/bearer", () => {
...
+        expect(JSON.parse(res._getData()).error).toEqual("Not Found");
+      });
     });
-
-    it("Should return a 200 status code and a valid token if it exists for a form.", async () => {
-      // Mocking executeQuery to return a valid bearer token
-
-      (prismaMock.template.findUnique as jest.MockedFunction<any>).mockResolvedValue({
-        bearerToken: "testBearerToken",
+    describe("ManageForms Permissions", () => {
```
craigzour, line 162: In one of your previous pull request you have used `test.each` in order to test multiple http methods for a same test. Based on that I have used `describe.each` ([while I was adding tests to users and templates](https://github.com/cds-snc/platform-forms-client/commit/d7ed658069cb2c044681440f5eebf045045a334c#diff-ddb612901d321b473788904cf6b078ff5bc7168860ac034733876dc21bf0a3acR66)) in order to run the same tests on different privileges. 
If you like it I could take care of doing a little refactoring to these tests before making my own last pull request for permissions.

craigzour (approved):

__tests__/api/id/form/bearer.test.ts
bryan-robitaille: It would be great If you could refactor for those that are common and should have the same result for each.  Initially why I split them out was because some of same tests should fail between Base and ManageForms privileges and I wasn't sure how I was going to organize them in the test suite.

PR #1155, opened 2022-10-20
Feature/add permission tests

craigzour:
# Summary | Résumé

- Added more unit tests to Users API
- Added more unit tests to Templates API
- Refactored a few tests for Bearer token API


bryan-robitaille (approved):
LGTM

PR #1158, opened 2022-10-21
fix: Prevent prototype pollution when importing a form

### LINKED ISSUE #1118: Prototype Pollution from JSON file upload to Form Builder
When a JSON file is uploaded to the form builder. The lines posted below will check if the content is a string and then immediately call `JSON.parse` to create the object. This opens up a vector for prototype pollution. `JSON.parse` will parse out a `__proto__` key in JSON as the prototype of the object generated. See Mozilla docs on Object initialization. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#object_literal_syntax_vs._json 

https://github.com/cds-snc/platform-forms-client/blob/develop/components/form-builder/layout/Start.tsx#L63-L99

Since the Form Builder is to be made public and JSON will come from untrusted sources this should be addressed

Potential solution 

```
const cleaner = (key, value) => key === "__proto__" ? undefined : value;

// demo
let json = '{"user":{"__proto__":{"admin": true}}}';

let cleanJSON = JSON.parse(json, cleaner)
```

pcraig3:
# Summary | Résumé

Prototype pollution is the idea that you pass in a JSON object representation that after being parsed, sneakily sets attributes on your object by taking advantage of Javascript object prototypes.

Here is an article explaining it: https://futurestud.io/tutorials/node-js-securely-parse-json

They recommend pulling in a library but instead I am using a simpler method that disallows the two keys "__proto__" and "contstructor", which near as I can tell, [is what they are doing in their library](https://github.com/supercharge/json/blob/main/src/parse.ts#L48).

Resolves #1118 

Resources:
- https://futurestud.io/tutorials/node-js-securely-parse-json
- https://stackoverflow.com/a/63927372


bryan-robitaille (dismissed):
👍  LGTM

timarney (approved):

PR #1159, opened 2022-10-21
fix: change depreciated Buffer constructor

Moro-Code:
# Summary | Résumé

No issue. Staging for some reason is not setting the statusText on the response from axios


bryan-robitaille (approved):
👍

PR #1160, opened 2022-10-21
fix: remove statusText check from confirm callback in useAuth

Moro-Code:
# Summary | Résumé

Another minor PR to remove the `statusText` check from the useAuth code


bryan-robitaille (approved):
👍

PR #1161, opened 2022-10-21
feat: enhance getAllTemplates API function to return either all or some templates depending on user privileges

craigzour:
# Summary | Résumé

- Enhanced `getAllTemplates` API function to return either all or some templates depending on user privileges

### DIFF
--- a/lib/templates.ts
+++ b/lib/templates.ts
@@ -4,7 +4,7 @@ import { prisma, prismaErrors } from "@lib/integration/prismaConnector";
 import { PublicFormRecord, SubmissionProperties, FormRecord, BetterOmit } from "@lib/types";
 import { Prisma, User } from "@prisma/client";
 import jwt, { Secret } from "jsonwebtoken";
-import { checkPrivileges } from "./privileges";
+import { checkPrivileges, checkPrivilegesAsBoolean } from "./privileges";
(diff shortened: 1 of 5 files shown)


lib/templates.ts
```diff
@@ -64,11 +64,30 @@ async function _createTemplate(
...
  * @returns An array of Form Records
  */
-async function _getAllTemplates(ability: Ability): Promise<Array<FormRecord>> {
+async function _getAllTemplates(ability: Ability, userID: string): Promise<Array<FormRecord>> {
   checkPrivileges(ability, [{ action: "view", subject: "FormRecord" }]);
 
+  const canUserAccessAllTemplates = checkPrivilegesAsBoolean(ability, [
+    {
+      action: "view",
+      subject: {
```
bryan-robitaille, line 73: With the function signature we have you should just be able to pass in:
```js
[{ action: "view", subject: "FormRecord"}]

bryan-robitaille (approved):
One small little change if we really wanted to but I'll leave that up to you.

lib/templates.ts
bryan-robitaille: Nice, thanks for adding the comment in the code :)

bryan-robitaille (approved):
:+1

PR #1175, opened 2022-10-24
Page access restrictions

bryan-robitaille:
# Summary | Résumé
This PR:
- wraps the various pages that require privileges and ensures the user has the minimum privileges required to be interactive.
- fixes a security issue where on acceptable-use API any valid session had the ability to modify any other user's acceptance flag.  The user id is now pulled from the active session and is not passed in the body of the request.
- disables Vault page and API as the feature is not in use until a future major release
-   small refactors for code style commonality and more interactive state changes.

### DIFF
--- a/pages/admin/form-builder.tsx
+++ b/pages/admin/form-builder.tsx
@@ -3,7 +3,6 @@ import { serverSideTranslations } from "next-i18next/serverSideTranslations";
 import { requireAuthentication } from "@lib/auth";
 import { Layout } from "../../components/form-builder/layout/Layout";
 import { User } from "next-auth";
-import { createAbility } from "@lib/policyBuilder";
 
 type WelcomeProps = {
   user: User;
@@ -21,16 +20,7 @@ const Welcome: React.FC<WelcomeProps> = () => {
   );
 };
 
-export const getServerSideProps = requireAuthentication(async ({ user, locale }) => {
-  const userAbility = user?.privileges && createAbility(user.privileges);
-  if (userAbility?.cannot("update", "FormRecord")) {
-    return {
-      redirect: {
-        destination: `/${locale}/admin/unauthorized/`,
-        permanent: false,
-      },
-    };
-  }
+export const getServerSideProps = requireAuthentication(async ({ user: { ability }, locale }) => {
   return {
     props: {
       ...(locale && (await serverSideTranslations(locale, ["common", "form-builder"]))),
--- a/pages/admin/form-builder.tsx
+++ b/pages/admin/form-builder.tsx
@@ -3,6 +3,7 @@ import { serverSideTranslations } from "next-i18next/serverSideTranslations";
 import { requireAuthentication } from "@lib/auth";
 import { Layout } from "../../components/form-builder/layout/Layout";
 import { User } from "next-auth";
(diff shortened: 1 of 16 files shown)


pages/admin/form-builder.tsx
```diff
@@ -21,16 +20,7 @@ const Welcome: React.FC<WelcomeProps> = () => {
...
-  const userAbility = user?.privileges && createAbility(user.privileges);
-  if (userAbility?.cannot("update", "FormRecord")) {
-    return {
-      redirect: {
-        destination: `/${locale}/admin/unauthorized/`,
-        permanent: false,
-      },
-    };
-  }
+export const getServerSideProps = requireAuthentication(async ({ user: { ability }, locale }) => {
```
craigzour, line 23: `ability` is detected as not being used. Did you just add it for the future?

pages/admin/privileges.tsx
```diff
@@ -223,15 +224,17 @@ const Privileges = ({ allPrivileges }: { allPrivileges: Privilege[] }): React.Re
...
-                setModifyMode(true);
-              }}
-            >
-              Create
-            </Button>
+            {ability?.can("create", "Privilege") && (
+              <Button
+                type="button"
+                onClick={() => {
+                  setSelectedPrivealge(null);
```
craigzour, line 231: small typo `privilege`

pages/admin/users.tsx
```diff
@@ -178,6 +179,14 @@ const Users = ({
 export default Users;
 
 export const getServerSideProps = requireAuthentication(async ({ user: { ability }, locale }) => {
+  checkPrivileges(
+    ability,
+    [
+      { action: "view", subject: "User" },
+      { action: "view", subject: "Privilege" },
+    ],
+    "all"
```
craigzour, line 188: `all` is already the default behavior if I am correct

pages/id/[form]/retrieval.tsx
```diff
@@ -2,6 +2,7 @@ import { serverSideTranslations } from "next-i18next/serverSideTranslations";
 import { requireAuthentication } from "@lib/auth";
 import { useTranslation } from "next-i18next";
 import React from "react";
+import { checkPrivileges } from "@lib/privileges";
```
Comment (craigzour) [inline pages/id/[form]/retrieval.tsx:5]: Looks like this is not needed anymore.

craigzour (changes requested):
We can now see the light at the end of the tunnel! There is only a few small changes. 
I also did notice a bunch of unused variables/imports reported by the ESLint report while doing the review. If you do not have time right now I can take care of it when we will be doing the final PR.

pages/admin/users.tsx
bryan-robitaille: It is the default, I just added it for clarity.

pages/admin/form-builder.tsx
bryan-robitaille: weird... I had added a line checking for a rule but it seems to have disappeared.  
Good catch

pages/admin/privileges.tsx
bryan-robitaille: 👍

craigzour (approved):

PR #1180, opened 2022-10-26
fix: allow FileInput component to render in Storybook without error

mnigh:
# Summary | Résumé

Updates FileInput story's inputProps and Formik's initialValues to avoid type error that blocks rendering in Storybook.

| Screen Shot 2022-10-26 at 11 52 01 | Screen Shot 2022-10-26 at 11 52 58 |


Moro-Code (dismissed):
Hello @mnigh ! Thanks for your first PR to GC Forms ! For this component, the `value.name` in actuality will always be defined when used in our application. We use a library called `Formik` to handle state and actions from form components and their children. Formik allows us to define initial values for inputs and we use this function to do so based on the input types included in the form 

https://github.com/cds-snc/platform-forms-client/blob/develop/lib/formBuilder.tsx#L314  

What I would suggest is to modify the story file so that the initial value here https://github.com/cds-snc/platform-forms-client/blob/develop/components/forms/FileInput/FileInput.stories.tsx#L28 so that the name field is included. 

Once again thank you for contributing to GC Forms

mnigh:
> What I would suggest is to modify the story file so that the initial value here https://github.com/cds-snc/platform-forms-client/blob/develop/components/forms/FileInput/FileInput.stories.tsx#L28 so that the name field is included.

Thanks for your feedback @Moro-Code. I tried adding `name` to the storybook file but that did not solve the problem.

`initialValues={{ file: "", name: "" }}`

I did notice that `value` was being extracted from `field` just before being used as the default value for its state and so removed that and [set it directly instead](https://github.com/cds-snc/platform-forms-client/pull/1180/commits/360df7afa660f994c8f2dc2b12c7ac451ace724f). Is this an acceptable solution?

bryan-robitaille:
Hi @mnigh, I would rather not adapt the application code to fit a Storybook initialization issue.
The changes below fix the initialization of the component.  The 'name' of the component passed into props must match the property set on `initialValues` in order for Formik to associate the value with the form input.

Screenshot 2022-10-28 at 12 36 42 PM

bryan-robitaille (approved):
Thanks for making the changes.

PR #1182, opened 2022-10-27
fix: Update design for form title and rich text fields

pcraig3:
# Summary | Résumé

This PR brings us more into alignment with what's in the mockups. 

- Our rich text fields are no longer full-width
- The form title looks like a "title" input now (bottom underline) instead of a regular text input
- Tweaked the "design" icon on the start page because it was too small

Where I made CSS changes, I tried to convert things to tailwind as much as I could. There were still some holdouts, but it moves the needle a little bit.

### Design page screenshots

| Screen Shot 2022-10-27 at 12 07 37 | Screen Shot 2022-10-27 at 12 07 34 |

### Translate page screenshots

|   Screen Shot 2022-10-27 at 14 38 59   |      Screen Shot 2022-10-27 at 14 38 56  |

### DIFF
--- a/components/form-builder/elements/RichText.tsx
+++ b/components/form-builder/elements/RichText.tsx
@@ -1,12 +1,7 @@
 import React from "react";
-import styled from "styled-components";
 import useTemplateStore from "../store/useTemplateStore";
 import { RichTextEditor } from "../lexical-editor/RichTextEditor";
 import { LocalizedElementProperties } from "../types";
-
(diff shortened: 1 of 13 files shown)


timarney (dismissed):
Approving 

As a nice to have even in a follow-up it would be nice to equalize the rich text area "entry text area" as discussed.

timarney (approved):
🥇

PR #1187, opened 2022-10-31
Add ts-node to production dependencies

bryan-robitaille:
# Summary | Résumé
Fixes issue where `ts-node` was only listed as a development dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "styled-components": "5.3.6",
     "swr": "1.3.0",
     "ts-command-line-args": "^2.3.1",
+    "ts-node": "10.9.1",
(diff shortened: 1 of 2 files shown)


craigzour (approved):

package.json
```diff
@@ -88,6 +88,7 @@
     "styled-components": "5.3.6",
     "swr": "1.3.0",
     "ts-command-line-args": "^2.3.1",
+    "ts-node": "10.9.1",
```
thiessenp-cds, line 91: Looks like a change to fixed version and no longer a compatible version dependency. This was probably intentional and or doesn't matter but noting it

craigzour (approved):

PR #1189, opened 2022-10-31
SVG icon consistency and a11y improvements

dsamojlenko:
# Summary | Résumé

Improves the consistency and accessibility of our SVG icons.
All icons now have the same signature, ie: `<Icon className="" title="" />`

Both props are optional.

The Accessibility features are modelled after the @sytled-icons library. There are two accessibility modes. By default, icons are considered decorative and receive the `aria-hidden="true"` prop:

```
// this icon
<CircleCheckIcon />

// will result in
<svg aria-hidden="true" role="presentation">...</svg>
```

Alternatively, if the icon should convey meaning, you may set the title prop which adds a `<title>` element to the svg which can be read by screen readers, and sets `role="img"` prop:

```
// this icon
<CircleCheckIcon title="Completed item" />

// will result in
<svg role="img">
    <title>Completed item</title> 
    ...
</svg>
```


pcraig3 (approved):
Looks good, and helps me with my upcoming PR 👍

PR #1191, opened 2022-11-01
Lexical editor Tab/shift-Tab to escape

dsamojlenko:
# Summary | Résumé

Adds a Lexical plugin to intercept the Tab keypress to exit the editor rather than insert a Tab character.

Previously, once inside the editor textarea, pressing Tab would insert a Tab character. But since indents are not processed by the markdown renderer, they just got stripped anyway.

Now, once inside the editor, pressing Tab will escape the textarea and set focus to the next item in the tab order.


timarney (approved):
Works great

PR #1192, opened 2022-11-01
Tailwind buttons

pcraig3:
# Summary | Résumé

We are wanting to move away from styled-components and the button component is among the most used we have.

This PR replaces the buttons we use in the app with a new tailwind button component. Once this is shipped, we can also do something similar with the text input.

Instead of screenshots (they would just look the same), I am going to list all the places that buttons appear and check off if they have been replaced.

Buttons:

- [x] "Save" buttons
- [x] "Copy to clipboard" button (on "Save" page) 
- [x] "Save" secondary button (under test-data-delivery)
- [x] "Delete" button (on settings page)
- [x] "Switch language" and "Download .CSV" buttons (on Translate page)
- [x] "Publish" button (on Publish page)
- [x] Form builder buttons
  - [x] "Add element"
  - [x] "Remove option" icon button in the form builder
  - [x] "Save" and "Cancel" in the modal
  - [x] "Close" button at the top of the modal
  - [x] "Move up/Move down/Remove/More" buttons in Modal
  - [x] "Add option" in modal (this button has no border)
 

All the buttons should look more or less like they did before, but they are all based in tailwind now.


components/form-builder/panel/Modal.tsx
```diff
@@ -3,7 +3,8 @@ import styled from "styled-components";
 import { useTranslation } from "next-i18next";
 import PropTypes from "prop-types";
 
-import { Button, FancyButton } from "./Button";
+import { Button as OldButton } from "./Button";
```
timarney, line 6: Will this get swapped at some point ?

components/form-builder/translate/DownloadCSV.tsx
```diff
@@ -99,5 +99,9 @@ export const DownloadCSV = () => {
     URL.revokeObjectURL(url);
   };
 
-  return <FancyButton onClick={generateCSV}>Download .csv</FancyButton>;
+  return (
+    <Button onClick={generateCSV} theme="secondary">
+      Download .csv
```
timarney, line 104: Maybe wrap this in a translation while your in here

components/form-builder/panel/Modal.tsx
pcraig3: Yes, I will do it later in this PR. I am going to remove that other "Button" file entirely.

components/form-builder/translate/DownloadCSV.tsx
pcraig3: 👍

timarney (approved):

PR #1195, opened 2022-11-01
Update link editor

dsamojlenko:
# Summary | Résumé

Link editor improvements. Adds an Edit button per the Lexical Sandbox example. Also restructures the code a bit to be more like the example from the Sandbox.


timarney:
1) Let's get this focus background fixed up

Screen Shot 2022-11-01 at 3 47 19 PM

2) Thinking it would be nice if we could avoid the link in the popup from being a real link --- too easy to click out

3) As discussed it would be nice to get link editor to close when you tab / esc out etc... but maybe that's in a followup PR

timarney (approved):

timarney:
@dsamojlenko --- you should be able to merge this one

PR #1194, opened 2022-11-01
Start - perf / re-render fixes add debounce

timarney:
# Summary | Résumé

This PR adds an initial perf improvement by adding debase to the Title Input ---   

debounce

I have also added a small example of using Zustand selectors (doesn't improve anything in this case but starts the process) 

# Test instructions | Instructions pour tester la modification

Ensure highlight re-renders is turned on --- update the question note the re-render is debounced

Screen Shot 2022-11-01 at 11 42 28 AM

<hr>

We can do much more here but looking to break these updates into multiple PRs

Sequential steps (1., 2., 3., ...) that describe how to test this change. You should include
anything outside the README that will help the developer access your changes and be able to test
them such as any environmental setup steps and/or any time-based elements that this requires.
This will help a developer test things out without too much detective work.

# Unresolved questions / Out of scope | Questions non résolues ou hors sujet

Are there any related issues or tangent features you consider out of scope for
this issue that could be addressed in the future? If so please create issues and provide
links in this section

# Pull Request Checklist

Please complete the following items in the checklist before you request a review:

- [x] Have you completely tested the functionality of change introduced in this PR? Is the PR solving the problem it's meant to solve within the scope of the related issue?
- [x] The PR does not introduce any new issues such as failed tests, console warnings or new bugs.
- [ ] If this PR adds a package have you ensured its licensed correctly and does not add additional security issues?
- [ ] Is the code clean, readable and maintainable? Is it easy to understand and comprehend.
- [ ] Does your code have adequate comprehensible comments? Do new functions have [docstrings](https://tsdoc.org/)?
- [ ] Have you modified the change log and updated any relevant documentation?
- [ ] Is there adequate test coverage? Both unit tests and end-to-end tests where applicable?
- [ ] If your PR is touching any UI is it accessible? Have you tested it with a screen reader? Have you tested it with automated testing tools such as axe?

PR #1197, opened 2022-11-01
Prototype form-builder routing

bryan-robitaille:
# Summary | Résumé

Enables routing through the URL to a specific action and form using the pattern `/form-builder/${action}/{formID}`
This PR also sets the initial values of the Navigation and Template stores through props which will reduce the number of page renders required before interactivity is ready for the user.

Some existing forms will throw errors when loaded into form-builder due to existing complex components or missing description fields.  Any additional work to make the feature more robust will be in a future PR.


components/form-builder/layout/Publish.tsx
```diff
@@ -34,6 +37,7 @@ export const Publish = () => {
     if (result && result?.error) {
       setError(true);
       return;
+      s;
```
timarney: typo?

bryan-robitaille: Just pushed a commit to fix.  A stray keyboard command gone awry.

timarney:
Looks good @bryan-robitaille 

Per you note we'll end up with some errors for description fields  etc.. but this moves things forward nicely

Are you looking for an approval to merge at this point?  Thinking that would be best so we can get some other WIP PRs updated to use this.

bryan-robitaille:
Let's merge so we can start using the new Store pattern and I'll create a new PR to fix the issues importing some of the forms into Form Builder.

timarney (approved):
🥇

timarney:
FYI - I started experimenting with debounce on https://github.com/cds-snc/platform-forms-client/pull/1194 

With this PR starting the move to selectors + some further updates we can stop all the extra renders for key "panel" updates.

PR #1206, opened 2022-11-03
fix: publishingStatus migration counter

craigzour:
# Summary | Résumé

- Fixed issue with `publishingStatus` migration counter


bryan-robitaille (approved):
👍

PR #1207, opened 2022-11-03
fix: Use tailwind for inputs

pcraig3:
# Summary | Résumé

Similar to [tailwind-buttons](https://github.com/cds-snc/platform-forms-client/pull/1192), this pull request removes the `styled-component` Input component we were using previously in favour of an Input component that uses `tailwind`.
- It also introduces a `TextArea` component, which is actually only used in one place (I thought we had more of them). 
- I also moved the MultipleChoice component into `/shared` because it's basically where it should live

Note that I didn't replace the inputs or textareas on the `Translate` page, because they don't follow the same styling as the rest of our inputs/textareas. They are more or less naked HTML elements with a smidge of CSS thrown at them. Probably best idea is to put all the CSS rules for Translate into the `form-builder.scss` file instead of going element by element on that page.

## Inputs replaced

- [x] Form title
- [x] Question Title
- [x] Option name
- [x] Modal > Question title
- [x] Modal > Description (this is the only textarea)
- [x] Modal > Character limit
- [x] Settings > Save delivery address


timarney:
Not an issue with this PR but wondering of the "title / question" inputs should have the "blue" outline

Screen Shot 2022-11-04 at 8 51 42 AM

timarney:
Looks good let me know if you want to make updates per 
https://github.com/cds-snc/platform-forms-client/pull/1207#issuecomment-1303483200 in a new PR

timarney (approved):

Issue #1203, opened 2022-11-02
Add admin feature to assign users to template


craigzour:
**Acceptance criteria**
- [ ] Create API to assign users to template
- [ ] Create UI (in /admin) to assign users to template

PR #1210, opened 2022-11-04
feat: admin can now assign users to template

craigzour:
# Summary | Résumé

closes https://github.com/cds-snc/platform-forms-client/issues/1203

- Added admin feature to assign users to template

Screen Shot 2022-11-04 at 2 23 33 PM

Screen Shot 2022-11-04 at 2 23 52 PM

- Navigate to http://localhost:3000/admin/view-templates
- Assign users to template using the Assign button in the forms table (you may want to create fake users to see how the UI behaves)
- Using `yarn prisma:studio` you can verify that the assignations are saved in the database

# Pull Request Checklist

### DIFF
--- a/pages/id/[form]/users.tsx
+++ b/pages/id/[form]/users.tsx
@@ -128,7 +128,9 @@ export const getServerSideProps = requireAuthentication(
     const templateWithAssociatedUsers = await getTemplateWithAssociatedUsers(ability, formID);
     if (!templateWithAssociatedUsers) return redirect(locale);
 
-    const allUsers = await getUsers(ability);
+    const allUsers = (await getUsers(ability)).map((user) => {
+      return { id: user.id, name: user.name };
+    });
 
     return {
       props: {
--- a/pages/id/[form]/users.tsx
+++ b/pages/id/[form]/users.tsx
@@ -1,0 +1,144 @@
+import { serverSideTranslations } from "next-i18next/serverSideTranslations";
+import { requireAuthentication } from "@lib/auth";
+import { checkPrivileges } from "@lib/privileges";
+import { useTranslation } from "next-i18next";
+import React, { useState } from "react";
+import Select from "react-select";
+import makeAnimated from "react-select/animated";
+import { getUsers } from "@lib/users";
+import { Button } from "@components/forms";
+import { getTemplateWithAssociatedUsers } from "@lib/templates";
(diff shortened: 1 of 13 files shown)


pages/id/[form]/users.tsx
```diff
@@ -0,0 +1,144 @@
...
+      { action: "update", subject: "User" },
+    ]);
+
+    const formID = params?.form;
+    if (!formID || Array.isArray(formID)) return redirect(locale);
+
+    const templateWithAssociatedUsers = await getTemplateWithAssociatedUsers(ability, formID);
+    if (!templateWithAssociatedUsers) return redirect(locale);
+
+    const allUsers = await getUsers(ability);
```
Comment (bryan-robitaille) [inline pages/id/[form]/users.tsx:131]: I'd be careful about passing `allUsers` directly to the page props as it contains much more information about a user than a the page needs.
I'd suggest parsing out the properties needed by the page `{id, name}` before passing the object down as props.

pages/id/[form]/users.tsx
```diff
@@ -0,0 +1,144 @@
...
+  async ({ user: { ability }, locale, params }) => {
+    checkPrivileges(ability, [
+      { action: "update", subject: "FormRecord" },
+      { action: "update", subject: "User" },
+    ]);
+
+    const formID = params?.form;
+    if (!formID || Array.isArray(formID)) return redirect(locale);
+
+    const templateWithAssociatedUsers = await getTemplateWithAssociatedUsers(ability, formID);
```
Comment (bryan-robitaille) [inline pages/id/[form]/users.tsx:128]: The users object returned from the function contains a lot of data that isn't used.  Just a though but I wonder if the select on the users should be limited to `id` and `name`.

bryan-robitaille (changes requested):

bryan-robitaille:
Weird that my review comment didn't make it through but...

Great Work!
I would just suggest that, as a security best practice, we reduce the amount of data sent to page to only the data that will be used.

pages/id/[form]/users.tsx
Comment (craigzour) [reply pages/id/[form]/users.tsx:128]: You are right we should just pass the information that we need!

pages/id/[form]/users.tsx
Comment (craigzour) [reply pages/id/[form]/users.tsx:131]: Did change it :)

bryan-robitaille (approved):
LGTM 👍

PR #1231, opened 2022-11-07
Adds initial work for feature myforms

thiessenp-cds:
# Summary | Résumé

*The all new, better than ever, myforms PR!*

Adds MyForm feature to display a list of templates in cards. Using a tab interface, the cards can be filtered by draft, published and all. 

Run the app and click the MyForms link in the top admin nav. If you have no templates, first create a few in the FormBuilder. Try clicking the different tabs "draft", "published", "all". The related templates should be displayed. Each card also has a menu that displays a number of actions (currently pointing to fake links).

Add action links to the card menu. 

# Pull Request Checklist


pages/myforms/[[...path]].tsx
```diff
@@ -0,0 +1,138 @@
...
+}
+
+export const getServerSideProps = requireAuthentication(
+  async ({ user: { ability, id }, locale }) => {
+    {
+      checkPrivileges(
+        ability,
+        [
+          { action: "view", subject: "FormRecord" },
+          { action: "update", subject: "FormRecord" },
```
Comment (bryan-robitaille) [inline pages/myforms/[[...path]].tsx:108]: To be able to view the page we may just want to leave it as a check on `view / FormRecord`.  The abilility to `update / FormRecord` will be checked later on when they attempt that action.

bryan-robitaille (changes requested):
Great work!
I really like the tailwind/line-clamp package you implemented.
One quick change on a permission and we should be good to go for this initial implementation.

craigzour:
@thiessenp-cds I have tested your code locally and it works fine but I have one question for you. When I press on the `My Forms` tab and I land on that page there is no cards being displayed even though I have two existing draft forms. Is it a behaviour that was requested by the design team? Should not we display all forms by default?

lib/templates.ts
```diff
@@ -419,19 +421,25 @@ const _parseTemplate = (template: {
   id: string;
   jsonConfig: Prisma.JsonValue;
   isPublished: boolean;
+  updated_at?: Date;
```
craigzour, line 424: With the addition of the new `update_at` and `created_at` properties, you will need to make sure that whenever the `_parseTemplate` function is called the prisma query has to get those values. It is either that or we set both values to `n/a` (line 441, 442) if they are undefined.

craigzour (changes requested):
Nice work @thiessenp-cds !! Just found a bug that will impact a few functions within the `lib/templates.ts` file.

pages/myforms/[[...path]].tsx
Comment (thiessenp-cds) [reply pages/myforms/[[...path]].tsx]: Thanks, and thanks for re-reading this new PR :). Makes sense, done.

lib/templates.ts
thiessenp-cds: Thanks for taking a look. That makes sense about a check vs. having props with undefined/NaN.. Sure done. I went with the spread idea (from Bryan), probably the most obvious about what's going on.

thiessenp-cds:
> @thiessenp-cds I have tested your code locally and it works fine but I have one question for you. When I press on the `My Forms` tab and I land on that page there is no cards being displayed even though I have two existing draft forms. Is it a behaviour that was requested by the design team? Should not we display all forms by default?

From what I remember the design had Published as the default initial card view. Though I think you're right about it being confusing to show an empty templates page when you have templates, just not in that form state. Maybe we could add some logic to show the "All" or "Draft" templates page if "Published" is empty.

bryan-robitaille:
We could also just ask design if we can set 'all' as the default instead of published.

thiessenp-cds:
> We could also just ask design if we can set 'all' as the default instead of published.

I like it. Will do that :)

bryan-robitaille (approved):

craigzour (approved):

PR #1244, opened 2022-11-09
feat: Change the top "save" button depending on if users are logged in or not 

pcraig3:
# Summary | Résumé

This PR changes the top "Save" button's functionality depending on if users are logged in or not.

- If logged in, users will save a draft 
- If not logged in, users will get a CSV download

It also does a handful of other things:
- Add `withMessage` wrapper for Button components
  - Update Buttons on Save/Share pages to use it
- Exports the button classes so that we can use them for links
- Fixes the RichText areas on the "translate" tab not being the same size
- Fixes some TypeScript errors we were seeing in our tests.

A note here that the `WithButton` component was pretty complicated to get working.

You can see the general approach for creating a wrapping component [in this StackOverflow answer](https://stackoverflow.com/questions/54602013/component-definition-is-missing-display-name-on-hoc). Then I wanted to add an extra function to run on the "onClick" handler, for Button components that already had an "onClick" defined. I saw that Kent C. Dodds has an example that uses `callAll` [in his "props getters" post](https://kentcdodds.com/blog/how-to-give-rendering-control-to-users-with-prop-getters). 

The idea is that we can create a new component (using `React.createElement`), passing in all the `props` it comes with, and then we can inject a new function into "onClick" after the original function runs. I think it's small enough to make sense of at this point, but it was very annoying to get working.

### Button With Message

| Save (logged in)                                          | Save (not logged in)                                        | Share page                                    |
| ----------------------------------------------- | -------------------------------------------- | -------------------------------------------- |
| Screen Recording 2022-11-09 at 15 41 22  |  Screen Recording 2022-11-09 at 15 41 48  | Screen Recording 2022-11-09 at 15 42 30 |

### Translate page screenshots

This broke previously but we are back, baby.

|   Screen Shot 2022-10-27 at 14 38 59   |      Screen Shot 2022-10-27 at 14 38 56  |


timarney (approved):
Works great --- thinking we should have a follow-up PR maybe with a live region to announce the `saved` action etc...

PR #1250, opened 2022-11-10
Feature/request publish UI

thiessenp-cds:
# Summary | Résumé

Adds Request to Publish UI screens.

Navigate to /signup/account-created and click the "Unlock publishing" link (styled like a button). Try filling in empty fields and an invalid email, error messages should be displayed when keying "enter" or clicking "Submit". Clicking "Submit" with valid form detail should display the "what's next page".
Clicking on the "Skip this step for now" link should take the user to the "My forms" page.

**Important: Test with email `forms-formulaires@cds-snc.ca`** (can be changed in `publish.ts` line 7)

Request to publish form submission errors need more work. Currently a generic catch all error is displayed in the Alert component. Also the Alert component does not really work well for error alerts and probably a new component should be created.

# Pull Request Checklist


.eslintrc.js
```diff
@@ -16,6 +16,14 @@ module.exports = {
       plugins: ["@typescript-eslint", "jsx-a11y", "prettier"],
       rules: {
         "@typescript-eslint/await-thenable": "error",
+        "jsx-a11y/anchor-is-valid": [
```
thiessenp-cds, line 19: The Next/Link component oddly does not receive a classes but requires a child anchor link to customize the styles. But the link also requires a URL and errors if a URL is added on anchor. But an anchor without an href is invalid HTML. 
The below rule is used to work with (around) the Next/Link implementation.

bryan-robitaille: Have you tried using `passHref` on the Link component?
https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag

bryan-robitaille: I tried with `passHref` and the rule still thinks it's an invalid anchor.  You're right we'll need to add an exception to the rule.

pages/unlock-publishing.tsx
```diff
@@ -0,0 +1,256 @@
...
+          </div>
+        </>
+      )}
+    </>
+  );
+}
+
+export const getServerSideProps = requireAuthentication(async ({ user: { ability }, locale }) => {
+  {
+    checkPrivileges(ability, [{ action: "view", subject: "FormRecord" }]);
```
thiessenp-cds, line 248: This may not be the right privilege/way to check if a new user can request publishing access.

components/form-builder/layout/TestDataDelivery.tsx
```diff
@@ -57,6 +57,8 @@ export const TestDataDelivery = () => {
     <div>
       <div className="mb-8 bg-blue-200 p-5">
         {t("submittedResponsesText", { email })}{" "}
+        {/* TODO this should probably be a button - for a future refactor */}
```
thiessenp-cds, line 60: Thinking about this now, flagging TODOs in the code may not be the best way. But I'll leave this here for now - curious what others thing?

pages/signup/account-created.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+import React from "react";
+import { useTranslation } from "next-i18next";
+import { StyledLink } from "@components/globals/StyledLink/StyledLink";
+
+export default function AccountCreated() {
+  const { t, i18n } = useTranslation(["signup"]);
+
+  return (
+    <>
```
thiessenp-cds, line 13: I wasn't sure if there was enough content to justify a markdown page or a normal html/templated page. I went with html/templated pages.

components/globals/StyledLink/StyledLink.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const { children, className } = props;
+
+    return (
+      <a href={to} className={className} ref={ref}>
+        {children}
+      </a>
+    );
+  }
+);
+WrappedLink.displayName = "WrappedLink";
```
thiessenp-cds, line 48: Could also wrap in a normal component to get a "displayName" but this seemed simpler

pages/unlock-publishing.tsx
```diff
@@ -0,0 +1,256 @@
...
+      {!hasRequestedPublishing && (
+        <>
+          <Formik
+            initialValues={{ managerEmail: "", department: "", goals: "" }}
+            onSubmit={async ({ managerEmail, department, goals }) => {
+              setSubmitting(true);
+              try {
+                const response = await handleRequestPublishing(managerEmail, department, goals);
+                setSubmitting(false);
+                if (response.status !== 200) {
```
thiessenp-cds, line 75: Error handling needs more work in the future.

pages/unlock-publishing.tsx
bryan-robitaille: To check if a user has publishing ability you can use `ability.can("update", "FormRecord", "isPublished");`

craigzour (dismissed):
LGTM!

pages/unlock-publishing.tsx
bryan-robitaille: I also tweaked `response.status` as `response` could be void if an error is thrown in `handleRequestPublising`.
For now just adding `response?.status` is good enough but you're right that the error handling could use a refactor in the future.

pages/signup/account-created.tsx
thiessenp-cds: Markdown would have been easier - note for future when having this many strings to localize

craigzour (approved):

lib/auth.ts
```diff
@@ -56,12 +57,15 @@ export function requireAuthentication(
...
 
-      if (!session.user.acceptableUse && context.resolvedUrl !== "/auth/policy") {
+      if (!session.user.acceptableUse && !context.req.url?.startsWith("/auth/policy")) {
         // If they haven't agreed to Acceptable Use redirect to policy page for acceptance
         // If already on the policy page don't redirect, aka endless redirect loop.
+        // Also check that the path is local and not an external URL
         return {
           redirect: {
-            destination: `/${context.locale}/auth/policy`,
+            destination: `/${context.locale}/auth/policy?referer=${
```
thiessenp-cds, line 66: @bryan-robitaille I wonder if this is what you had in mind?

pages/unlock-publishing.tsx
```diff
@@ -0,0 +1,254 @@
...
+                  </StyledLink>
+                </div>
+              </form>
+            </>
+          )}
+        </Formik>
+      )}
+      {hasRequestedPublishing && (
+        <>
+          <h1>{t("unlockPublishingSubmitted.title")}</h1>
```
thiessenp-cds, line 219: Future Todo: Currently the page updates and is announced which works.  But adding a tabindex=-1 and focussing the H1 on load would make it really clear the page was updated. Though this may take some work and also other places in the app may want to follow this. Probably best to do when refactoring.

pages/unlock-publishing.tsx
```diff
@@ -0,0 +1,254 @@
...
+          </div>
+        </>
+      )}
+    </div>
+  );
+}
+
+export const getServerSideProps = requireAuthentication(async ({ user: { ability }, locale }) => {
+  {
+    checkPrivileges(ability, [{ action: "update", subject: "FormRecord", field: "isPublished" }]);
```
thiessenp-cds, line 246: @bryan-robitaille I think this is closer to being right but the "isPublished", I'm not sure that is how "field" should be used?

lib/auth.ts
bryan-robitaille: Yup, that works 👍

pages/unlock-publishing.tsx
bryan-robitaille: The check is right but with this logic implentation a user that does not have the ability to publish would never be able to reach this page.  The `checkPrivileges` with throw an access control error and the user will be redirected to a Forbidden page if not caught.  There is another function in the `privileges` lib that we can slightly modify and use here: `checkPrivilegesAsBoolean`.  We can then redirect the user back to myforms if they have already have the publishing ability on their profile.

bryan-robitaille (approved):
LGTM 👍

PR #1252, opened 2022-11-10
Fix: announce button messages & allow empty emails when editing

pcraig3:
# Summary | Résumé

This PR fixes 2 small-ish bugs:

1. When you are editing the email address for the form to send to, you were previously not able to delete the whole string, as it would repopulate with the original email. 
2. The screen reader announcement for the button messages would previously only happen once. 

Now, empty emails are okay while editing (if it is empty when you leave the tab, your previous email will end up in that field again). 

And button messages are reset after the message fades away, so now we are physically removing the message from the DOM rather than just hiding it with CSS. Tested with VoiceOver and it seems like we are good now.

## Email editing

| Everytime the field is empty, it re-populates with my email | Empty fields are allowed, and I can enter new emails afterwards |
| Screen Recording 2022-11-10 at 14 33 45 | Screen Recording 2022-11-10 at 14 32 09  |

## Button messages are read by VoiceOver

| VoiceOver reads the message                                          |
| ----------------------------------------------- |
| The message is repeated by VoiceOver (you can see the caption at the bottom) | 
|  Screen Recording 2022-11-10 at 14 27 19 |

### DIFF
--- a/components/form-builder/layout/Layout.tsx
+++ b/components/form-builder/layout/Layout.tsx
@@ -55,8 +55,8 @@ export const Layout = () => {
         updateField("submission.email", data.user.email);
       }
     };
-    !email && setEmail();
-  }, [email, data]);
+    !email && currentTab !== "settings" && setEmail();
+  }, [email, data, currentTab]);
 
   const renderTab = (tab: string) => {
     switch (tab) {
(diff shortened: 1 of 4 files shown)


timarney (dismissed):

timarney:
For next week - let's add email validation to the publish checklist 
- Also can look at the loss of focus for save  --- 2x tab issue

PR #1253, opened 2022-11-10
Add missing translation strings

dsamojlenko:
# Summary | Résumé

Adds a bunch of missing translation strings


pcraig3 (approved):
Thanks for doing this! Conflicts galore if we leave this up for too long.

PR #1254, opened 2022-11-10
Feature/acceptable use page

bryan-robitaille:
# Summary | Résumé

Adds an acceptable use page that a User must accept after logging in to interact with site.

Screenshot 2022-11-10 at 3 56 48 PM

Also adds common form-builder styling to '/myforms', 'login', 'logout', 'acceptableUse', and 'signup/register'

closes #1223

# Test instructions | Instructions pour tester la modification
Acceptable Use Page:
- Login
- Try to navigate to another page that requires authentication or the form-builder
- You should be continually redirected back to acceptable use until you accept
- No path memory is current in play so after accepting you are redirected to myforms.

Styling:
- Browse the site... you hopefully should see any major differences.

# Pull Request Checklist

Please complete the following items in the checklist before you request a review:

- [ ] Have you completely tested the functionality of change introduced in this PR? Is the PR solving the problem it's meant to solve within the scope of the related issue?
- [ ] The PR does not introduce any new issues such as failed tests, console warnings or new bugs.
- [ ] If this PR adds a package have you ensured its licensed correctly and does not add additional security issues?
- [ ] Is the code clean, readable and maintainable? Is it easy to understand and comprehend.
- [ ] Does your code have adequate comprehensible comments? Do new functions have [docstrings](https://tsdoc.org/)?
- [ ] Have you modified the change log and updated any relevant documentation?
- [ ] Is there adequate test coverage? Both unit tests and end-to-end tests where applicable?
- [ ] If your PR is touching any UI is it accessible? Have you tested it with a screen reader? Have you tested it with automated testing tools such as axe?

PR #1255, opened 2022-11-10
Add a skip link to the form builder

pcraig3:
# Summary | Résumé

Adding a skip link to the form builder that skips our 3 levels of navigation and shoots you down straight into the normal page content. `SkipLink` is a super basic component that takes no properties and expects an id of "#content" to be on the page somewhere.

Note: there is a black outline on the divs when we focus them. Is this good/bad? Happy to remove them if we want, or maybe it's good to show people where they are focused. The border lasts for 1 second and then goes away.

Note: I didn't use the translate function (`t`) here on purpose because I figure the `SkipLink` might get used elsewhere and I  didn't want to introduce the whole namespace problem. I'm easy on the approach though. 

- On the one hand, it is also totally self contained right now and the copy is taken from Canada.ca, so it should be pretty reliable
- On the other hand, it's just not how we do things normally.

Used this tutorial for the basic approach:

- https://levelup.gitconnected.com/build-an-accessible-skip-to-content-anchor-link-with-react-140903f3bd7e

### gif

Screen Recording 2022-11-10 at 17 25 35


bryan-robitaille:
Just a heads up that a `SkipLink` component already exists in the app for all other screens and can be found under `components/globals`.  Can this existing component meet your needs for form-builder?

pcraig3:
Ah good point, Bryan. I'll have a look.

pcraig3:
Closing in favour of #1254

PR #1263, opened 2022-11-14
feat: Add page titles to the form builder

pcraig3:
# Summary | Résumé

This PR adds page `<title>`s to the form builder pages. It is not using any dynamic variables (eg, `Preview {your form}` or anything, so it's all pretty simple. We have also changed one of the option titles from "Test data delivery" to "Response delivery", so I updated that as well.

Credit to @craigzour for the French page title translations.

| Page titles from site map w/ translations   |
| ---------------------------------------------------------------------- |
|  Screen Shot 2022-11-14 at 10 40 51 AM  |
|   Screen Shot 2022-11-14 at 10 41 11 AM  |


public/static/locales/fr/form-builder.json
```diff
@@ -43,7 +43,16 @@
   "formSubmissionDisabledInPreview": "[FR] Form submission is disabled in preview",
   "formTitle": "[FR] Form title",
   "french": "Français",
-  "gcForms": "[FR] GC Forms",
+  "gcForms": "Formulaires GC",
+  "gcFormsStart": "Créer",
+  "gcFormsEdit": "Editer",
+  "gcFormsPreview": "Prévisualiser",
+  "gcFormsResponseDelivery": "Livraison des réponses",
+  "gcFormsTranslate": "Traduction",
```
craigzour: @ioanacontu suggested (on Slack) using `Traduire` instead of `Traduction` here. I do agree with her.

craigzour (changes requested):
Just a small translation change but the rest looks good to me :)

craigzour (approved):

PR #1267, opened 2022-11-15
Fix block translations with links

dsamojlenko:
# Summary | Résumé

ConfirmationDescription.tsx was importing the translation function incorrectly.

Updated link markup to include target attribute


pcraig3 (approved):
Looks good, especially that confirmation page description.

👍

PR #1302, opened 2022-11-21
Set PanelActions role to toolbar and add roving index

dsamojlenko:
# Summary | Résumé

Adds role="toolbar" to the Element Actions panel and adds a roving index for keyboard navigation.

Also improves responsiveness of the actions panel.


timarney (dismissed):
Nice -- this is working much better.

Just the one issue as discussed with tabbing to the first action panel given the initial buttons being disabled for the first panel.

timarney (approved):
Tab to first panel actions now works --- for a follow-up PR looks like adding an element doesn't focus the question input.

PR #1304, opened 2022-11-21
Bugfix: set focus on Question when new element added

dsamojlenko:
# Summary | Résumé

Fixes a small bug introduced in the previous PR that caused focus to go to the element actions toolbar after adding a new element. This restores the original behaviour that gives focus to the Question when a new element is added.


timarney (approved):
Nice -- 

Tested with add, remove and duplicate

PR #1321, opened 2022-11-23
Update cypress and github action to latest version

bryan-robitaille:
# Summary | Résumé
Updates Cypress to version 11 (previously version 9) and the related github action to version 4 (previously version 2).
Cypress 11 is apparently optimized for performance and speed so hopefully, this will help our randomly failing Cypress tests in GitHub.

Cypress 11 also has a new feature that allows for Component tests, which could be interesting to explore.


Moro-Code (approved):

thiessenp-cds (approved):
LGTM :)

PR #1242, opened 2022-11-09
wip

craigzour:
# Summary | Résumé

closes https://github.com/cds-snc/platform-forms-client/issues/1226

- Added new API path to handle publishing permission requests (`/api/request/publish`)

# Test instructions | Instructions pour tester la modification

- ?

# Pull Request Checklist

Please complete the following items in the checklist before you request a review:

- [ ] Have you completely tested the functionality of change introduced in this PR? Is the PR solving the problem it's meant to solve within the scope of the related issue?
- [ ] The PR does not introduce any new issues such as failed tests, console warnings or new bugs.
- [ ] If this PR adds a package have you ensured its licensed correctly and does not add additional security issues?
- [ ] Is the code clean, readable and maintainable? Is it easy to understand and comprehend.
- [ ] Does your code have adequate comprehensible comments? Do new functions have [docstrings](https://tsdoc.org/)?
- [ ] Have you modified the change log and updated any relevant documentation?
- [ ] Is there adequate test coverage? Both unit tests and end-to-end tests where applicable?
- [ ] If your PR is touching any UI is it accessible? Have you tested it with a screen reader? Have you tested it with automated testing tools such as axe?

PR #1342, opened 2022-11-28
Feature/contact us

thiessenp-cds:
# Summary | Résumé

Adds two new support pages, Contact Us and Support, and a related API endpoint and component. Note that the pages are for both authenticated and unauthenticated users.

Note: recommend setting the emails to yourself in the help.ts api to make it easier to test.

1. Navigate to /form-builder/support and fill in the form. An email should be sent (see attached). 
2. Navigate to /form-builder/support/contactus and fill in the form. An email should be sent (see attached). 

# Future Work

- Adding a JSON schema to validate the body data for this email API and other similar email related APIs
- Creating a util function to send emails
- Some sort of SPAM prevention for emails e.g. Firewall rule(s) or something

# Pull Request Checklist

(removed images for now - design has changed)


thiessenp-cds:
I'd really like to combine the two pages contactus.tsx and support.tsx into one page. So something like an index that when reached would determine to route to /form-builder/contactus or /form-builder/support. Those paths are part of the design. 

BUT /form-builder already has a catch-all route so I think something like an index/* would break the form-builder router. I briefly looked into whether NextJS supported something like a static file configurable route(s) but I don't think it does. @craigzour @Moro-Code can think of any way to solve the above routing problem?

If not, I guess two separate files with duplicated code are not the end of the world?

craigzour:
> I'd really like to combine the two pages contactus.tsx and support.tsx into one page. So something like an index that when reached would determine to route to /form-builder/contactus or /form-builder/support. Those paths are part of the design.
> 
> BUT /form-builder already has a catch-all route so I think something like an index/* would break the form-builder router. I briefly looked into whether NextJS supported something like a static file configurable route(s) but I don't think it does. @craigzour @Moro-Code can think of any way to solve the above routing problem?
> 
> If not, I guess two separate files with duplicated code are not the end of the world?

@thiessenp-cds I have no idea if this is possible. There are other examples where we have duplicated code in different pages/paths so this is no big deal.
What if we just combine the two pages at the root level something like `https://forms-staging.cdssandbox.xyz/contact`?

craigzour:
@thiessenp-cds Just noticed that the form builder dev team as already reworked the way the form builder navigation works (that was part of the Tech Debt document). It is not a single page anymore so you can probably go back to your initial implementation idea :)

thiessenp-cds:
> @thiessenp-cds Just noticed that the form builder dev team as already reworked the way the form builder navigation works (that was part of the Tech Debt document). It is not a single page anymore so you can probably go back to your initial implementation idea :)

Thanks and thanks for letting me run the idea by you. I'll go with that :)

thiessenp-cds:
> > @thiessenp-cds Just noticed that the form builder dev team as already reworked the way the form builder navigation works (that was part of the Tech Debt document). It is not a single page anymore so you can probably go back to your initial implementation idea :)
> 
> Thanks and thanks for letting me run the idea by you. I'll go with that :)

On second thought, I don't think this won't work anymore. A user going to the root of /form-builder would hit index.tsx but a user going to e.g. /form-build/contactus would not.  So a catchall route would be needed and that would mess up index.tsx I think.

pages/api/request/help.ts
```diff
@@ -0,0 +1,88 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import { cors, middleware, sessionExists } from "@lib/middleware";
+import { NotifyClient } from "notifications-node-client";
+import { logMessage } from "@lib/logger";
+import { MiddlewareProps } from "@lib/types";
+
+const SUPPORT_EMAIL_ADDRESS = "assistance+forms@cds-snc.freshdesk.com";
+const CONTACTUS_EMAIL_ADDRESS = "jose.jimenez@cds-snc.ca";
```
Moro-Code, line 8: Should these be environment variables. It doesn't seem like a good idea to have them hardcoded in the application

pages/api/request/help.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const requestSupport = async (
+  req: NextApiRequest,
+  res: NextApiResponse,
+  { session }: MiddlewareProps
+) => {
+  try {
+    if (!session) return res.status(403).json({});
```
Moro-Code, line 16: Not sure you need this check. you do have the sessionExists middleware which does it

pages/api/request/help.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    const { type, name, email, request, description } = req.body;
```
Moro-Code, line 18: Perhaps it's worth creating a jsonschema for this and using that to validate the body of the request through our existing middleware

pages/api/request/help.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    // Here is the documentation for the `sendEmail` function: https://docs.notifications.service.gov.uk/node.html#send-an-email
+    await notifyClient.sendEmail(templateID, TO_EMAIL_ADDRESS, {
+      personalisation: {
+        subject:
+          type === "support"
+            ? "Support request / Demande de soutien"
+            : "Contact request / Demande de soutien",
+        formResponse:
+          type === "support"
```
Moro-Code, line 40: This seems very fragile and difficult to read. It might be worth to separate the two cases as if statements to increase readability

pages/api/request/help.ts
```diff
@@ -0,0 +1,88 @@
+import { NextApiRequest, NextApiResponse } from "next";
```
Moro-Code, line 1: Unit tests are needed for this file. Since we're looking to address tech debt as of late it might be worth it so as we don't add more

pages/api/request/help.ts
```diff
@@ -0,0 +1,88 @@
...
+${description}
+`,
+      },
+      reference: null,
+    });
+
+    return res.status(200).json({});
+  } catch (error) {
+    logMessage.error(error);
+    return res.status(500).json({ error: "Failed to send request" });
```
Moro-Code, line 84: Nit: We usually start off 500 errors with something along the lines of "Internal Service Error". Honestly we need to evaluate how responses as a whole for different error states. Just making a note of it

pages/form-builder/contactus.tsx
```diff
@@ -0,0 +1,281 @@
+import { serverSideTranslations } from "next-i18next/serverSideTranslations";
+import { GetServerSideProps } from "next";
+
```
Moro-Code, line 3: Is there a reason imports are seperated ? I assume server vs client side things ?

pages/form-builder/contactus.tsx
```diff
@@ -0,0 +1,281 @@
...
+  ErrorListItem,
+  MultipleChoiceGroup,
+  TextArea,
+  Description,
+} from "@components/forms";
+import { StyledLink } from "@components/globals/StyledLink/StyledLink";
+
+export default function Contactus() {
+  const { t, i18n } = useTranslation(["form-builder", "common"]);
+  const [errorState, setErrorState] = useState({ message: "" });
```
Moro-Code, line 25: Hmmm if this state only has one key. It's probably best for it to be a string as opposed to an object. See the following 
https://beta.reactjs.org/learn/updating-objects-in-state

pages/form-builder/contactus.tsx
```diff
@@ -0,0 +1,281 @@
...
+    }).catch((err) => {
+      logMessage.error(err);
+      setErrorState({ message: t("submissionError") });
+    });
+  };
+
+  const validationSchema = Yup.object().shape({
+    name: Yup.string()
+      .required(t("input-validation.required", { ns: "common" }))
+      .max(50, t("signUpRegistration.fields.name.error.maxLength")),
```
Moro-Code, line 53: Probably not a good idea here to use the same translation as the name field in the signUp scope. Yes its repetition but it ensures we know what translation strings affect the different parts of the app. Also I don't see this namespace in the `useTranslation` so this probably would not work

pages/form-builder/contactus.tsx
```diff
@@ -0,0 +1,281 @@
...
+  const [isSuccessScreen, setIsSuccessScreen] = useState(false);
+
+  const handleRequest = async (
+    name: string,
+    email: string,
+    request: string,
+    description: string
+  ) => {
+    const type = "contactus";
+    return await axios({
```
Moro-Code, line 36: Probably best to use try catch here

pages/form-builder/contactus.tsx
```diff
@@ -0,0 +1,281 @@
...
+  });
+
+  return (
+    <div aria-live="polite">
+      {!isSuccessScreen && (
+        <Formik
+          initialValues={{ name: "", email: "", request: "", description: "" }}
+          onSubmit={async ({ name, email, request, description }) => {
+            setSubmitting(true);
+            try {
```
Moro-Code, line 68: Not exactly sure why a try catch would be needed here. The handleRequest function already handles the error and does not bubble it up. Not sure its good practice to throw a new error to handle when a bad response is returned. Furthermore why not just do all of this in handleRequest ? Otherwise an if statement should suffice on the response status here

pages/form-builder/contactus.tsx
```diff
@@ -0,0 +1,281 @@
...
+                      );
+                    })}
+                  </ol>
+                </Alert>
+              )}
+
+              <h1>{t("contactus.title")}</h1>
+              <p className="mb-6 mt-[-2rem]">
+                {t("contactus.paragraph1")}&nbsp;
+                <Link href={`/${i18n.language}/form-builder/support`}>
```
Moro-Code, line 125: No need to actually include the language here sine you're using the Link component. Next will already have the context of the language the user has selected and will direct them to the appropriate language regardless

pages/form-builder/contactus.tsx
```diff
@@ -0,0 +1,281 @@
...
+                  {t("contactus.paragraph1Link")}
```
Moro-Code, line 126: Nit: Perhaps the naming of these translation strings can be a little more clear. `paragraph1` `paragraph2` is kind of opaque but this is minor

pages/form-builder/contactus.tsx
```diff
@@ -0,0 +1,281 @@
...
+                </Link>
+                .
```
Moro-Code, line 128: Is this . intentional ? Perhaps its better included in the translation string

pages/form-builder/contactus.tsx
```diff
@@ -0,0 +1,281 @@
...
+                      part2: t("formElements.characterCount.part2"),
+                      part1Error: t("formElements.characterCount.part1-error"),
+                      part2Error: t("formElements.characterCount.part2-error"),
+                    }}
+                  />
+                </div>
+                <div className="flex mt-14">
+                  <Button
+                    type="submit"
+                    className={`
```
Moro-Code, line 228: So we have a style for this now. I know we are revising styling but probably better to use this one rather than repeat a bunch of tailwind classes everywhere and so to keep focus and hover states somewhat consistent 

https://github.com/cds-snc/platform-forms-client/blob/develop/pages/auth/login.tsx#L152

pages/form-builder/contactus.tsx
```diff
@@ -0,0 +1,281 @@
...
+        </Formik>
+      )}
+      {isSuccessScreen && (
+        <>
+          <h1>{t("requestSuccess.title")}</h1>
+          <p className="mb-16 mt-[-2rem] font-bold">{t("requestSuccess.paragraph1")}</p>
+          <div className="mb-16">
+            <StyledLink
+              href={`/${i18n.language}/myforms`}
+              className={`
```
Moro-Code, line 251: Perhaps use the button mentioned above and extend it such that it overrides the default link styles

pages/form-builder/contactus.tsx
```diff
@@ -0,0 +1,281 @@
...
+            >
+              {t("support.backToForms")}
+            </StyledLink>
+          </div>
+          <p className="mb-8">
+            {t("requestSuccess.paragraph2Part1")}&nbsp;
+            <Link href={`https://www.canada.ca/${i18n.language}/contact.html`}>
+              {t("requestSuccess.paragraph2Link")}
+            </Link>
+            &nbsp;{t("requestSuccess.paragraph2Part2")}.
```
Moro-Code, line 265: probably best to add the period to the translated string

pages/form-builder/support.tsx
```diff
@@ -0,0 +1,286 @@
+import { serverSideTranslations } from "next-i18next/serverSideTranslations";
```
Moro-Code, line 1: This file is very similar so see comments above

timarney:
@thiessenp-cds are these links per the designs --- the links look a little odd to me.
Screen Shot 2022-11-30 at 1 33 03 PM

Moro-Code (dismissed):
Great work Peter I have some comments. I recommend we have a single directory for support and an index.ts file as a catchall 

So 

```
pages
    |
    |
    form-builder
        |
        |
        support
            |
            | 
            [[...supportType]].tsx
```

You can then use `useRouter` to get the value of supportType 

so for `/form-builder/support/contactus` the value of `supportType` would be contactus. So you can use this to either show the contactus form or the default support form if the value is anything else. 

You can see an example of how this works here https://github.com/cds-snc/platform-forms-client/blob/develop/pages/id/%5Bform%5D/%5B%5B...step%5D%5D.tsx#L34-L39

Just an idea anyways

pages/api/request/help.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (!session) return res.status(403).json({});
+
+    const { type, name, email, request, description } = req.body;
+
+    if (!name || !email || !request || !description) {
+      return res.status(404).json({ error: "Malformed request" });
+    }
+
+    const templateID = process.env.TEMPLATE_ID;
+    const notifyClient = new NotifyClient(
```
timarney, line 25: Perhaps for a follow-up PR but would be nice if this was a utility function to reduce the duplicate code.

pages/api/request/help.ts
timarney: @craigzour was there any take on this from @bryan-robitaille in the previous PR that contained and email https://github.com/cds-snc/platform-forms-client/pull/1242/files#diff-1bf840116a067de2cd60bd5404d945dd8e22a23bf079e14cde2801986b078e77R7

pages/api/request/help.ts
timarney: If this was turned into a util it would make it easier 
```
sendEmail(to, subject, message)
```

pages/api/request/help.ts
craigzour: @timarney I remember asking him if we should hide it under a Github secret. He said that it was fine to keep it in the code like that. cc @Moro-Code

pages/form-builder/contactus.tsx
timarney: Agreed on this just ""

pages/api/request/help.ts
craigzour: You need it because the session is still an optional value when retrieved from the `MiddlewareProps`. This was the reason why I wanted to introduce the `assert(session)` at some point (we had a discussion on that topic).

timarney:
Tested --- everything is working as expected.  There's some cleanup tasks in here as noted in the comments I would suggest tackling the routing one + all small cleanups and making a follow-up PR to make a utility for the Notify code as it's very repetitive.

pages/api/request/help.ts
Moro-Code: It's not about sensitivity as opposed to the path to modify. If this needs to be changed we would have to change lines of code as opposed to config. Just my opinion

thiessenp-cds:
> @thiessenp-cds are these links per the designs --- the links look a little odd to me. Screen Shot 2022-11-30 at 1 33 03 PM

Agreed but yup that is the design: https://www.figma.com/file/dbezuEglWNQ8qyO64eJJiA/v1.-Form-builder?node-id=2%3A4&t=5nvxBQTl6cAOODVp-0

thiessenp-cds:
> Great work Peter I have some comments. I recommend we have a single directory for support and an index.ts file as a catchall
> 
> So
> 
> ```
> pages
>     |
>     |
>     form-builder
>         |
>         |
>         support
>             |
>             | 
>             [[...supportType]].tsx
> ```
> 
> You can then use `useRouter` to get the value of supportType
> 
> so for `/form-builder/support/contactus` the value of `supportType` would be contactus. So you can use this to either show the contactus form or the default support form if the value is anything else.
> 
> You can see an example of how this works here https://github.com/cds-snc/platform-forms-client/blob/develop/pages/id/%5Bform%5D/%5B%5B...step%5D%5D.tsx#L34-L39
> 
> Just an idea anyways

I would really like to do something like. The design has routes as `/form-builder/support` and `form-builder/contactus`. So I don't think it would be possible to change the path, or we could change the path later and combine them

pages/api/request/help.ts
thiessenp-cds: We could add this as a sort of Tech Debt question? Then later if we decide to, move any emails to env vars?

pages/api/request/help.ts
thiessenp-cds: I agree it's not pretty :) Sure a send email utility sounds good

pages/form-builder/contactus.tsx
thiessenp-cds: Heh oh right, yes that was intentional. The link is at the end of the paragraph and I didn't think it look correct to make the period part of the link. I agree it looks odd though

pages/form-builder/contactus.tsx
thiessenp-cds: No way tailwind classes are so hot right now! :) Heh I've been putting that decision off. To either go all in with the tailwind classes or use a combination of tailwind classes with custom classes.
I think we talked about this at some point with Bryan but I forget if we decided anything?

pages/api/request/help.ts
Moro-Code: But let me see if I can solve this problem more generally. We shouldn't need a check to account for TypeScript since its not actually present in the code when it runs. We know the session will always be passed down 

I can define the following type `type WithRequired<T, K extends keyof T> = T & Required<Pick<T, K>>`  Which would allow us to use it like this 

```TS
{ session } :  WithRequired<MiddlewareProps, "session">
```

No more TS errors. I will open a PR for this

pages/api/request/help.ts
Moro-Code: Think thats probably the better idea. Follow @timarney advice

Moro-Code:
> > Great work Peter I have some comments. I recommend we have a single directory for support and an index.ts file as a catchall
> > So
> > ```
> > pages
> >     |
> >     |
> >     form-builder
> >         |
> >         |
> >         support
> >             |
> >             | 
> >             [[...supportType]].tsx
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You can then use `useRouter` to get the value of supportType
> > so for `/form-builder/support/contactus` the value of `supportType` would be contactus. So you can use this to either show the contactus form or the default support form if the value is anything else.
> > You can see an example of how this works here https://github.com/cds-snc/platform-forms-client/blob/develop/pages/id/%5Bform%5D/%5B%5B...step%5D%5D.tsx#L34-L39
> > Just an idea anyways
> 
> I would really like to do something like. The design has routes as `/form-builder/support` and `form-builder/contactus`. So I don't think it would be possible to change the path, or we could change the path later and combine them

I'm sure design might be able to change. But if not we could use route.rewrite to make the route form-builder/contactus when we land on `form-builder/support/contactus`

pages/form-builder/support.tsx
thiessenp-cds: True, almost a copy+paste. I went with this because of the routing problem above.
One solution I was think about is to use a next.config with rewrites (https://nextjs.org/docs/api-reference/next.config.js/rewrites) but that seemed like overkill for one case of code duplication

pages/api/request/help.ts
craigzour: That is nice!

pages/api/request/help.ts
craigzour: @Moro-Code Oh yes I see what you mean. I guess the .env variable would now be more appropriate since there is two different places where we have the same email address set as a const in the code.

Whether it is done now or later (new ticket required) is another question and will depend on how important we need this code to be merged.

timarney: Outside of .env vars does this project have a central spot for global type constants so at least it doesn't end up in multiple files and make it easy to update?

i.e. one place to change vs X # of files to search and change

pages/api/request/help.ts
Moro-Code: https://github.com/cds-snc/platform-forms-client/pull/1374

pages/api/request/help.ts
Moro-Code: We do have a types folder. https://github.com/cds-snc/platform-forms-client/tree/develop/lib/types in the lib folder. We could add a constants file to have all the constants there so we could import them like so 

```TS
import { CONSTANT } from "@lib/types"
```

thiessenp-cds:
> > > Great work Peter I have some comments. I recommend we have a single directory for support and an index.ts file as a catchall
> > > So
> > > ```
> > > pages
> > >     |
> > >     |
> > >     form-builder
> > >         |
> > >         |
> > >         support
> > >             |
> > >             | 
> > >             [[...supportType]].tsx
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > You can then use `useRouter` to get the value of supportType
> > > so for `/form-builder/support/contactus` the value of `supportType` would be contactus. So you can use this to either show the contactus form or the default support form if the value is anything else.
> > > You can see an example of how this works here https://github.com/cds-snc/platform-forms-client/blob/develop/pages/id/%5Bform%5D/%5B%5B...step%5D%5D.tsx#L34-L39
> > > Just an idea anyways
> > 
> > 
> > I would really like to do something like. The design has routes as `/form-builder/support` and `form-builder/contactus`. So I don't think it would be possible to change the path, or we could change the path later and combine them
> 
> I'm sure design might be able to change. But if not we could use `router.replace` to make the route form-builder/contactus when we land on `form-builder/support/contactus`

I implemented this now about four different ways, each a little different. I think having separate pages makes the most sense after looking at the pros and cons. Below are my thoughts on the two simple ways either combining into one file or separating into two files.

Combining pages into one page with dynamic routing works but has some problems
 - code validation reusable only if form structure doesn’t change (name, email, radios, desc) and it probably will change in the future
 - putting condition logic for two pages is a lot of content and makes it hard to read and debug
 - adding a router adds complexity

Separating pages has some problems
 - duplicates form validation code (may be unavoidable)
 - duplicates email logic (solved by creating a util function)

My plan is to go with a separate pages and help reduce code duplication with an email util function

lib/helpers.ts
```diff
@@ -51,6 +53,45 @@ export function extractFormData(submission: Submission): Array<string> {
   return dataCollector;
 }
 
+export interface Email {
```
thiessenp-cds, line 56: Do "we" have a strong between interface vs type?

pages/form-builder/contactus.tsx
thiessenp-cds: I saw that pattern in some commits, I think from Bryan. I liked it so I stuck with it. I don't have a strong opinion though so if you think it looks messy, no problem to combine?

pages/form-builder/contactus.tsx
thiessenp-cds: Good point. I copy pasted this code and didn't even think about it

pages/form-builder/contactus.tsx
thiessenp-cds: Oh crap I didn't even notice that, yup totally agree :)

thiessenp-cds: There is also no max length in the design so I'll remove it

thiessenp-cds:
> > > Great work Peter I have some comments. I recommend we have a single directory for support and an index.ts file as a catchall
> > > So
> > > ```
> > > pages
> > >     |
> > >     |
> > >     form-builder
> > >         |
> > >         |
> > >         support
> > >             |
> > >             | 
> > >             [[...supportType]].tsx
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > You can then use `useRouter` to get the value of supportType
> > > so for `/form-builder/support/contactus` the value of `supportType` would be contactus. So you can use this to either show the contactus form or the default support form if the value is anything else.
> > > You can see an example of how this works here https://github.com/cds-snc/platform-forms-client/blob/develop/pages/id/%5Bform%5D/%5B%5B...step%5D%5D.tsx#L34-L39
> > > Just an idea anyways
> > 
> > 
> > I would really like to do something like. The design has routes as `/form-builder/support` and `form-builder/contactus`. So I don't think it would be possible to change the path, or we could change the path later and combine them
> 
> I'm sure design might be able to change. But if not we could use `router.replace` to make the route form-builder/contactus when we land on `form-builder/support/contactus`

Design was Ok with changing the route (nice). Pages now combined

lib/helpers.ts
```diff
@@ -51,6 +53,25 @@ export function extractFormData(submission: Submission): Array<string> {
   return dataCollector;
 }
 
+// Sends an email using the Notify API
```
thiessenp-cds, line 56: @timarney I was thinking about how to unit test this function. I'm not really sure about Notify dependency and what to mock. I guess the test could throw similar errors as Notify but if Notify changes..
Do you have any ideas? 
If this isn't trivial, the sendEmail util function could be done in another task

thiessenp-cds: Actually I should probably ask questions over slack - never mind :)

timarney:
@thiessenp-cds 

I was responding to a comment but I see it's been deleted now.

Here's what I was saying about the sendEmail function

> I think it's more about the code calling sendEmail how it handles responses 

You could mock some responses.

I would suggest creating a follow up PR so we can get these pages in for further testing with the support team.

pages/api/request/help.ts
thiessenp-cds: Would this apply to a user that is not authenticated? My assumption was no

pages/api/request/help.ts
thiessenp-cds: Sounds like a good idea. Though IMO we should do this in another task and include the other related email related APIs

pages/api/request/help.ts
Moro-Code: Hey Peter. If a user is unauthenticated there would be no session and thus the sessionCheck would return a 403 error

thiessenp-cds:
> @thiessenp-cds
> 
> This link doesn't seem to be working --- takes you to the page your currently on "support" ... should go to contact us Screen Shot 2022-12-07 at 12 38 59 PM

Good find! fixed

thiessenp-cds:
> Maybe a follow-up PR for this but noting the button text here:
> 
> Screen Shot 2022-12-07 at 12 55 16 PM

I set the active and visted text states to white. At least the buttons are readable now for visited/active

components/globals/Attention/Attention.tsx
```diff
@@ -0,0 +1,38 @@
...
+}: AttentionProps & JSX.IntrinsicElements["div"]): React.ReactElement => {
+  let color = "";
+  let icon = null;
+
+  switch (type) {
+    case "warning":
+      color = "bg-amber-100";
+      icon = <WarningIcon title="Warning" width="50" />;
+      break;
+    // TODO case "info":
```
timarney, line 23: Never-mind I missed your setting some default higher up,

timarney (approved):
Approving - tested locally.

public/static/locales/en/form-builder.json
```diff
@@ -249,13 +249,6 @@
     "forOtherEnquiriesPart2": "to find the department or agency that can support you.",
     "theGCFormsTeam": "The GC Forms team provides support to people who work in the government. We cannot give general advice to the public about the forms that we host.",
     "backToForms": "Back To Forms"
-  },
```
craigzour: Not sure why this change is cannot be seen in the pull request `Files changed` section but you have somehow deleted strings that are still being used.

public/static/locales/fr/form-builder.json
```diff
@@ -249,13 +249,6 @@
     "forOtherEnquiriesPart2": "pour trouver le ministère ou l'organisme qui peut vous aider.",
     "theGCFormsTeam": "L'équipe de Formulaire GC fournit des services aux personnes qui travaillent au sein du gouvernement. Nous ne pouvons pas donner de conseils généraux au public sur les formulaires que nous hébergeons.",
     "backToForms": "Retourner au Formulaires GC"
-  },
```
craigzour: Not sure why this change is cannot be seen in the pull request `Files changed` section but you have somehow deleted strings that are still being used.

craigzour (changes requested):

public/static/locales/en/form-builder.json
craigzour: See https://github.com/cds-snc/platform-forms-client/pull/1342/commits/a267993f1a53ea331d9e4bbfb2487d80f0d14355#diff-54857a04f57e243d2aad32b79912c9e694f5869d7cf69b759ed958da0643ba96L252

public/static/locales/fr/form-builder.json
craigzour: See https://github.com/cds-snc/platform-forms-client/pull/1342/commits/a267993f1a53ea331d9e4bbfb2487d80f0d14355#diff-a6e486b021382fa7935492752ec9890787007f55c22fdc5c300ec95e11355a7bL253

craigzour:
```
"mainNavAriaLabel": "[FR] Main",

  "defaultElementDescription": {
    "email": "Entrez une adresse email comme name@example.com",
    "phone": "Entrez un numéro de téléphone comme 111-222-3333",
    "date": "Entrez une date comme mm/dd/yyyy",
    "number": "N'entrez que des nombres"
  }
```

public/static/locales/en/form-builder.json
thiessenp-cds: From what I can see, those strings are still there, though they were strangely moved

thiessenp-cds: I'll mark this as resolved since the keys are there but wow very strange

timarney:
@craigzour if you can confirm + approve (once your done reviewing)

We can create follow-ups as needed looking to get this merged so we can test with the support team.

Issue #1324, opened 2022-11-23
French characters are not registering on the "users page"


srtalbot:
Clients signing up with French characters in their name are not showing up in the ["Users" screen.](https://forms-staging.cdssandbox.xyz/admin/users)

For example, Clement's name shows up as "ClC)ment Janin"

PR #1343, opened 2022-11-28
fix: JWT token is now decoded as a UTF8 string instead of ASCII

craigzour:
# Summary | Résumé

closes https://github.com/cds-snc/platform-forms-client/issues/1324

- Fixed issue where accent in string are not properly handled because of ASCII decoding

@Moro-Code Was there a reason why you used ASCII? Did not notice any difference while testing except that accent are now properly handled.


Moro-Code:
> # Summary | Résumé
> closes #1324
> 
> * Fixed issue where accent in string are not properly handled because of ASCII decoding
> 
> @Moro-Code Was there a reason why you used ASCII? Did not notice any difference while testing except that accent are now properly handled.

I must have missed it honesty. I'm used to encoding things in ascii. Great catch and fix @craigzour

Moro-Code (approved):
Good catch

PR #1345, opened 2022-11-28
Fix tailwindcss config

dsamojlenko:
# Summary | Résumé

Some TailwindCSS sections were misconfigured, partly I believe based on an [old way of extending tailwind](https://v1.tailwindcss.com/docs/customizing-colors#overriding-or-adding-a-single-shade) in the case of colors.

The colors section should be placed in `theme.extend` block which is merged with the base colors.

Default colors (ie, `bg-white`, `bg-black`, `bg-red`) were also misconfigured with `default` vs `DEFAULT`, so to use them you had to type `bg-white-default`. By correctly using the `DEFAULT` keyword, you can now type `bg-white` `bg-red` etc.

Likewise, fontSize should be placed in theme.extend so that font sizes aren't just replaced, but extended so that original tailwindcss classes are still available (ie, text-sm, text-xs), in addition to the new ones being added.

This same issue was found with zIndex, borderWidth, and boxShadow. This PR moves all of them into the theme.extend section.


thiessenp-cds (approved):
LGTM :) Thanks for this!
Also did a quick local test and the major pages etc. look good (unchanged)

Issue #1375, opened 2022-11-30
Enforce response retrieval to only send to acceptable domains


srtalbot:
Acceptable domains:
- @*.gc.ca
- @canada.ca
- @cds-snc.ca
- @cds-snc.freshdesk.com
- @elections.ca
- @rcafinnovation.ca
- @canadacouncil.ca

### Acceptance criteria:
- [x] Review acceptable domains list above
- [x] Provide error messaging and system visibility (e.g.: let people know if they put in the wrong address), review with design. 
- [x] Ensure the ["Response delivery destination"](https://forms-staging.cdssandbox.xyz/form-builder/preview/settings) can only send to acceptable domains. 
- [ ] Update staging to be able to send to the domains above
- [ ] Provide some additional assurance that the email has been input correctly

craigzour:
@srtalbot We already have email validation function implemented. It is used on some specific features so we could use that one to validate the email we will be sending responses to.
Here is the list of domains that we are allowing in the current implementation:

`
{
    "domains": ["canada.ca",
        "gc.ca",
        "cds-snc.ca",
        "ontario.ca",
        "ontariohealth.ca",
        "novascotia.ca",
        "gov.ab.ca",
        "gouv.qc.ca",
        "rcafinnovation.ca",
        "canadacouncil.ca",
        "idrc.ca",
        "elections.ca",
        "gov.bc.ca",
        "gov.mb.ca",
        "gov.sk.ca",
        "gnb.ca",
        "gov.pe.ca",
        "gov.nl.ca",
        "gov.yk.ca",
        "gov.nt.ca",
        "gov.nu.ca",
        "nshealth.ca"] 
}`

Did you only want the 4 domains (listed in the ticket description) for that feature or can we have all the other ones? I would just need to add the `@cds-snc.freshdesk.com` to the list.

craigzour:
New GC Notify Live API key has been generated. Will now have to replace the existing one in our Github secrets and redeploy the terraform project.

craigzour:
@srtalbot 
One of the acceptance criteria is "Provide some additional assurance that the email has been input correctly". 
Did we already decide what we would like to have? Should not this be tackled in a different task?

PR #1411, opened 2022-12-07
Feature/new email validation

craigzour:
# TODO WHEN MERGING

- Replace existing STAGING_NOTIFY_API_KEY (https://github.com/cds-snc/forms-terraform/settings/secrets/actions) with the newly generated one
- Trigger new Terraform deployment
- Create new ticket to keep in mind that the new API Key (Live type) is only a temporary solution so that we can simplify testing but that it should be reverted to the previous one (Team and Safelist type)

# Summary | Résumé

closes https://github.com/cds-snc/platform-forms-client/issues/1375

- Changed government email validation regex
- Added government email validation to form builder settings page

Screen Shot 2022-12-07 at 9 26 13 AM

- Go to http://localhost:3000/form-builder/settings
- Input an invalid email address. You should see an alert popping up. (See valid domain [here](https://github.com/cds-snc/platform-forms-client/issues/1375))
- Input a valid email address. There should not be any error.

# Pull Request Checklist


lib/validation.tsx
```diff
@@ -295,23 +295,12 @@ export const setFocusOnErrorMessage = (props: FormikProps<Responses>, errorId: s
...
-  //Get the domain from email
-  const emailDomain = email.substring(email.lastIndexOf("@") + 1);
-  //Check the email's domain against the list of domains
-  //Search from the end of the string since we care about top-ish level domains most.
-  const domainList = domains.join("|");
-  const regGovDomain = new RegExp(`.?(?:${domainList})$`);
-  return regGovDomain.test(emailDomain);
+export const isValidGovEmail = (email: string): boolean => {
+  const regex =
+    /^([a-zA-Z0-9!#$%&'*+-/=?^_`{|}~.])+@((?:[a-zA-Z0-9-.]+\.gc\.ca|cds-snc\.freshdesk\.com)|(canada|cds-snc|elections|rcafinnovation|canadacouncil)\.ca)$/;
```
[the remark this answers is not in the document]
craigzour: I understand but the fact that we have to allow dynamic `*.gc.ca` subdomains as well as having that same validation in both the code and the [JSON validation file](lib/middleware/schemas/templates.schema.json) makes it more consistent to have it implemented this way.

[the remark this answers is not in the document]
thiessenp-cds: True that makes sense (insert coder sad tear :)

timarney (dismissed):
@craigzour approving as is but let's create a follow-up ticket to improve a11y 

Currently when the error pops there's no alert / dialog for screen readers

timarney:
Recorded what I see / hear ... it does pop up that text for a split second
https://user-images.githubusercontent.com/62242/206265763-8b41cda2-de55-43d9-bc89-feadfeb9c598.mp4

craigzour:
> Recorded what I see / hear ... it does pop up that text for a split second https://user-images.githubusercontent.com/62242/206265763-8b41cda2-de55-43d9-bc89-feadfeb9c598.mp4

Alright yeah I guess I was lucky when I tested it because it now starts reading the title of that error div and then just stops :(

components/form-builder/app/Settings.tsx
```diff
@@ -69,6 +70,22 @@ const HintText = ({ id, children }: { id: string; children?: JSX.Element | strin
   );
 };
 
+const InvalidEmailError = ({ isActive }: { isActive: boolean }) => {
+  const { t } = useTranslation("form-builder");
+
+  return isActive ? (
+    <div className="mt-2 mb-2" role="alert">
```
dsamojlenko, line 77: The alert area should be present on the page before it is filled with content. See this bit in the [aria alert docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role): 

> As with all other [live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions), alerts will only be announced when the content of the element with role="alert" is updated. Make sure that the element with the role is present in the page's markup first

So the outer container with `role="alert"` should be in the dom but invisible, and then populated with the message when triggered.

dsamojlenko: Also could add a `aria-invalid` attribute to the input when the error is triggered https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-invalid

components/form-builder/app/Settings.tsx
```diff
@@ -69,6 +70,22 @@ const HintText = ({ id, children }: { id: string; children?: JSX.Element | strin
...
+      <span className="text-red text-sm font-bold">{t("settingsInvalidEmailAlertTitle")}</span>
```
dsamojlenko, line 78: This should probably be a heading. I don't think there are any other headings on the page, so could be H2 and just use tailwind classes to size it appropriately.

craigzour:
PR was updated with recent commits from develop as well as some improvements for accessibility (which are still under testing).

timarney:
🥇 

Screen Shot 2022-12-07 at 3 47 05 PM

timarney:
>  For this reason, it's usually best to briefly "clear" the contents of the alert container before then injecting the alert message.

Wondering if clearing an adding a small delay might help


https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role

dsamojlenko (approved):
Nice work, thanks for incorporating the recommended a11y fixes ... we may be able to improve a bit more later.

PR #1419, opened 2022-12-09
Adds lang attribute for localization requirements

thiessenp-cds:
# Summary | Résumé

WCAG requires any text in a language not matching the parent container (probably the html tag) should override with its language.  The work mainly involved adding lang={LANG} in the form-builder code.

Test the form builder edit mode:
-set the page language to English (or reverse all steps for French)
-open a form in edit mode and toggle the translation to french
-enter some data in the fields
-open the browser console and verify the related fields have a parent element with a `lang="fr"`
Screenshot 2022-12-09 at 1 31 51 PM
Screenshot 2022-12-09 at 1 32 39 PM

Test the form builder preview mode:
-same steps as above though finding the elements is a bit tricky
Screenshot 2022-12-09 at 12 57 35 PM

Test the form builder test mode:
-same steps as above though finding the elements is a bit tricky
Screenshot 2022-12-09 at 1 00 15 PM

Adding an automated test would be nice once the cyrpress test direction is decided.

# Pull Request Checklist


components/form-builder/app/Preview.tsx
```diff
@@ -26,14 +26,19 @@ export const Preview = () => {
...
 
   return (
     <>
       <span className="bg-purple-200 p-2 inline-block mb-1">{t("page1")}</span>
-      <div className="border-3 border-dashed border-blue-focus p-4 mb-8 pointer-events-none">
+      <div
+        className="border-3 border-dashed border-blue-focus p-4 mb-8 pointer-events-none"
+        {...(i18n.language !== translationLanguagePriority && {
+          lang: translationLanguagePriority,
+        })}
```
thiessenp-cds, line 40: Would be cleaner to put the lang attribute on each form element vs. higher up in a parent element and then overriding with buttons/text not matching. The preview and translate was somewhat complex and this seemed better since it resulted in touching less code

components/form-builder/app/edit/Edit.tsx
```diff
@@ -72,7 +72,13 @@ export const Edit = () => {
       <RichTextLocked
         beforeContent={
           <>
-            <label htmlFor="formTitle" className="visually-hidden">
+            <label
+              htmlFor="formTitle"
+              className="visually-hidden"
+              {...(i18n.language !== translationLanguagePriority && {
```
timarney, line 78: Something that might be worth trying 


Setup a store helper something like:

```
useTemplateStore
...
getLocalizationLang: () => get().lang !== get().translationLanguagePriority ? translationLanguagePriority : "",
```

lang in the store should be in sync with i18n.language and this would avoid having the same check in multiple spots if we need to update.

thiessenp-cds: Makes sense. Were you thinking something like placing a function like this for re-use at the top of files like Edit.tsx or more a util like function that would go in something like useTemplateStore.tsx? 
The second, useTemplateStore, I might need some guidance. I skimmed it and most files in the Form Builder but that one I think understand but not 100% :)

timarney: A helper in useTemplateStore.tsx --- that we can pull in via useTemplateStore

timarney: A helper in useTemplateStore.tsx --- that we can pull in via useTemplateStore

timarney: Thinking something like https://github.com/cds-snc/platform-forms-client/commit/bde632bf72a3430abad122cf00086d8ed45a3b32

thiessenp-cds: Looks good, I'll add/merge that

components/form-builder/app/edit/Edit.tsx
```diff
@@ -77,7 +79,13 @@ export const Edit = () => {
       <RichTextLocked
         beforeContent={
           <>
-            <label htmlFor="formTitle" className="visually-hidden">
+            <label
```
timarney, line 82: This can be replaced with ` {...getLocalizationAttribute()}` ya?

components/form-builder/app/TestDataDelivery.tsx
```diff
@@ -94,7 +104,12 @@ export const TestDataDelivery = () => {
         )}
         <li>{t("fillFormClickSubmit")}</li>
       </ol>
-      <div className="border-3 border-dashed border-blue-focus p-4 mb-8">
+      <div
+        className="border-3 border-dashed border-blue-focus p-4 mb-8"
```
timarney, line 108: Can replace this with `getLocalizationAttribute:`

timarney (approved):

PR #1447, opened 2022-12-19
Misc updates for WCAG conformance

thiessenp-cds:
# Summary | Résumé

This began as fixing a label and grew into a few more accessibility related fixes.

- invisible label reference in the form-build upload button
- Form Edit and Preview Language Switcher to announce updates
- Main nav to use a list for screen reader user conveniences

Navigate to the form-builder start page and verify the JSON upload works as expected. Then try a WCAG automated test tool and verify there are no easy to find errors.

Navigate to the form-builder edit or preview screen and try switching the form language by mouse or keyboard and verify it works as expected. Try this again with a screen reader and try switching the form language. The new language should be announced and focus stay on the "switch-like" button.

Navigate to any gc forms page. With a screen reader on, try navigating the main list by keyboard. Each list item should be announced along with its position in the list etc.

The "switch-like" button and form-build layout uses absolute and similar positioning some places. In the future this could be made more fluid using flex or grids.

# Pull Request Checklist


components/form-builder/app/Start.tsx
```diff
@@ -112,22 +112,24 @@ export const Start = () => {
           <h2 className="p-0 mb-1 group-hover:underline group-focus:underline">{t("startH2")}</h2>
           <p className="text-sm">{t("startP1")}</p>
         </button>
-        <label
```
thiessenp-cds, line 115: The problem was the invisible input referencing a visible label. The fix is just nesting the input which arguably is still referencing a visible input but... it will now past WCAG automated test tools.
Also making a label interactive will make some but not all WCAG automated test tools grumpy. The fix is just to make a div interactive instead.
Oh and the ARIA spec requires that button be can be keyboard activated by both ENTER and SPACE keys

components/form-builder/app/navigation/EditNavigation.tsx
```diff
@@ -20,32 +20,51 @@ export const EditNavigation = () => {
   };
 
   return (
-    <div className="relative">
```
thiessenp-cds, line 23: Because the design does not fit a traditional component like a Switch or Radio or.. the below required some umm.. creativity. The main problem this solves is having the button state announce when it changes. 
This could be moved into a component in the future but I tried to change as little as possible.

components/form-builder/app/navigation/EditNavigation.tsx
```diff
@@ -20,32 +20,51 @@ export const EditNavigation = () => {
...
-          {translationLanguagePriority === "fr" && (
-            <button id="lang" onClick={() => switchLang()}>
-              <ToggleRight className="inline-block w-12 fill-fuchsia-300" />
-            </button>
-          )}
-          {translationLanguagePriority === "en" && (
-            <button id="lang" onClick={() => switchLang()}>
+          </span>
+          <div
+            role="button spinbutton"
```
thiessenp-cds, line 34: This is a hack :) So...... a button cannot use the aria-activedescendant property which is used to announce state changes. However aria allows multiple roles to be assigned to an element. A browser will read and probably only pay attention to the first role, current behavior. The second problem involving automated WCAG test tools, the second role will be picked up by automated tools (or just confuse it with two roles) and allow the aria-activedescendant on a spinbutton role. So it works but it-is-not-pretty :)

components/form-builder/app/shared/LangSwitcher.tsx
```diff
@@ -0,0 +1,70 @@
...
+    <div className="flex items-baseline">
+      <span
+        id="switch-heading"
+        className="font-bold text-sm whitespace-nowrap mr-2"
+        aria-hidden="true"
+      >
+        {t(descriptionLangKey)}
+      </span>
+      <div
+        role="button spinbutton"
```
thiessenp-cds, line 34: Because the design does not fit a traditional component like a Switch or Radio or.. the this required some creativity. The main problem this solves is having the button state announce when it changes through aria-activedescendant. The role=button gives the semantics of a button (which this is close to) but a button should not have an aria-activedescendent. To get around this I added a second role that allows this attribute. It's not pretty but it works.

timarney: I'm might do doing something wrong but I don't hear anything announced when toggling between languages using VoiceOver --- perhaps some other readers pick this up better?

timarney (approved):

PR #1453, opened 2022-12-20
fix: switch getOrCreate user function to use email instead of token sub

Moro-Code:
# Summary | Résumé

No issue 

It was previously possible to create an account with cognito and gsuite that used the same email. Redirect logic has been added such that user's can only sign in with the original way they registered (The way they signed into GC Forms the first time). I.E if one registered with the cognito method and attempted to sign in with a gsuite account with the same email they would be redirected to sign in with the cognito account and vice versa. 

This removes the possibility of an account takeover and as such we can return to using emails to lookup users. This will allow us to use the csv option to migrate users from one cognito user pool to another and have user's still be able to access their accounts after a password reset

1. Test to make sure the login still works 

N/A

# Pull Request Checklist

### DIFF
--- a/lib/users.ts
+++ b/lib/users.ts
@@ -9,12 +9,12 @@ import { MongoAbility } from "@casl/ability";
  * Get or Create a user if a record does not exist
  * @returns A User Object
  */
-export const getOrCreateUser = async ({ sub, name, email, picture }: DefaultJWT) => {
-  if (!sub) throw new Error("Sub does not exist on token");
+export const getOrCreateUser = async ({ name, email, picture }: DefaultJWT) => {
+  if (!email) throw new Error("Sub does not exist on token");
   const user = await prisma.user
     .findUnique({
       where: {
-        id: sub,
+        email: email,
       },
       select: {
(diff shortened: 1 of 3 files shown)


lib/users.ts
```diff
@@ -9,12 +9,12 @@ import { MongoAbility } from "@casl/ability";
  * Get or Create a user if a record does not exist
  * @returns A User Object
  */
-export const getOrCreateUser = async ({ sub, name, email, picture }: DefaultJWT) => {
-  if (!sub) throw new Error("Sub does not exist on token");
+export const getOrCreateUser = async ({ name, email, picture }: DefaultJWT) => {
+  if (!email) throw new Error("Sub does not exist on token");
```
bryan-robitaille, line 13: We may want to change the text in the Error thrown as well to `"Email does not exist on token"`

bryan-robitaille (changes requested):
One quick change to the text and of an error and we should be good to go.

bryan-robitaille (approved):
👍  LGTM

Issue #400, opened 2021-08-27
Browser version incompatibility for end-users


srtalbot:
As an end-user
I need to know what my next steps are
When I come to GC forms with a browser version or don't have javascript enabled
So that I can choose an alternative path to continue my journey, or take steps to update my version / settings

### Context:
[See Slack thread for usage at Accessibility Standards Canada](https://gcdigital.slack.com/archives/CAG3WKMC4/p1649856420926099)

[Designs](https://www.figma.com/file/XZsVxBlweLrmys0AvwYzg1/Platform-CORE-team-design-work?node-id=0%3A1)

### Acceptance criteria:
- [x] End-users know when their browser does not meet the requirements
- [x] End-users are provided an alternative path to continue their service request
- [x] Pages are WCAG 2.1 AA compliant
- [ ] Pages can be navigated by a screen reader

srtalbot:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/GCForms-60cb8929764d71000e481cab/issues/cds-snc/platform-forms-client/400?planning-poker) with ZenHub @omartehsin1

srtalbot:
On chrome it seems to work, but on IE there seems to be some issues. Needs to be retested to see if it's working.

PR #1487, opened 2023-01-04
Feature/browser incompatibility update

falila:
# Summary | Résumé

FYI: a special shout-out to Omar Tehsin who has done the majority of the  work @omartehsin1 [ here](https://github.com/cds-snc/platform-forms-client/pull/831)

Additional  fixes in this PR: 
- removal of the nested middleware not allow anymore in Nextjs 12.2
- typing and props mismatch
- the use of useragent helper instead of ua object from request which was deprecated. 
- brand not found rendering issue 

Figma [ here](https://www.figma.com/file/XZsVxBlweLrmys0AvwYzg1/Platform-CORE-team-design-work?node-id=930%3A1899&t=pEdxSFDBj24MVkDA-0)

###  Testing JS disable 

Chrome

1. Click the three dots menu in the top right corner.
2. Choose “Settings”
3. On the left-side menu click “Privacy and security”
6. Choose “Site settings”
7. Scroll to the bottom to the “Content” section and click “JavaScript”
8. Choose the option “Don't allow sites to use JavaScript”

IE11

1. Open browser > select Tools 
2.  in drop-down menu
3.  select Internet options > Security.
4. In the Security level for this zone section, choose Custom level > locate Scripting section.
5. Locate Active scripting subhead 
6. choose Disable-ON > OK.

Refer to[ One](https://github.com/cds-snc/platform-forms-client/pull/831#issue-1272412367) and [two](https://github.com/cds-snc/platform-forms-client/issues/400#issuecomment-1182002406)

## Testing Not supported browser and branding .

How to Test  a form with branding.

For Chrome/ IE11 incompatibility, tests can be done using IE11 or even in _middleware.ts, changing the conditional to `url.pathName === (form path)` to a form with branding and you will be redirected."

OR simply paste the url of a form into IE11 and you should be redirected on not-supported page.
If really want to mock IE11 use virtualbox or Chrome browser replace `Internet Explorer` in middleware.ts by `Chrome` before pasting the url. 

# Known issues

1. Css issue while rendering not-supported browser IE11 page.  

# Pull Request Checklist

- [X] Have you modified the change log and updated any relevant documentation?

### DIFF
--- a/components/form-builder/app/edit/ElementPanel.tsx
+++ b/components/form-builder/app/edit/ElementPanel.tsx
@@ -5,6 +5,7 @@ import { useTemplateStore } from "../../store";
 import { PanelActions, PanelBodyRoot, MoreModal } from "./index";
 import { isValidatedTextType } from "../../util";
 import { FormElementTypes } from "@lib/types";
+import { useIsWithin } from "@components/form-builder/hooks/useIsWithin";
 
 export const ElementPanel = ({ item }: { item: FormElementWithIndex }) => {
   const {
@@ -66,13 +67,19 @@ export const ElementPanel = ({ item }: { item: FormElementWithIndex }) => {
     [add, setFocusInput, updateField]
   );
 
+  const { ref, focusWithinProps, isWithin } = useIsWithin();
+
   return (
     <div
-      key={lang}
-      className={`element-${item.index} ${className} border border-black max-w-[800px] h-auto -mt-1`}
+      {...focusWithinProps}
+      ref={ref}
+      className={`element-${item.index} ${className} group ${
+        isWithin ? "active" : ""
+      } hover:bg-violet-50 focus:bg-violet-50 border border-t-0 border-black max-w-[800px] h-auto relative`}
     >
       <PanelBodyRoot item={item} />
       <PanelActions
+        subIndex={-1}
         elements={elements}
         lang={lang}
         item={item}
--- a/components/form-builder/app/edit/PanelActions.tsx
+++ b/components/form-builder/app/edit/PanelActions.tsx
@@ -1,5 +1,6 @@
-import React, { useEffect, useRef } from "react";
+import React, { useEffect, useRef, useState, useCallback } from "react";
 import { useTranslation } from "next-i18next";
+import { useMediaQuery } from "usehooks-ts";
 
 import { FormElementTypes, FormElement } from "@lib/types";
 import { Button } from "../shared";
@@ -11,14 +12,17 @@ import {
   Close,
   Duplicate,
   ThreeDotsIcon,
+  AddIcon,
 } from "@components/form-builder/icons";
 
 import { usePanelActions } from "@components/form-builder/hooks";
+import { ElementDialog } from "./elements/element-dialog/ElementDialog";
+import { useFlag } from "@lib/hooks";
 
 const buttonClasses =
-  "group border-none transition duration-100 h-0 !py-5 lg:!pb-3 !pl-4 !pr-2 m-1 !bg-transparent hover:!bg-gray-600 focus:!bg-blue-hover disabled:!bg-transparent";
+  "group border-none transition duration-100 h-0 !py-5 lg:!pb-3 !pl-4 !pr-2 m-1 !bg-transparent xl:hover:!bg-gray-600 focus:!bg-blue-hover focus:text-black xl:focus:text-white disabled:!bg-transparent";
 const iconClasses =
-  "group-hover:group-enabled:fill-white-default group-disabled:fill-gray-500 group-focus:fill-white-default transition duration-100";
+  "group-hover:group-enabled:fill-black group-disabled:fill-gray-500 group-focus:fill-white transition duration-100";
 
 export interface RenderMoreFunc {
   ({ item, moreButton }: { item: FormElementWithIndex; moreButton: JSX.Element | undefined }):
@@ -29,8 +33,7 @@ export interface RenderMoreFunc {
 
 export const PanelActions = ({
   item,
-  lang,
-  elements,
+  subIndex,
   renderMoreButton,
   handleAdd,
   handleRemove,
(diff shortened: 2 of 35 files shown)


craigzour:
@falila While the main logic is working as intended. I feel like there is probably some improvements that we can bring to that JS not enabled page design. cc @connorscarolyns @srtalbot 
Screen Shot 2023-01-05 at 11 12 10 AM

pages/_document.tsx
```diff
@@ -1,6 +1,8 @@
 import Document, { Html, Head, Main, NextScript, OriginProps } from "next/document";
 import React from "react";
 import { googleTagManager, cspHashOf } from "@lib/cspScripts";
+import JSDisabled from "./js-disabled";
+const noJsScss = `#__next {display:none;}`;
```
craigzour, line 5: This is only used on line 222 so I wonder if we should move it there instead of creating a `const` and then using it later on.

pages/_document.tsx
```diff
@@ -216,18 +218,14 @@ class MyDocument extends Document {
           <script async type="text/javascript" src="/static/scripts/form-polyfills.js"></script>
           {GoogleTagScript}
         </CustomHead>
+        <noscript>
+          <style type="text/css">{noJsScss}</style>
+          <JSDisabled
+            formProps={this.props.__NEXT_DATA__.props.pageProps.formRecord}
```
craigzour, line 224: It looks like the `JSDisabled` component only uses a logo out of the formRecord object. What if we rename `formProps` into `brandLogo` and only pass the relevant information?

pages/_document.tsx
```diff
@@ -216,18 +218,14 @@ class MyDocument extends Document {
...
+            lang={this.props.__NEXT_DATA__.locale}
```
craigzour, line 225: If the JS not enabled page will always display content if both languages should not we remove the `lang` prop?

pages/js-disabled.tsx
```diff
@@ -0,0 +1,123 @@
...
+            style={{
+              flexDirection: lang == "en" ? "column" : "column-reverse",
+              height: "80vh",
+            }}
+          >
+            <div id="en">
+              <div className="flex justify-between items-center my-10">
+                <div className="xxs:w-flag-fold xs:w-flag-5s md:w-44 w-flag-desktop">
+                  {brandImg?.brand ? (
+                    <img src={brandImg?.brand?.logoEn} alt={brandImg?.brand?.logoTitleEn} />
```
craigzour, line 33: To avoid having eslint warnings you could add `/* eslint-disable @next/next/no-img-element */` at the top of the file with a comment explaining why you had to use `` instead of `<Image>`.

pages/js-disabled.tsx
```diff
@@ -0,0 +1,123 @@
...
+              </p>
+            </div>
+          </div>
+        </div>
+      </main>
+      <Footer formData={formProps} lang={lang} />
+    </>
+  );
+};
+const Footer = ({ formData, lang }: { formData: PublicFormRecord; lang: string | undefined }) => {
```
craigzour, line 83: Do we really need a footer on this page? The link to the Figma file is not valid anymore. Could you update it?

pages/not-supported.tsx
```diff
@@ -0,0 +1,185 @@
...
+  return (
+    <footer className="lg:mt-10 border-0 mt-16">
+      <div className="border-t-4 border-gray-900 py-10 flex flex-col px-64 lg:px-16 xl:px-32 sm:py-2 sm:px-6">
+        <div className="md:mb-10 flex self-start xxl:flex self-start">
+          <a href={t("footer.terms.link")}>{t("footer.terms.desc")}</a>
+        </div>
+
+        <div className="flex self-end sm:flex sm:self-start">
+          {formProps?.brand ? (
+            <Image
```
craigzour, line 96: Won't we have the same issue with `<Image>` here?

pages/not-supported.tsx
```diff
@@ -0,0 +1,185 @@
...
+            </div>
+          </div>
+        </div>
+      </main>
+      <Footer formData={formInfo} lang={lang} />
+    </>
+  );
+};
+
+const Footer = ({ formData, lang }: { formData: PublicFormRecord; lang: string }) => {
```
craigzour, line 84: Looks like the same Footer that was used in `js-disabled.tsx`. Maybe you could extract it in a different file in order to reuse it. I think there is a Layout folder somewhere in the project.

craigzour (changes requested):
Have a few questions and suggestions for you. This is mainly related to the design of that page. I wonder if the design is still valid or if it should be reviewed by the design team.

pages/js-disabled.tsx
craigzour: I get this when I click on that link:
Screen Shot 2023-01-06 at 8 07 28 AM

craigzour (changes requested):
We are almost there I think! The comments I made last time have been addressed. The only thing that is not correct is the design of the page when compared to what is in Figma. Also it appears that the footer is not properly connected (see screenshot below).

Here is what I see on Chrome:
Screen Shot 2023-02-02 at 8 52 33 AM

craigzour (dismissed):

craigzour:
@falila Looks like something is broken because I always get the `You need JavaScript` page even though my JS is enabled.

craigzour:
I can confirm that it is now working as expected (with your [add missing tag](https://github.com/cds-snc/platform-forms-client/pull/1487/commits/54e17efe07e3092af553ad039e6de90dceb3570f) fix). 
I also tested the fact that the CSS is loaded in production mode (https://nextjs.org/docs/basic-features/built-in-css-support#does-it-work-with-javascript-disabled).

Design still looks like:
Screen Shot 2023-02-14 at 2 27 21 PM
I would like your approval on this @connorscarolyns 

@falila I will let @bryan-robitaille also do a review before approving because I know he was waiting for your last commit.

connorscarolyns:
This looks good to me now.

PR #1493, opened 2023-01-05
Fix Form Timer

bryan-robitaille:
# Summary | Résumé

Main purpose of this PR is to refactor and fix glitches with the Form Timer.

- The Submit Button has been refactored into it's own component and the associated hooks required are now contained within it instead of the InnerForm.
- A countdown now appears instead of a static "try again in x seconds" to provide more dynamic feedback to the user.
- The Form Timer does not reset after trying to submit the form.  
- The Loader spinner is only shown after form validation has passed and the form is being sent to the server.

|Stops a user from submitting | A11y |
| ----- | ------- |
|form_timer_fixed|button_submit_a11y|

 The `isSubmitting` value from Formik that was previously used to switch between the Loader and Form component was causing issues with state management.  `isSubmitting` is set to true at the beginning of the Form Submission cycle which includes Pre-Submission, Validation, and then Submission.  If a form contained error the Loader would flash on the screen and then errors would appear as the Form component was rerendered.  I've switched to using the `status` state variable on Formik that instead only shows the loader when the form is being handled in `handleSubmit` and sending to the server.

References: 
- [Form Submission Cycle](https://formik.org/docs/guides/form-submission)
- [setSubmitting is still required](https://formik.org/docs/guides/form-submission)  

Secondary Fixes:
- Removes non-existant "buttons" class from components
- Removed duplicate useTranslation hook, namespaces can be added as an Array
- Switched useFlag SWR (stale-while-revalidate) to use Axios instead of Fetch to align with other application network calls.
- Enabled and modified form-builder Cypress test for latest design.

- Try to submit a form immediately
- Fix any required field errors
- Submit again

# Pull Request Checklist


craigzour (approved):
Nice work :) Testing went well and it looks good to me!

Issue #1473, opened 2022-12-28
Richtext editor bugs


srtalbot:
### Bugs and usability fixes

- [x] Bug: When you add a link to the rich text, the pop-up link doesn't clear unless you click within the box. If you click elsewhere - like on the page outside of the richtext border - the pop-up persists. 
- [x] Usability: If you copy text from a rich text box that has a translation already, and move it to the introduction box, it does not keep the french translation. 
- [x] Focus state in the "Translate" screens kept jumping to the bottom of a rich text field when I added a link or typed. 

https://images.zenhubusercontent.com/315708034/8ccc5e21-bb4e-4e2d-9dc9-7b66935022f0/screen_recording_2022_12_21_at_2_51_27_pm.mov
- [x] Rich text editor adds an extra space after an "enter"

https://images.zenhubusercontent.com/315708034/af9c6b67-a697-4bcd-bd12-b4660fb43b53/screen_recording_2022_12_22_at_8_27_14_am.mov

- [ ] Text editor makes all text a heading level instead of just highlighted text and adds a space.

https://images.zenhubusercontent.com/315708034/b21be690-6869-4958-a4bf-619eca12b548/screen_recording_2022_12_22_at_10_18_15_am.mov

- [x] Link modal jumps to the top of the screen when deleting content that comes after the link. 

https://images.zenhubusercontent.com/315708034/ccb4a988-3617-499e-9d0f-41e7f896cb71/screen_recording_2022_12_22_at_10_21_35_am.mov

- [ ] When adding a line at the beginning of a link, it moves the entire rich text content down one line.

https://images.zenhubusercontent.com/315708034/9d42760d-3a8c-42aa-9a40-8717bbd3d36d/screen_recording_2022_12_28_at_11_05_17_am.mov

- [x] Focus issue when pressing enter on the translation page (file won't upload, message me on Slack)

- [ ] Extra bullet added to a line above when pressing enter at the end of an unordered list.

https://images.zenhubusercontent.com/315708034/430c8243-5864-4745-8b62-612b8f9b624d/screen_recording_2022_12_30_at_8_23_25_am.mov

dsamojlenko:
I think these two items:
- Text editor makes all text a heading level instead of just highlighted text and adds a space.
- When adding a line at the beginning of a link, it moves the entire rich text content down one line.

Are caused by the same underlying issue. Discovered this issue and PR in the Lexical repo that seem to be addressing same:
[...]

PR #1498, opened 2023-01-06
RichTextEditor bug fixes

dsamojlenko:
# Summary | Résumé

Includes fixes for a few items in #1473 Richtext editor bugs

### Hide floating link editor when editor loses focus
The latest updates to Lexical included some code that allows you to dismiss the Floating Link Editor by pressing the ESC key. In addition, this PR will dismiss the Floating Link Editor when the editor loses focus.

### Add two trailing spaces to previous lines for line breaks
The Markdown serializer for Lexical does not add two trailing spaces to the line before a line break, which our markdown renderer `markdown-to-jsx` requires. This meant that line breaks entered in the RichTextEditor were not rendered when displaying the form.

This fix was a bit brute-force. We were unable to patch the Serializer directly, so instead we're just massaging the markdown that comes back from the serializer, adding two spaces to the end of any line that precedes another line of text. This has the side effect of adding some extra double spaces to the end of list items unnecessarily, but it has no discernible effect on those items.

### Other items
- Small refactor of RichText and RichTextLocked for readability
- Bugfix in RichText addresses issue where content was always being saved to English
- Added some cypress regression tests for the bugfix
- Added a basic cypress test for the RichTextEditor LinkEditor (more tk)
- Refactored cypress commands into a folder for better organization
- Added a text selection cypress command


components/form-builder/app/edit/elements/lexical-editor/Editor.tsx
```diff
@@ -88,7 +88,18 @@ export const Editor = ({
             editorState.read(() => {
               // Read the contents of the EditorState here.
               const markdown = $convertToMarkdownString(TRANSFORMERS);
-              onChange(markdown);
+
+              // Add two spaces to previous line for linebreaks (this is not handled properly by $convertToMarkdownString)
+              const lines = markdown.split("\n");
+              lines.map((currentLine, i) => {
```
timarney, line 94: Thinking we should be using a forEach here + updates lines to use `let`

timarney (approved):

PR #1509, opened 2023-01-09
feat: Docker vulnerability scan and generate SBOM

patheard:
# Summary
Add a nightly Docker vulnerability scan and update the generate SBOM action to only create an SBOM for the Docker image.

GitHub is now automatically indexing and scanning our manifests so the Dependency-Track SBOM generate workflow is no longer needed.

# Related
- cds-snc/platform-core-services#204


bryan-robitaille (approved):
LGTM 👍

PR #1560, opened 2023-01-18
Update templates-related API routes

dsamojlenko:
# Summary | Résumé

The API route to request a single Template does not work as expected. It is a GET request that is looking for the formID in the body of the request, but GET requests do not have a body.

So when we were attempting to retrieve a single Template, the endpoint was falling back to retrieving all Templates, and then we were filtering the array client side.

This also meant that the test coverage for this endpoint should have been failing but was returning a false positive.

This PR refactors the `/api/tempates` routes to be more RESTful by adding a resource ID to the URL for resource-specific routes. The templates resource now looks like this:
- `GET /api/templates` retrieve all templates
- `POST /api/templates` create a new template
- `GET /api/templates/[formID]` retrieve a single template
- `PUT /api/templates/[formID]` update a template
- `DELETE /api/templates/[formID]` delete a template

In addition, all calls to the `/api/templates` endpoints were reviewed and updated as necessary, along with any related tests.

You can test with PostMan or another REST api client, or you can test by navigating around the site inspecting the Network requests if you know all the places the endpoints are used.

To test with a REST API client, you'll need to login to the app through a browser and copy the next session cookie to your client.

I can provide an Insomnia or Postman collection for testing if desired.

# Pull Request Checklist


__tests__/api/templates.test.ts
```diff
@@ -5,7 +5,8 @@
 /* eslint-disable  @typescript-eslint/no-explicit-any */
 import { createMocks, RequestMethod } from "node-mocks-http";
 import Redis from "ioredis-mock";
-import templates from "@pages/api/templates";
+import templates from "@pages/api/templates/[...formID]";
+import { default as templatesRoot } from "@pages/api/templates/index";
```
bryan-robitaille, line 9: For readability we could use the same pattern as the other default imports with aliases:  `import templatesRoot from "@pages/api/templates/index"`

pages/api/templates/[...formID].ts
```diff
@@ -94,27 +84,23 @@ const templates = async (
...
   ability: MongoAbility;
   method: string;
-  user: Session["user"];
-  formID?: string;
+  request: NextApiRequest;
   formConfig?: BetterOmit<FormRecord, "id" | "bearerToken">;
   isPublished: boolean;
   users: { id: string; action: "add" | "remove" }[];
 }) => {
+  const formID = request.query.formID ? (request.query.formID[0] as string) : undefined;
```
Comment (bryan-robitaille) [inline pages/api/templates/[...formID].ts:99]: If we change the route type from catch all to dynamic `[...formID]` to `[formID]` we can implicity type the `query.formID` as string.

bryan-robitaille (changes requested):
Nice Work!
The only suggestion I might make is to swap the catch all route of `[...formID]` for a pure dynamic route of `[formID]`.  This will ensure only a single formID can be passed to the API endpoint and any attempt to pass more should result in a 404 page.

PR #1614, opened 2023-02-13
Add unauthenticated Share flow

dsamojlenko:
# Summary | Résumé

Adds the Share flow for unauthenticated users. Similar to the authenticated ShareDropDown/ShareModal, except we don't send the email through notify. Instead, we provide a downloadable JSON file, and copyable instructions.

Note that the instructions are copied as plain text for flexibility.

image

image

- Create a form as an unauthenticated user
- Use the ShareDropdown to select "Share by email"
- Follow the instructions


timarney:
Looks good - we should add links + a bit of space between the paragraphs.

Screen Shot 2023-02-13 at 12 58 10 PM

timarney (approved):

PR #1634, opened 2023-02-16
Feature/retrieval download modal UI

thiessenp-cds:
# Summary | Résumé

Adds a new Dialog for "Confirm receipt of responses" and "Report problems with responses".

The work adds a new component LineItemEntries which borrows heavily from the TagInput component. The first implementation used the TagInput component but it was not possible to match a Tag with the design for "Report problems". I then tried to add a bunch more options to Tags but the code was becoming complex and when behavior started diverge I went with a new but similar component.
The unit tests are very similar between LineItemEntries and TagInput but as more behavior is added, they definitely diverge and make more sense to separate.

Figma Design:
https://www.figma.com/file/6QAmUEq6d8tyXzWDGKO78s/v2.-Form-builder?node-id=2889%3A12570&t=T3ryFt9ECHV2v49l-0

| N/A | d1 |
| N/A | d2 |

Open a Form, e.g. from the MyForms page, chose a Template, from the Menu, select "Settings". Click the left nav "Responses". Click the top nav button "Confirm receipt". The "Confirm receipt of responses" dialog should open. It should match the Figma design. Only UUIDs should be able to be entered. Try adding and removing entries. Escape or "close" or "Cancel" should close the dialog. Note, the "Confirm receipt" button currently does nothing.

Do the same for the "Report problems with responses" dialog. Click the "Report Problems" nav button. etc.

- On “submitting” data in a dialog, a notification should show success/fail to the user
- Bonus, focus the input when a dialog opens
- Add API calls to the dialogs and handle errors etc.
- Add end-to-end tests

# Pull Request Checklist


components/form-builder/app/shared/Dialog.tsx
```diff
@@ -56,16 +58,14 @@ export const Dialog = ({
   }, [close]);
 
   return (
-    <dialog
-      className="p-0 m-0 bg-clip-padding w-full h-full bg-transparent"
-      aria-labelledby="modal-title"
```
thiessenp-cds, line 61: Removed this intentionally. I don't think this ID exists. Below an H2 is used if a title is present

lib/validation.tsx
```diff
@@ -358,3 +358,33 @@ export const isSymbol = (field: string): boolean => {
...
+
+/**
+ * This function tests whether a string contains a UUID
+ * @param field A string containing a UUID
+ * @returns {boolean} The validation result
+ */
+export const isUUID = (field: string): boolean => {
+  // Regex from: https://github.com/microsoft/uuid-validate/blob/06554db1b093aa6bb429156fa8964e1cde2b750c/index.js#L2
+  const reg = new RegExp(
+    "^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[0-9a-f]{4}-[0-9a-f]{12}$",
```
craigzour, line 370: The Regex I have used to validate the payload that will be passed to the API is `"^[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}$"`.
I noticed that the one you have uses a limitation for letters between A and F which appears to be correct. We could change the one I have in file `uuid-array.schema.json` as part of this pull request.

However I am not sure about the middle section of your expression `[0-9a-f]{4}-[1-5][0-9a-f]{3}-[0-9a-f]{4}`. It may be correct for UUID v5 but in our system we are using UUID v4.

craigzour: @thiessenp-cds Let me know what you think and if you want me to deal with the change in `uuid-array.schema.json `.

thiessenp-cds: Oh good point, thanks for pointing that out! I just assumed the latest UUID version was used. In that case, probably best to not use this microsoft schema.  I'll update the validation.tsx to use the same UUID schema in uuid-array.schema.json

craigzour: You may want to keep the limitation in range of letters though since it appears to be a valid one. From `A to Z` to `A to F`. You will have to also apply that change in the `uuid-array.schema.json` file

thiessenp-cds: Sounds good. Done
I also pointed the form id/name regex to point to the schema validation one - better to have the regex in one place

craigzour: Perfect! Thank you :)

PR #1635, opened 2023-02-16
Add woff and woff2 fonts with priority

bryan-robitaille:
# Summary | Résumé
Adds Woff and Woff2 fonts which are roughly 10x small file size than TTF files.  The TTF fonts are kept as fallback however Woff and Woff2 are supported by most modern browsers.

Priority listing in app.scss is Woff2 -> Woff -> TTF.


craigzour (changes requested):
Two files are missing:
- NotoSans-Regular.woff
- NotoSans-Bold.woff

Otherwise it does work well!

craigzour (approved):

thiessenp-cds (approved):
LGTM - nice! :)

PR #1639, opened 2023-02-20
Cypress: Check for H1 before starting test

bryan-robitaille:
# Summary | Résumé
This should hopefully fix some of failing E2E tests on github.  The A11y test for the Registration page often failed as the test sometimes started while the spinning loader was still displaying.

This new check for the A11y tests ensures that an H1 is in the DOM before starting the A11y check.


craigzour (approved):
LGTM :)

PR #1642, opened 2023-02-20
Proof of concept - Design system in the tailwind config

amazingphilippe:
# Summary | Résumé

This PR is a draft to introduce tokens from the design system. This PR is one way we could do it:

1. Introduce tokens in the main tailwind config file. 
2. Load tokens from the [gcds-tokens npm package](https://www.npmjs.com/package/@cdssnc/gcds-tokens)
3. Reworking the _forms.scss partial. I made it into several partials, we could not do that, or split partials differently depending on the preference. Had some difficulty moving things around because of `url()` paths for images. 
5. Swapping Lato Regular for Lato Bold. And adding a comma so that the ttf font loads. 

| before | after|

You can test this PR by looking at a form preview, or a published form

# Unresolved questions | Questions non résolues

- New partials for `_forms.scss` : Ok || Not ok || Not like that
- I noticed both lato and noto sans fonts are subsetted to the following languages:  Irish, Italian, Norwegian Bokmål, Swahili and Zulu. Is this on purpose?
- We can make use of "component" tokens more intentional. With the trade-off that we won't be able to easily use tailwind fallback values. Right now its still a bit all over the place. Notice some `gc-` components make use of `--gcds-component-token` values, while also using some base tokens from the tailwind config. 

# Todo
- Bug bash. To thoroughly test the impact of this PR. 

# Pull Request Checklist


lib/formBuilder.tsx
```diff
@@ -21,6 +21,7 @@ import {
   Response,
 } from "@lib/types";
 import { TFunction } from "next-i18next";
+import "@cdssnc/gcds-tokens/build/web/variables.css";
```
amazingphilippe, line 24: Is this the best place to load tokens? Made sense to me the most.

styles/forms/_typography.scss
```diff
@@ -0,0 +1,48 @@
...
+.gc-description {
+  @apply md:text-small_sm mb-3 text-sm w-3/4 xl:w-full;
+}
+
+.gc-richText {
+  p {
+    @extend .gc-p;
+  }
+
+  // Why only h4?
```
amazingphilippe, line 16: Forgot to delete this comment.

styles/app.scss
```diff
@@ -15,10 +15,10 @@
   @font-face {
     font-family: "Lato";
     font-style: normal;
```
bryan-robitaille, line 17: If we're changing the font-weight  from 400 -> 700 I'm wondering if we should also update the font-style to `bold`.  If we want `bold` to the default for this style I wonder if that's something we can achieve through the tailwind config?  I bring it up because I can see a scenario in the future where a developer is trying to style text like a design and can't figure out why their explicit `normal` font class declaration keeps coming out bold.

bryan-robitaille: In a nutshell should be leave both the 400 and 700 weight fonts declared but set the 700 as default for text.

thiessenp-cds (commented):
I went through myforms, form-builder,  and public forms. Looks good to me. The only change I really noticed was buttons - nice that they are more consistent now. 
Also, the token decision makes sense IMO. Maybe one day in the future when we have more consistent components/design/* we can come back to this.

timarney:
Edges don't look right for "active" in form builder left nav

Screen Shot 2023-02-24 at 1 32 09 PM

timarney:
Is this the correct colour set for the error state when focused?  For lack of a better way of putting it -- it's looking pretty 🇺🇸  

Screen Shot 2023-02-24 at 1 31 26 PM

timarney:
Just me or does the submit / button feel really small

Screen Shot 2023-02-24 at 1 42 35 PM

Maybe I'm just used to the older style but seems off to me.

timarney:
I haven't dug into performance research for CSS variables but at least slightly concerning the # of variables being loaded vs the actual usage.  I imagine this will grow over time.

Again maybe not impactful but would be worth looking into a bit.

Screen Shot 2023-02-24 at 2 11 32 PM

styles/app.scss
amazingphilippe: Ah I see what you mean! [font-style](https://developer.mozilla.org/fr/docs/Web/CSS/font-style) is for normal/italic/oblique. Usually, we should set a font-face to match the font file.

amazingphilippe:
> Edges don't look right for "active" in form builder left nav

This is probably because I don't scope things out enough. tailwind themes might prevent this. 

> error state when focused

This is a gap I shared with the DS team. I simply applied the regular focus styles to the error summary. I agree there must be a better way to design this!

> submit / button feel really small

I find it small too. Maybe the font size or padding values are not properly mapped. Will look into it.

> performance research for CSS variables

good question. My guess would be that unused CSS has no performance impact on rendering and recalculation since they are kept at the root level, other than loading them. I'd love to be able to prune unused variables, it would feel very nice. 
Found this: https://lisilinhart.info/posts/css-variables-performance/ 
and also this: https://blog.okopress.dev/css-variables-performance-benchmark-2021-f56cd26ae9f5

styles/forms/_theme.scss
```diff
@@ -0,0 +1,39 @@
+[data-theme="gcds"] {
```
amazingphilippe, line 1: Option A is to map each token to a theme variable.

timarney:
Closing as this is falling out of date and isn't in active development.  Can re-open in the future.

PR #1718, opened 2023-03-09
Discovery - GCDS components in form builder script. Also Next 13

amazingphilippe:
# Summary | Résumé

This PR show the issues we get when trying to use components in the form builder script.

the form builder script is a big "switch" that simply check through a list of form input definitions, and generates form fields. 

Using the GCDS components in here, rather than in components, would in theory work best, as the gcds components are less modular. Example: `<GcdsInput>` contains, the label, description, validator and the input. The equivalent `TextInput`component in GC Forms does not contain the label and description.

Also there seems to be some incompatible types. Commented out in this PR. 

Here is how the component renders. I was able to briefly make the styles work, but now they seem broken. More investigation needed!

Capture d’écran, le 2023-03-09 à 13 02 43

### DIFF
--- a/lib/formBuilder.tsx
+++ b/lib/formBuilder.tsx
@@ -10,7 +10,6 @@ import {
   MultipleChoiceGroup,
   RichText,
   TextArea,
-  TextInput,
 } from "@components/forms";
 import { GcdsInput } from "@cdssnc/gcds-components-react";
 import {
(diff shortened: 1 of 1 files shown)


amazingphilippe:
@timarney Another PR on top of your next 13 work.

timarney:
Closing as this is falling out of date and isn't in active development.  Can re-open in the future.

PR #1742, opened 2023-03-14
Fix DynamoDB connector for localstack setup

bryan-robitaille:
# Summary | Résumé
PR fixes an error that crept in during the Audit Log feature work.


craigzour (dismissed):
I am guessing the previous implementation was reliant on a specific local setup because it worked on my machine when I tested the audit log feature.

thiessenp-cds (dismissed):
LGTM - thanks!

craigzour (approved):

Issue #1026, opened 2022-09-14
Retrieval UI - Data Retention Policy Enforcement - Warning/Alert notifications (NextJS server side detection)


craigzour:
**Dependencies**

- Response table feature

**User story**

As a system:
- I should detect how many form responses are over 21 days old (will trigger warning notification)
- I should detect how many form responses are over 35 days old (will trigger alert notification)

**Story context**

Feature: https://docs.google.com/document/d/1RUgysJFq3I_8gEbUE-3Wrb6BqsRcLKElfYyRWHjQ_9s
Implementation documentation for the data retention policy enforcement subset: https://docs.google.com/document/d/1LI4bhaTa710AD8YpCGTTXHi3nQPuKE03YX7LGdexibY

**Suggested solution**

- https://docs.google.com/document/d/1LI4bhaTa710AD8YpCGTTXHi3nQPuKE03YX7LGdexibY/edit#heading=h.j1jztlb4x1bj

**Acceptance criteria**

- [ ] Server side detection logic conveys the information (number of responses over 21 and 35 days old) to the UI side

Issue #1027, opened 2022-09-14
Retrieval UI - Data Retention Policy Enforcement - Warning/Alert notifications (UI)


craigzour:
**Dependencies**

- https://github.com/cds-snc/platform-forms-client/pull/1698
- https://github.com/cds-snc/platform-forms-client/issues/1026

**User story**

As a program administrator:
- I can not select responses without the "Overdue" flag if the alert notification is displayed
- I can click on buttons located inside notifications (see Figma design) to auto select "Overdue" responses

**Story context**

Feature: https://docs.google.com/document/d/1RUgysJFq3I_8gEbUE-3Wrb6BqsRcLKElfYyRWHjQ_9s
Implementation documentation for the data retention policy enforcement subset: https://docs.google.com/document/d/1LI4bhaTa710AD8YpCGTTXHi3nQPuKE03YX7LGdexibY

**Suggested solution**

- https://docs.google.com/document/d/1LI4bhaTa710AD8YpCGTTXHi3nQPuKE03YX7LGdexibY/edit#heading=h.dmpihd1m21hh
- https://docs.google.com/document/d/1LI4bhaTa710AD8YpCGTTXHi3nQPuKE03YX7LGdexibY/edit#heading=h.j1jztlb4x1bj
- https://docs.google.com/document/d/1LI4bhaTa710AD8YpCGTTXHi3nQPuKE03YX7LGdexibY/edit#heading=h.xv6bnu2nzco0

**Acceptance criteria**

- [ ] Only "Overdue" responses can be selected when an alert notification is displayed
- [ ] Add new buttons in Nagware notifications (located on top of the response table)
- [ ] Buttons in notification can be interacted with and automatically select "Overdue" responses

PR #1772, opened 2023-03-23
Feature/nagware server side detection

craigzour:
# Summary | Résumé

closes https://github.com/cds-snc/platform-forms-client/issues/1026

- Fixed issue where `Date` type was used to store `number`
- Added Nagware detect function to determine if the response table should display a warning or alert notification when form submissions are going over 21 days old
- Added initial version of visual nagware notifications on top of the response table

Screen Shot 2023-03-27 at 9 06 49 AM
Screen Shot 2023-03-27 at 9 06 38 AM
Screen Shot 2023-03-27 at 9 06 24 AM
Screen Shot 2023-03-27 at 9 06 04 AM

- Add following line `return { level: NagLevel.UnsavedSubmissionsOver21DaysOld, numberOfSubmissions: 22 };` to function `detectOldUnprocessedSubmissions` in `lib/nagware.ts` file. It will allow you to decide which scenario you want to test by forcing a specific notification to appear on the screen
- Navigate to the response table of any form you have and check whether the notification is displayed or not

- There is still a little bit of work to do to fully complete the Nagware (client side) feature. See https://github.com/cds-snc/platform-forms-client/issues/1027

# Pull Request Checklist


components/form-builder/app/Nagware.tsx
```diff
@@ -0,0 +1,106 @@
+import React from "react";
+import { useTranslation } from "next-i18next";
+import { Attention } from "@components/globals/Attention/Attention";
+import { NagLevel, NagwareResult } from "@lib/nagware";
```
bryan-robitaille, line 4: Be wary of importing `@lib/*` into tsx components.  It has happened in the past where a server side module was later added as an import into a `@lib/*` module which caused errors as it's dependencides could not run on the browser or vice versa.  Might suggest moving the enum and types for Nagware to a `@types` to be safe.

components/form-builder/app/Nagware.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+export const Nagware = ({ nagwareResult }: { nagwareResult: NagwareResult }) => {
+  const { t } = useTranslation("form-builder");
+
+  if (nagwareResult.level === NagLevel.None) return <></>;
+
+  const generateNotificationContent = (): {
```
bryan-robitaille, line 11: I'd suggest maybe using a [useMemo](https://react.dev/reference/react/useMemo) function here to memoize the `generateNotificationContent`.  This way the function isn't recalculated on every render and will only recalculate is one of the listed dependencies change.

components/form-builder/app/Nagware.tsx
```diff
@@ -0,0 +1,106 @@
...
+        };
+      case NagLevel.None:
+        throw new Error("Case should not be reached");
+    }
+  };
+
+  const notificationContent = generateNotificationContent();
+
+  return (
+    <>
```
bryan-robitaille, line 100: The React fragments `<> </>` are not needed as the return of `<Attention>` is already a single parent element.

bryan-robitaille (changes requested):
First off... AWESOME TESTS!
Great work!  A few minor suggestions and we should be good.
Thinking we may also want to include a new flag to not show the Nag message until the whole feature is complete?

lib/nagware.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export type NagwareResult = {
+  level: NagLevel;
+  numberOfSubmissions: number;
+};
+
+export function detectOldUnprocessedSubmissions(submissions: NagwareSubmission[]): NagwareResult {
+  const results = submissions.reduce(
+    (acc, current) => {
+      const diffMs = Math.abs(Date.now() - current.createdAt);
```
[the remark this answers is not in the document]
craigzour: Oh yeah definitely!

components/form-builder/app/Nagware.tsx
craigzour: Oh ok I see. So this is the or one of the reason why we try to always put types in a specific type file instead of keeping them close to where they originate from?

bryan-robitaille: Exactly.

bryan-robitaille (approved):
Great Work!

PR #1879, opened 2023-04-06
Fix/response table sorting

### LINKED ISSUE #1865: Response page table has responses in the incorrect order
Tested on: Tablet, MS Edge

New responses should be at the top, above confirmed, older responses. The confirmed response is older and should be below the new response.

thiessenp-cds:
Quick fix that adds sorting to the responses download table. 
Expected behaviour here:
https://app.zenhub.com/workspaces/gcforms-60cb8929764d71000e481cab/issues/gh/cds-snc/platform-forms-client/1865

closes #1865


craigzour (approved):
LGTM!

PR #1940, opened 2023-04-13
Listen for changes to Template through hook

bryan-robitaille:
# Summary | Résumé
A new hook `useSubscibeToTemplateStore` can provide a listener for changes to a selection of store states.  Effectively allowing for creating unique listening slices for side effects.

In this use case the TemplateContext uses an `isDirty` ref which is set by listening to state changes for `form, isPublished, name, deliveryOption, securityAttribute`.

This helps reduce calls to the server and database by ensuring Templates have changed before saving.

An `onClick` handler for the leftNav checks to see if the form needs saving before navigating away ensuring that a user's work is not lost.

- Make modifications to a Template in Form Builder
- Browser console should identify that template is out of sync with server
- Navigate using left nav
- Broswer console should identify that template is saved to server
- Navigate back to edit but do not change the form
- Navigate using left nav
- Browser console should not show new messages that template is out of sync of saving occured. 

# Pull Request Checklist


components/form-builder/app/navigation/LeftNavLink.tsx
```diff
@@ -17,9 +34,11 @@ export const LeftNavLink = ({ children, href }: { children: ReactElement; href:
     }
   }, [activePathname, isReady, href, setActive]);
   return (
-    <Link href={href}>
+    <Link href={href} passHref>
+      {/* passHref does ensure the child a tag has the correct HREF attribute */}
       <a
-        href={href}
+        href={"will_be_replaced_by_nextjs"}
+        onClick={() => saveOnNavigate()}
```
timarney, line 41:
```suggestion
        onClick={saveOnNavigate}
```

components/form-builder/app/navigation/LeftNavLink.tsx
```diff
@@ -1,10 +1,27 @@
...
   const [active, setActive] = useState(false);
   const { isReady, activePathname } = useActivePathname();
+  const { saveForm } = useTemplateContext();
+  const { setId, getIsPublished } = useTemplateStore((s) => ({
+    setId: s.setId,
+    getIsPublished: s.getIsPublished,
+  }));
+
+  const saveOnNavigate = useCallback(async () => {
+    if (!getIsPublished()) {
```
timarney, line 17: Wondering if we should move more logic into saveForm

Potentially it could handle the `if` check for `getIsPublished` and also handling setting the id as needed.

components/form-builder/store/useTemplateStore.tsx
```diff
@@ -432,6 +435,15 @@ export const useTemplateStore = <T,>(
   return useStore(store, selector, equalityFn);
 };
 
+export const useSubscibeToTemplateStore = <T,>(
+  selector: (state: TemplateStoreState) => T,
+  listener: (selectedState: T, previousSelectedState: T) => void
+) => {
+  const store = useContext(TemplateStoreContext);
+  if (!store) throw new Error("Missing Template Store Provider in tree");
+  useEffect(() => store.subscribe(selector, listener), [store, selector, listener]);
```
timarney, line 444: Is the `store` dep here the store from the const (line 442)?

bryan-robitaille: It is.  I included it for the time being because of the `exhaustive-deps` hooks rule but haven't had the time to test yet whether or not state changes to the store fire the UseEffect.

components/form-builder/app/navigation/LeftNavLink.tsx
timarney: > allow to save only parts of the template that are allowed to be modified after publishing

nice ya 

>  Thinking that in the near future we would allow users to change delivery option if the form is already published.

👍

timarney (approved):

Issue #1934, opened 2023-04-13
Classification of HTML responses should match selected classification


anikbrazeau:
# Issue/opportunity

Currently classification on HTML response data is Protected B, should match the classification selected: Protected A or Unclassified.

# Screenshots

![PROTECTED B (1)](https://user-images.githubusercontent.com/38330843/231802526-16a73c88-e4e3-41b7-be1a-d6a30c36f977.PNG)

<img width="835" alt="Screenshot 2023-04-13 at 11 05 26 AM" src="https://user-images.githubusercontent.com/38330843/231802802-de975075-9214-458a-b0e4-a756ac559920.png">

# Test case

- 4.01 (Build + share a form)
- Surface Tablet
- Microsoft Edge
- In French

PR #1946, opened 2023-04-14
fix: data classification in HTML response does not match form settings

craigzour:
# Summary | Résumé

closes https://github.com/cds-snc/platform-forms-client/issues/1934

- Fixed issue where data classification was not properly set in the HTML response. It was static but now relies on the `SecurityAttribute` retrieved from the database.

### DIFF
--- a/components/myforms/HTMLDownload/ProtectedWarning.tsx
+++ b/components/myforms/HTMLDownload/ProtectedWarning.tsx
@@ -3,20 +3,35 @@ import { useTranslation } from "next-i18next";
 
 // Note: use lang prop if you want to force a specific language e.g. HTMLDownload component
 interface ProtectedWarningProps {
+  securityAttribute: string;
   lang?: string;
 }
 
 export const ProtectedWarning = (props: ProtectedWarningProps) => {
   const { t, i18n } = useTranslation(["my-forms"]);
   const { lang = i18n.language } = props;
+
+  const renderSecurityAttribute = () => {
+    switch (props.securityAttribute) {
+      case "Unclassified":
+        return t("dataClassification.unclassified", { lng: lang });
+      case "Protected A":
+        return t("dataClassification.protectedA", { lng: lang });
+      case "Protected B":
+        return t("dataClassification.protectedB", { lng: lang });
+      default:
+        throw new Error(`Unsupported security attribute ${props.securityAttribute}`);
+    }
+  };
+
   return (
     <div className="flex p-4 justify-between items-center bg-gray-200 border border-gray-300 text-sm">
       <div className="flex flex-col">
-        <p className="font-bold">{t("protectedBWarning.officialRecord", { lng: lang })}</p>
-        <p>{t("protectedBWarning.retainCopy", { lng: lang })}</p>
+        <p className="font-bold">{t("dataClassification.officialRecord", { lng: lang })}</p>
+        <p>{t("dataClassification.retainCopy", { lng: lang })}</p>
       </div>
       <div className="px-4 py-2 bg-white-default font-bold border border-gray-300">
-        {t("protectedBWarning.protectedB", { lng: lang })}
+        {renderSecurityAttribute()}
       </div>
     </div>
   );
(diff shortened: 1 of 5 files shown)


components/myforms/HTMLDownload/index.tsx
```diff
@@ -14,6 +14,7 @@ interface HTMLDownloadProps {
   // submissionID: string;
   responseID: string;
   createdAt: number;
+  securityAttribute: string;
```
timarney, line 17: We have a type for this now 

form-types

```
export type SecurityAttribute = "Unclassified" | "Protected A" | "Protected B";
```

bryan-robitaille (approved):

components/myforms/HTMLDownload/ProtectedWarning.tsx
```diff
@@ -3,20 +3,35 @@ import { useTranslation } from "next-i18next";
...
+  securityAttribute: string;
   lang?: string;
 }
 
 export const ProtectedWarning = (props: ProtectedWarningProps) => {
   const { t, i18n } = useTranslation(["my-forms"]);
   const { lang = i18n.language } = props;
+
+  const renderSecurityAttribute = () => {
+    switch (props.securityAttribute) {
```
timarney, line 15: Something for now or maybe a future PR

In SetResponseDelivery.tsx we have:

```
const classificationOptions = [
  { value: "Unclassified", en: "Unclassified", fr: "Non classifié" },
  { value: "Protected A", en: "Protected A", fr: "Protégé A" },
] as const;
```

So we should figure out how we want to implement so we don't duplicate / have one place to update as need be.

pages/api/id/[form]/[submission]/download.tsx
```diff
@@ -61,23 +61,33 @@ const handler = async (req: NextApiRequest, res: NextApiResponse, props: Middlew
...
+      securityAttribute,
+    } = ((
       vaultResult
     ): {
       formResponse: Responses;
       createdAt: number;
       confirmReceiptCode: string;
       responseID: string;
       submissionID: string;
+      securityAttribute: string;
```
Comment (timarney) [inline pages/api/id/[form]/[submission]/download.tsx:90]: https://github.com/cds-snc/platform-forms-client/pull/1946/files#r1167144926

public/static/locales/en/my-forms.json
```diff
@@ -58,9 +58,11 @@
     "submissionID": "Submission ID",
     "copyResponse": "Copy answer row to clipboard"
   },
-  "protectedBWarning": {
+  "dataClassification": {
     "officialRecord": "This is an official record",
     "retainCopy": "Retain this copy in accordance to your retention policy.",
+    "unclassified": "UNCLASSIFIED",
```
timarney, line 64: Should these be all uppercase?

We use `Protected A` etc... on other spots

craigzour: In the Figma file we can see that the data classification label is written in full uppercase

pages/api/id/[form]/[submission]/download.tsx
Comment (craigzour) [reply pages/api/id/[form]/[submission]/download.tsx:90]: Will change

components/myforms/HTMLDownload/index.tsx
craigzour: Will change

thiessenp-cds (commented):
From testing locally, I can see the default is correctly set to unclassified (in French and English). I'm not sure how to test for protected A/B but the logic looks right to me. I see some other comments/requests here so I'll leave this open.
But from my end LGTM :)

timarney:
@thiessenp-cds 

> From testing locally, I can see the default is correctly set to unclassified (in French and English). I'm not sure how to test for protected A/B but the logic looks right to me. 

1) The default is now Protected A
2) you can update to Protected A or Unclassified via Settings (Protected B isn't available)

Screen Shot 2023-04-14 at 2 36 23 PM

components/myforms/HTMLDownload/ProtectedWarning.tsx
timarney: Ya - we can make another PR to update classificationOptions to use the lang files.

timarney (approved):

Issue #1933, opened 2023-04-13
Spacing within bulleted list in Markdown


anikbrazeau:
# Issue/opportunity

Spacing in "Know your responsibilities" / "Service-level agreement" / "Terms of use" / "Terms and conditions" 

Spacing seems wonky within bullets of different hierarchies -- should appear as a cohesive list.

Spacing seems excessive between paragraphs or before bullets generally.

These do not appear with a space in the Markdown.

# Screenshots

<img width="1653" alt="Screenshot 2023-04-13 at 10 58 58 AM" src="https://user-images.githubusercontent.com/38330843/231800913-d8c3d400-64a4-4828-8b88-f34a97fc195d.png">

PR #1949, opened 2023-04-14
fix: removed space in bulleted list for Markdown rendered pages

craigzour:
# Summary | Résumé

closes https://github.com/cds-snc/platform-forms-client/issues/1933

- Removed space in bulleted list for Markdown renderer content

### DIFF
--- a/styles/_forms.scss
+++ b/styles/_forms.scss
@@ -95,6 +95,7 @@
 .gc-richText {
   p {
     @extend .gc-p;
+    @apply mb-0;
   }
   h4 {
     @extend .gc-section-header;
--- a/styles/_forms.scss
+++ b/styles/_forms.scss
@@ -105,6 +105,14 @@
     @apply mb-10;
   }
 
+  ul,
+  ol {
+    ul,
+    ol {
+      @apply mb-0;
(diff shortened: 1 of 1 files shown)


bryan-robitaille (approved):
I don't see too much of an impact across the site as this only impacts a child list of a parent list.

bryan-robitaille (changes requested):
This really wrecks havoc with the content.
Screenshot 2023-04-14 at 3 45 45 PM

What exactly are we trying to achieve / fix?

craigzour:
> This really wrecks havoc with the content. Screenshot 2023-04-14 at 3 45 45 PM
> 
> What exactly are we trying to achieve / fix?

Yes I just noticed it while testing. I will try to find another solution but the goal was to remove the space between the paragraph and the bulleted list
Screenshot 2023-04-14 at 3 34 44 PM

craigzour:
@bryan-robitaille Found a solution using a CSS selector https://caniuse.com/?search=has() . It may not work on all browsers but according to the website 86% of global users have a browser that supports it.

bryan-robitaille (approved):

PR #1960, opened 2023-04-18
Fix / yarn install all dependencies in mono-repo

bryan-robitaille:
Implicitly install all packages in mono-repo style sub directories for github actions.


craigzour (approved):
LGTM :)

PR #1962, opened 2023-04-18
fix: draft/publish button in view-templates admin page is not working because of API implementation

craigzour:
# Summary | Résumé

- Fixes issue with `isPublished` API not able to unpublish a form

### DIFF
--- a/pages/api/templates/[formID].ts
+++ b/pages/api/templates/[formID].ts
@@ -103,7 +103,7 @@ const templateCRUD = async ({
           deliveryOption: deliveryOption,
           securityAttribute: securityAttribute,
         });
-      } else if (formID && isPublished) {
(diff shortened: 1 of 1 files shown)


bryan-robitaille (commented):
What are your thoughts on adding an ability check in `updateIsPublishedForTemplate` so that only a user with Manage All Forms privilege can set it to 'unpublished'?

craigzour:
> What are your thoughts on adding an ability check in `updateIsPublishedForTemplate` so that only a user with Manage All Forms privilege can set it to 'unpublished'?

Good idea! As we are not allowing users (clients) to unpublish a form in order to edit it I guess it makes sense to add an admin type of permission check when the API is called to unpublish.

bryan-robitaille (approved):
👍

PR #1966, opened 2023-04-19
feat: removed logger function

craigzour:
# Summary | Résumé

- Removed `logger` function. It was something we included in our Tech Debt document a while ago.

### DIFF
--- a/lib/formBuilder.tsx
+++ b/lib/formBuilder.tsx
@@ -1,5 +1,5 @@
 import React, { ReactElement } from "react";
-import { logger, logMessage } from "@lib/logger";
+import { logMessage } from "@lib/logger";
 import {
   Description,
   Dropdown,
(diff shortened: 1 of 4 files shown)


lib/formBuilder.tsx
```diff
@@ -308,7 +308,7 @@ const _getElementInitialValue = (element: FormElement, language: string): Respon
  * @param formRecord
  * @param language
  */
-const _getFormInitialValues = (formRecord: PublicFormRecord, language: string): Responses => {
+export const getFormInitialValues = (formRecord: PublicFormRecord, language: string): Responses => {
```
timarney, line 311: There's a comment that references the old name here 

https://github.com/cds-snc/platform-forms-client/blob/06739009234cd6152f166df1f24b6496f653948e/lib/formBuilder.tsx#L273

craigzour: Updated

timarney (approved):

Issue #1978, opened 2023-04-24
Nagware failure when processing responses associated to deleted forms


craigzour:
The current scenario is: A user can delete a form without confirming their responses. Because there are unconfirmed responses the nagware still fires but is unable to associate an owner to the response.

I think we have two options:
1. Stop a user from deleting a form if they have unconfirmed responses
2. Warn a user when they delete a form that they will be unable to retrieve responses. Then, confirm and delete the response for them.

samsadasivan:
Next steps: Discipline input and review cc: @srtalbot @connorscarolyns @craigzour @timarney @sarahhobson

timarney:
> Stop a user from deleting a form if they have unconfirmed responses

@craigzour --- outside of stopping them at a UI level should we also handle this at the API level and handle the error inside the current UI?

--

@samsadasivan I think we should look at My Form as the `entry point` for deletions.  It's currently the only way to delete. 

The app doesn't match Figma at the moment --- in a previous iteration delete from my forms redirected to the settings page.  The code was later updated to handle deletes directly from my forms.  I haven't got back around to re-introducing deleting from the settings page.

The above does make me wonder if we need the forms management page under settings --- it's all duplicated actions from My Form .   My understanding was My Form was for forms management.

craigzour:
> > Stop a user from deleting a form if they have unconfirmed responses
> 
> @craigzour --- outside of stopping them at a UI level should we also handle this at the API level?
> 
> --
> 
> @samsadasivan I think we should look at My Form as the `entry point` for deletions. It's currently the only way to delete.
> 
> The app doesn't match Figma at the moment --- in a previous iteration delete from my forms redirected to the settings page. The code was later updated to handle deletes directly from my forms. I haven't got back around to re-introducing deleting from the settings page.
> 
> The above does make me wonder if we need the forms management page under settings --- it's all duplicated actions from My Form . My understanding was My Form was for forms management.

@timarney Yes I think we should double check at the REST API or `lib/templates.ts` API because we may want to allow users to delete forms using the API at some point in the future. Also if we ever introduce a bug at the UI level we will have a second barrier.

I can start working on that API check while you tackle the UI part.

samsadasivan:
@timarney that makes sense. If we need to add another spot to find a way to delete the form from user insights, we can revisit this, but I like the idea that the builder interface is for editing and downloading. Then All Forms is for Form Management. Separating the workflows will be helpful for future features. 

[...]

PR #1998, opened 2023-04-26
Feature/delete form api check for unprocessed responses

craigzour:
# Summary | Résumé

part of https://github.com/cds-snc/platform-forms-client/issues/1978

- Added check to `deleteTemplate` function in order to make sure there is no unprocessed (New or Downloaded) submissions associated to the targeted form.

- Deleting a form that still has unprocessed responses (New or Downloaded) should be blocked at the API level (until new UI flow is implemented)

# Pull Request Checklist


timarney:
Working on the front-end here https://github.com/cds-snc/platform-forms-client/pull/1999 .

thiessenp-cds (dismissed):
Clean and easy to follow - nice! LGTM :)

Here's a quick visual of this working for anyone else that may help
Screenshot 2023-04-26 at 4 28 07 PM
Screenshot 2023-04-26 at 4 28 46 PM

bryan-robitaille (commented):
I'm a little worried about having an API endpoint `unprocessed` that can create so much traffic on the DynamoDB.  I'm wondering if there's a different pattern we can use here to inform the front end that a form has responses left.

thiessenp-cds (approved):
LGTM! :)

craigzour:
> I'm a little worried about having an API endpoint `unprocessed` that can create so much traffic on the DynamoDB. I'm wondering if there's a different pattern we can use here to inform the front end that a form has responses left.

I understand your concern. The API will only be called when the user goes to the settings panel so in a regular use case it should not happen too often.
If you really think we should review our approach then the other solution would be to make the delete call and deal with the `TemplateHasUnprocessedSubmissions` afterwards but that would change the user experience for sure.
Do you want me to share your concern with the design team?

timarney:
> The API will only be called when the user goes to the settings 

I'm only calling the endpoint when the user goes to settings + opens the modal (call is only when the modal opens)

bryan-robitaille:
My concern is more from a security perspective from outside the normal user flow.  The API endpoint accepts a GET request so it is not limited by the Rate Limiting rules we have in place.  A malicious user could create 1000's of test responses and then hit the API endpoint to cause significant traffic on the database.
I'm thinking that to keep the user experience the same but to mitigate the security issue we could create a new value in the cache for the number of form responses.  The value would be updated when `listAllSubmissions` is run but the cache checked first by `numberOfUnprocessedSubmissions`

timarney:
> My concern is more from a security perspective 

Should there be an ability check on that route - thinking there should be.  That might help limit some outside calls ... not sure if we can cache at the ability / user access level.

craigzour:
> > My concern is more from a security perspective
> 
> Should there be an ability check on that route - thinking there should be. That might help limit some outside calls ... not sure if we can cache at the ability / user access level.

We already only allow this API call to users that are logged in but even then we could still have malicious users 😕 . Let's discuss about it at the end of our stand-up

pages/api/templates/[formID].ts
```diff
@@ -58,10 +59,15 @@ const templates = async (
     // If not GET then we're authenticated and can safely return the complete Form Record.
     return res.status(200).json(response);
   } catch (err) {
-    if (err instanceof AccessControlError) return res.status(403).json({ error: "Forbidden" });
-    else if (err instanceof TemplateAlreadyPublishedError)
+    if (err instanceof AccessControlError) {
+      return res.status(403).json({ error: "Forbidden" });
+    } else if (err instanceof TemplateAlreadyPublishedError) {
       return res.status(500).json({ error: "Can't update published form" });
```
Comment (bryan-robitaille) [inline pages/api/templates/[formID].ts:65]: We need to be careful when using http error code 500. Our alarms look for this code to sense if anything is critically wrong with the application.
As a team we should try to ensure we stick with the appropriate http status response requests depending on what we're trying to convey.  
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

Comment (bryan-robitaille) [reply pages/api/templates/[formID].ts:65]: For example `Can't update published form` should probably be returned as a `409`

pages/api/templates/[formID].ts
```diff
@@ -58,10 +59,15 @@ const templates = async (
...
-    else return res.status(500).json({ error: "Malformed API Request" });
+    } else if (err instanceof TemplateHasUnprocessedSubmissions) {
+      return res.status(500).json({ error: "Found unprocessed submissions" });
```
Comment (bryan-robitaille) [inline pages/api/templates/[formID].ts:67]: Could possibly use http error code 405 (Method not allowed)

pages/api/templates/[formID].ts
```diff
@@ -58,10 +59,15 @@ const templates = async (
...
+    } else {
+      return res.status(500).json({ error: "Malformed API Request" });
```
Comment (bryan-robitaille) [inline pages/api/templates/[formID].ts:69]: Should probably be http error code 400 (Bad Request)

lib/vault.ts
```diff
@@ -135,15 +139,48 @@ export async function listAllSubmissions(
...
     logEvent(
       ability.userID,
       { type: "Response" },
       "ListResponses",
       `List all responses for form ${formID}`
     );
-    return accumulatedResponses;
+
+    const numOfUnprocessedSubmissions = accumulatedResponses.filter((submission) =>
+      [VaultStatus.NEW, VaultStatus.DOWNLOADED].includes(submission.status)
```
bryan-robitaille, line 151: VaultStatus.PROBLEM should also be included in this list.  All problem responses should be resolved before deleting the form as well.

lib/tests/vault.test.ts
```diff
@@ -0,0 +1,173 @@
...
+          Status: "Downloaded",
+        },
+      ],
+    };
+
+    ddbMock.on(QueryCommand).resolves(dynamodbMockedReponses);
+
+    const numOfUnprocessedSubmissions = await numberOfUnprocessedSubmissions(ability, "formtestID");
+
+    expect(numOfUnprocessedSubmissions).toEqual(3);
```
bryan-robitaille, line 101: If we include 'Problem' status this should be '4'

lib/tests/templates.test.ts
```diff
@@ -779,6 +789,37 @@ describe("Template CRUD functions", () => {
...
+
+      mockNumberOfUnprocessedSubmissions.mockResolvedValueOnce(1);
+
+      await expect(async () => {
+        await deleteTemplate(ability, "formtestID");
+      }).rejects.toThrowError(new TemplateHasUnprocessedSubmissions());
+
+      mockNumberOfUnprocessedSubmissions.mockResolvedValueOnce(1);
+
+      await expect(async () => {
```
bryan-robitaille, line 817: Possible duplicate of the previous assertion?

lib/cache/unprocessedSubmissionsCache.ts
```diff
@@ -0,0 +1,42 @@
+import { logMessage } from "@lib/logger";
+import { getRedisInstance } from "../integration/redisConnector";
+
+// Return a random number between 180 and 300  (3 minutes <-> 5 minutes)
+const randomCacheExpiry = () => Math.floor(Math.random() * 180 + 120);
+
+export const numberOfUnprocessedSubmissionsCacheCheck = async (
+  formID: string
+): Promise<number | null> => {
+  const checkParameter = `unprocessed-submissions:${formID}`;
```
bryan-robitaille, line 10: Most of our cache keys start with the object name; form, auth, setting.  What do you think about renaming the key to `form:submissions:unprocessed:${formID}`?  This leaves room in case in the future we want to continue to add values to the cache like `form:submissions:all:${formID}` or `form:submissions:processed:${formID}`, etc.

bryan-robitaille (changes requested):
A couple of potentially small quick tweaks and I think the server side is done :)

pages/api/templates/[formID].ts
Comment (craigzour) [reply pages/api/templates/[formID].ts:65]: I was trying to find a code that would make sense but I think I ended up on a website that was not giving that many options. I will use 409!

lib/cache/unprocessedSubmissionsCache.ts
craigzour: Yep that works!

bryan-robitaille (approved):
Nice work!

PR #2005, opened 2023-04-28
Fix/add overlay for menu

thiessenp-cds:
# Summary | Résumé

Adds an overlay component to help with components like the Menu component. It allows an overlay on top of the UI to act as a page wide click area. For the Menu component this helps by giving an area outside the Menu that when clicked will close the menu.

Test by going to the myforms screen. Try opening a menu. Move your mouse around the screen, only the menu links should change the cursor. Clicking outside the menu should close the menu.

### DIFF
--- a/components/globals/Overlay/Overlay.cy.tsx
+++ b/components/globals/Overlay/Overlay.cy.tsx
@@ -1,0 +1,27 @@
+import React from "react";
+import { Overlay } from "./Overlay";
+
+// see: https://on.cypress.io/mounting-react
+
+describe("<Ovleray />", () => {
+  it("Opens and closes as expected", () => {
+    cy.mount(<Overlay />);
+
+    cy.get(`[data-testid="overlay"]`).should("be.visible");
+  });
+
+  it("clicking executes the passed callback", () => {
+    let callbackCalled = false;
+    const callback = () => {
+      callbackCalled = true;
+    };
+
+    cy.mount(<Overlay callback={callback} />);
+
+    cy.get(`[data-testid="overlay"]`).should("be.visible");
+    cy.get(`[data-testid="overlay"]`).click();
+    cy.get(`[data-testid="overlay"]`).then(() => {
+      assert.isTrue(callbackCalled);
+    });
+  });
+});
--- a/components/globals/Overlay/Overlay.tsx
+++ b/components/globals/Overlay/Overlay.tsx
@@ -20,7 +20,8 @@ export const Overlay = ({
   hasOpacity?: boolean;
   hasCursor?: boolean;
 }) => {
-  function handleClose() {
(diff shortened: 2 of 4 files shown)


timarney:
I think this fixes the issue but create some new issues 

i.e. 

with 2 form cards

- click more to open the menu on one of the cards
- click more on the second card 

In this case the initial click on the second card `more` is a click on the overlay ... I have to click more again to open the menu on the second card.

The same scenario is true for other links on the page i.e. resume editing form, left nav links etc...

thiessenp-cds:
> I think this fixes the issue but create some new issues
> 
> i.e.
> 
> with 2 form cards
> 
> * click more to open the menu on one of the cards
> * click more on the second card
> 
> In this case the initial click on the second card `more` is a click on the overlay ... I have to click more again to open the menu on the second card.
> 
> The same scenario is true for other links on the page i.e. resume editing form, left nav links etc... when the overlay is open it blocks the intended link action.

Yup that is correct. The overlay idea was to allow only one active menu at a time. I think overlay's normally work this way, when click outside the active area whatever was active is closed but nothing behind is clicked. Similar to a dialog. The background could be dimmed to make it clearer though - I added an opacity option that could be enabled.

The behaviour was never defined so this is just my take on how a menu should behave. We could run it by Carolyn and Sam?

timarney:
> The overlay idea was to allow only one active menu at a time. 

Taking the multiple menus out of the picture 

The issue with the overlay is it's blocking the intended link click actions nav items

See:
https://www.youtube.com/watch?v=zls6Vo4URhk

thiessenp-cds:
From a conversation with @connorscarolyns (email) the behaviour for an open myforms card menu is currently: 
- the overlay blocks clicks outside the menu, this triggers the menu to close
  - Note: this means only one menu is active/visible at a time
  - Note: this means any controls behind the overlay (and menu) are not clickable/active while the overlay is active
- an active overlay is invisible to the user (no opacity)

timarney (approved):
Based on https://github.com/cds-snc/platform-forms-client/pull/2005#issuecomment-1527850275

Re-tested using the Share Dropdown under the form builder 

Opened the share menu 
Clicked on a link something outside the menu 
Takes an additional click --- to navigate

Approving

thiessenp-cds:
> Based on [#2005 (comment)](https://github.com/cds-snc/platform-forms-client/pull/2005#issuecomment-1527850275)
> 
> Re-tested using the Share Dropdown under the form builder
> 
> Opened the share menu Clicked on a link something outside the menu Takes an additional click --- to navigate possibly because we're passing a `modal` true prop
> 
> Approving

Thanks for looking into this and re-testing!

For the future it would be interesting to see if how users respond to the overlay behaviour with the cards menu

PR #2008, opened 2023-04-28
refactor: reworked API error responses when message was 'Malformed API Request'

craigzour:
# Summary | Résumé

- Reworked API error responses when message was 'Malformed API Request'

### DIFF
--- a/lib/middleware/csrfProtected.ts
+++ b/lib/middleware/csrfProtected.ts
@@ -4,27 +4,22 @@ import { getCsrfToken } from "next-auth/react";
 
 export const csrfProtected = (protectedMethods: string[]): MiddlewareRequest => {
   return async (req: NextApiRequest, res: NextApiResponse): Promise<MiddlewareReturn> => {
-    try {
-      if (isProtected(req, protectedMethods)) {
-        // Need to remove body from NextApiRequest due to Next-Auth bug
-        // Next-Auth sets the method Call based on if a Body is present on the request.
-        // This causes issues with CSRF protection so only required arguments are passed.
-        const clonedReq = { body: undefined, headers: req.headers };
+    if (isProtected(req, protectedMethods)) {
+      // Need to remove body from NextApiRequest due to Next-Auth bug
+      // Next-Auth sets the method Call based on if a Body is present on the request.
+      // This causes issues with CSRF protection so only required arguments are passed.
+      const clonedReq = { body: undefined, headers: req.headers };
 
-        const csrfToken = await getCsrfToken({ req: clonedReq });
-        if (csrfToken && csrfToken === req.headers["x-csrf-token"]) {
-          return { next: true };
-        } else {
-          res.status(403).json({ error: "Access Denied" });
-          return { next: false };
-        }
-      } else {
-        //allow unrestricted method
+      const csrfToken = await getCsrfToken({ req: clonedReq }).catch(() => null);
+      if (csrfToken && csrfToken === req.headers["x-csrf-token"]) {
         return { next: true };
+      } else {
+        res.status(403).json({ error: "Access Denied" });
+        return { next: false };
       }
-    } catch (error) {
-      res.status(500).json({ error: "Malformed API Request" });
-      return { next: false };
+    } else {
+      //allow unrestricted method
+      return { next: true };
     }
   };
 };
--- a/lib/middleware/jsonIDValidator.ts
+++ b/lib/middleware/jsonIDValidator.ts
@@ -14,29 +14,22 @@ export type ValidateOptions = {
  */
 export const uniqueIDValidator = (options?: ValidateOptions): MiddlewareRequest => {
   return async (req: NextApiRequest, res: NextApiResponse): Promise<MiddlewareReturn> => {
-    try {
-      if (options?.runValidationIf?.(req) === false) {
-        return { next: true };
-      }
-      const formProperties: FormProperties = options?.jsonKey
-        ? req.body[options.jsonKey]
-        : req.body;
-      const elementIDs: Array<number> = formProperties.elements.map((element) => {
-        return element.id;
+    if (options?.runValidationIf?.(req) === false) {
+      return { next: true };
+    }
+    const formProperties: FormProperties = options?.jsonKey ? req.body[options.jsonKey] : req.body;
+    const elementIDs: Array<number> = formProperties.elements.map((element) => {
+      return element.id;
+    });
+    const duplicateElementIDs = elementIDs.filter(
+      (value, index, array) => array.indexOf(value) !== index
+    );
+    if (duplicateElementIDs.length === 0) {
+      return { next: true };
+    } else {
+      res.status(400).json({
+        error: `JSON Validation Error: Duplicate IDs detected: ${duplicateElementIDs.join()}`,
       });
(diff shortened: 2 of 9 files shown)


pages/api/verify.ts
```diff
@@ -17,7 +17,7 @@ const verifyReCaptchaToken = async (req: NextApiRequest, res: NextApiResponse) =
       res.status(404).json({ error: "Bad request" });
     }
   } catch (err) {
-    res.status(500).json({ error: "Malformed API Request" });
+    res.status(400).json({ error: "Malformed API Request" });
```
bryan-robitaille, line 20: In this case if an error is thrown in the axios call it is an internal server error.

bryan-robitaille (commented):
I think the Templates API needs a closer look.  When I was tracing back where errors could be thrown and where they would be caught I noticed a mismatch in the logic.

craigzour:
> I think the Templates API needs a closer look. When I was tracing back where errors could be thrown and where they would be caught I noticed a mismatch in the logic.

Alright, my bad! 
Went with a simpler approach where I trusted the error messages. I will spend more time looking at the logic of all functions.

lib/middleware/csrfProtected.ts
```diff
@@ -4,27 +4,22 @@ import { getCsrfToken } from "next-auth/react";
 
 export const csrfProtected = (protectedMethods: string[]): MiddlewareRequest => {
   return async (req: NextApiRequest, res: NextApiResponse): Promise<MiddlewareReturn> => {
-    try {
-      if (isProtected(req, protectedMethods)) {
-        // Need to remove body from NextApiRequest due to Next-Auth bug
-        // Next-Auth sets the method Call based on if a Body is present on the request.
-        // This causes issues with CSRF protection so only required arguments are passed.
-        const clonedReq = { body: undefined, headers: req.headers };
+    if (isProtected(req, protectedMethods)) {
```
craigzour, line 7: After looking at the code I did not find a reason why this would have to be wrapped in a `try/catch` block

lib/middleware/jsonIDValidator.ts
```diff
@@ -84,57 +70,48 @@ export const layoutIDValidator = (options?: ValidateOptions): MiddlewareRequest
...
 export const subElementsIDValidator = (options?: ValidateOptions): MiddlewareRequest => {
   return async (req: NextApiRequest, res: NextApiResponse): Promise<MiddlewareReturn> => {
-    try {
-      if (options?.runValidationIf?.(req) === false) {
-        return { next: true };
-      }
-      const formProperties: FormProperties = options?.jsonKey
-        ? req.body[options.jsonKey]
-        : req.body;
+    if (options?.runValidationIf?.(req) === false) {
```
craigzour, line 73: After looking at the code I did not find a reason why this would have to be wrapped in a `try/catch` block

lib/middleware/csrfProtected.ts
bryan-robitaille: I agree.  The only unknown is if the promise of `getCsrfToken` rejects.  After looking at next-auth source code we can see that they catch errors and return null.  If you really wanted to be safe in case they change that behaviour in the future you can add a .catch() to the promise.  This way if ever the next-auth code base behavior changes it won't cause an uncaught exception in the scope of your function.
```js
const csrfToken = await getCsrfToken({ req: clonedReq }).catch(() => null);
```

lib/middleware/jsonIDValidator.ts
```diff
@@ -14,29 +14,22 @@ export type ValidateOptions = {
...
-    try {
-      if (options?.runValidationIf?.(req) === false) {
-        return { next: true };
-      }
-      const formProperties: FormProperties = options?.jsonKey
-        ? req.body[options.jsonKey]
-        : req.body;
-      const elementIDs: Array<number> = formProperties.elements.map((element) => {
-        return element.id;
+    if (options?.runValidationIf?.(req) === false) {
```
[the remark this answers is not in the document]
bryan-robitaille: Nice refactor.

pages/api/templates/index.ts
```diff
@@ -16,36 +17,36 @@ import {
...
-const allowedMethods = ["GET", "POST"];
-const authenticatedMethods = ["POST"];
+class MalformedAPIRequest extends Error {}
 
-const templates = async (
-  req: NextApiRequest,
-  res: NextApiResponse,
-  { session }: MiddlewareProps
-) => {
+const templates = async (req: NextApiRequest, res: NextApiResponse, props: MiddlewareProps) => {
```
bryan-robitaille, line 24: Nicely done!

craigzour:
@bryan-robitaille Went through the APIs I originally touched in this pull request. Mainly reworked the Template API in order to catch and send proper responses depending on what did not work.

pages/api/templates/[formID].ts
```diff
@@ -24,54 +23,48 @@ import {
...
-const allowedMethods = ["GET", "POST", "PUT", "DELETE"];
-const authenticatedMethods = ["POST", "PUT", "DELETE"];
+class MalformedAPIRequest extends Error {}
 
-const templates = async (
-  req: NextApiRequest,
-  res: NextApiResponse,
-  { session }: MiddlewareProps
-) => {
+const templates = async (req: NextApiRequest, res: NextApiResponse, props: MiddlewareProps) => {
```
Comment (bryan-robitaille) [inline pages/api/templates/[formID].ts:33]: Refactor is super well done and easy to read.

pages/api/templates/[formID].ts
```diff
@@ -94,13 +87,17 @@ const templateCRUD = async ({
   users?: { id: string; action: "add" | "remove" }[];
   sendResponsesToVault?: boolean;
 }) => {
-  const formID = request.query.formID as string;
+  const formID = request.query.formID;
+
+  if (!formID || typeof formID !== "string") {
```
Comment (bryan-robitaille) [inline pages/api/templates/[formID].ts:92]: Nice removal/extraction of `formID` from each else if statement

craigzour:
@bryan-robitaille There is something I have been thinking about while working on this. It would probably be cleaner if our APIs were dispatched in multiple files with specific path depending on the requested operation. We could extract some of the common logic our APIs have in common while refactoring them. 

Could also be a good opportunity to see if we can implement the new API features from Next JS 13
```
import { NextResponse } from 'next/server';
 
export async function GET() {
  const res = await fetch('https://data.mongodb-api.com/...', {
    headers: {
      'Content-Type': 'application/json',
      'API-Key': process.env.DATA_API_KEY,
    },
  });
  const data = await res.json();
 
  return NextResponse.json({ data });
}
```

bryan-robitaille (approved):
Amazing refactoring work!  One small thing for consideration might be the `getCsrfToken` fn that could potentially reject in a future iteration of the next-auth lib.  I'll approve the PR because the error is currently handled in the lib and will return 'null' on any caught errors but if you ever wanted to change it I can reapprove after as well.

bryan-robitaille (approved):
👍

lib/middleware/jsonIDValidator.ts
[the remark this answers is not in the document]
thiessenp-cds: Same, nothing seems like it would throw an exception

thiessenp-cds:
LGTM also - I need to speed up my PR reviews, Bryan always finishes first :)

PR #2015, opened 2023-05-01
Form `name` check for share menu

### LINKED ISSUE #1965: Sharing fresh form will fail because name is not set
Sharing a freshly created form without a set name will result in a failure.

The share API will return a `res.status(400).json({ error: "Malformed request" });` if the `filename` is empty in the request payload.
````
const filename = name ? name : i18n.language === "fr" ? form.titleFr : form.titleEn;
````

@connorscarolyns @samsadasivan Should the share feature be deactivated until the form has a name at least?

### PR DESCRIPTION
**Pull request by timarney**
# Summary | Résumé

- Adds form name check for share menu
- Updated + added cypress test(s) for new functionality 

Fixes #1965

share-name

- Create a new empty form i.e. no name etc... 
- Try sharing the form
- New `You need to enter a name` view should show

# Pull Request Checklist

### CODE REVIEW

File: public/static/locales/fr/form-builder.json [hunk 1]
```diff
@@ -527,7 +527,12 @@
       "step2": "Étape 2",
       "step2Details": "Copiez et collez les instructions dans un courriel et joignez le fichier du formulaire"
     },
-    "unpublished": "Lien n'est pas disponible pour les formulaires non publiés"
+    "unpublished": "Lien n'est pas disponible pour les formulaires non publiés",
+    "missingName": {
```
timarney, line 531: @anikbrazeau some FR text here if you can update.

craigzour:
@timarney Found a problem while testing. If you set a title and then you just clear the field you can still share the form 😕

timarney:
@craigzour looks like in the case where you add a name and then "erase" the name the previously entered value persists -- I'll take a look at the Form title and see how it saves when data is erased.

timarney:
@craigzour looks like emptying the name field isn't saving because of the gitName check below

That was introduced in:
https://github.com/cds-snc/platform-forms-client/pull/1584

Removing the `getName() === ""` check seems to fix the issue but not sure if that causes other problems ... re: 1584.  Thinking for this PR I should leave as is and look at the name saving in a different PR given if the Share dropdown is given the right info it seems to behave correctly.


```diff
const updateName = useCallback(() => {
-   if (getName() === "") {
     updateField("name", value);
-   }
}, [value, getName, updateField]);
  ```

craigzour:
> @craigzour looks like emptying the name field isn't saving because of the getName check below.
> 
> That was introduced in: #1584
> 
> Removing the `getName() === ""` check seems to fix the issue but not sure if that causes other problems ... re: 1584. Thinking for this PR I should leave as is and look at the name saving in a different PR given if the Share dropdown is given the right info it seems to behave correctly.
> 
> Thoughts?
> 
> ```diff
> const updateName = useCallback(() => {
> -   if (getName() === "") {
>      updateField("name", value);
> -   }
> }, [value, getName, updateField]);
> ```

I do not mind having this fixed it in a different PR. Just want to make sure it is done at some point in the coming days as we still have that scenario where the form sharing process could fail and produce error notifications on Slack

craigzour (approved):
LGTM :)

PR #2027, opened 2023-05-03
use common cache per branch

bryan-robitaille:
# Summary | Résumé

Reuse the yarn global cache between github action workers running on the same branch.

This should dramatically reduce the number of github caches we have in use and remove the warning of being over the limit.


craigzour (dismissed):
LGTM 👍

Issue #2052, opened 2023-05-05
Call to download API will override status on responses that have already been confirmed


craigzour:
Thanks to bug report from Jimmy, we have identified a flaw in the response table section. 
If a user downloads a response after it was confirmed then the response status will become downloaded. Calling the confirm API will not have any effect on it anymore because the confirmation code will be ignored since it has already been used once.

PR #2054, opened 2023-05-05
fix: download API can only change status for new responses

craigzour:
# Summary | Résumé

closes https://github.com/cds-snc/platform-forms-client/issues/2052

- Fixed issue where the download API would override the `Status` of a response from `Confirmed` to `Downloaded`. Now it can only update it if it is a `New` response.
- Re enable `Renders a HTML file` unit test. It was only missing a mock for the `renderToStaticNodeStream` function
- Added `aws-sdk-client-mock-jest` package for new download API unit tests
- Upgraded `aws-sdk-client-mock` to version 2.1.1

- Get yourself in a state where a new response is available for download
- Download the response (status should be updated to Downloaded)
- Confirm the response (status should be update to Confirmed)
- Download the response once again (status SHOULD NOT be updated to Downloaded)

# Pull Request Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "@typescript-eslint/parser": "^5.59.0",
     "add": "^2.0.6",
     "autoprefixer": "^10.4.4",
-    "aws-sdk-client-mock": "^0.6.2",
+    "aws-sdk-client-mock": "^2.1.1",
     "axe-core": "^4.5.2",
     "babel-loader": "^8.2.4",
     "babel-plugin-dynamic-import-node": "^2.3.3",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### Fixed
+
+- Download API will not override a response status that is different than `New`. [2052](https://github.com/cds-snc/platform-forms-client/issues/2052)
+
 ## [3.0.4] 2023-05-05
 
 ### Fixed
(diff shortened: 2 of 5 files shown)


bryan-robitaille (changes requested):
The logic change to the application is good.
There is an issue however with using `aws-sdk-client-mock-jest`.  It installs Jest types from version 28 which conflict with the Jest 29 version we are running.  This is causing linting and type issues with the majority of Jest Tests.

Screenshot 2023-05-08 at 8 43 44 AM
Screenshot 2023-05-08 at 8 49 07 AM

thiessenp-cds (commented):
LGTM :)

I tested this out by downloading, confirming, then re-download 3 submissions. The submissions stayed in the confirmed state. I also tried changing the 3 confirmed stated to the "problem" state and that still works. Also downloading those submissions leaves the state in "problem" which I think is what we want. Looks good!

I'll leave the PR open encase anyone else has comments :)

craigzour:
> The logic change to the application is good. There is an issue however with using `aws-sdk-client-mock-jest`. It installs Jest types from version 28 which conflict with the Jest 29 version we are running. This is causing linting and type issues with the majority of Jest Tests.
> 
> Screenshot 2023-05-08 at 8 43 44 AM Screenshot 2023-05-08 at 8 49 07 AM

Will look into it and probably change it (will find another way to build the tests). 
Pretty sure I have been having those linting/types issues for a few weeks now (even before making this change) but did not take time to see what was going on since the tests were still passing.

bryan-robitaille:
I also wanted to make sure it wasn't something previously done in the repo.   There are a few different things causing the linting errors.  One is a cypress file type declaration that is overriding the globals but the other is this new package.  When I removed the package the Jest type errors did go away.  I had done some work in the past to move away from `@types/jest` because they have not updated the package since Jest 28 and the types package is not maintained by Jest.

bryan-robitaille:
This might help as an example of how we've mocked deep aws sdk functions in the past.  https://github.com/cds-snc/platform-forms-client/blob/4d4b489059b4f689ed629092acbe7219a346992d/__tests__/api/signup/confirm.test.ts#L69

package.json
```diff
@@ -124,7 +124,7 @@
     "@typescript-eslint/parser": "^5.59.0",
     "add": "^2.0.6",
     "autoprefixer": "^10.4.4",
-    "aws-sdk-client-mock": "^0.6.2",
+    "aws-sdk-client-mock": "^2.1.1",
```
craigzour, line 127: This one does not have Jest as a dependency so I figured we could keep it. It is the latest version available.

bryan-robitaille (approved):
:+1

PR #2071, opened 2023-05-08
chore/Cypress goes to Typescript

bryan-robitaille:
# Summary | Résumé

Cypress vs Jest globals fighting for dominance.
Problem:  The introduction of typescript for Cypress tests caused an issue where Cypress types for assertions (expect, it) were overriding the Jest types in Jest Tests.
Solution:  Cypress has gone full typescript with it's own `tsconfig.json` that ensures Cypress types remain excluded from the main root `tsconfig.json`. 

Clean up:
- Removes unused 'Agent' file in root of repo
- Added missing 'acceptableUse` property on a few mock user objects.


timarney (dismissed):
Nice 

At some point we might want to create a type for 

 let nextData: Record<string, unknown>;  and the  createdID.toString()));
 
super minor

bryan-robitaille:
> At some point we might want to create a type for
> 
> let nextData: Record<string, unknown>; and the createdID.toString()));

I'll do that now since Git is having issues.  I have one other local commit that I can't push to this PR until Git service resumes.

PR #2079, opened 2023-05-10
Content edits to account flows

anikbrazeau:
In reviewing flows around accounts and 2FA, found some opportunities for refining copy.

Includes:

- Linking to Terms of use (rather than SLA) in account creation
- Switching language around SLA to terms of use in unlocking publication
- Making the initial account creation code "verification code" to match email content
- Making subsequent codes for password reset and 2fa "security code" named to match their purpose


pages/signup/register.tsx
```diff
@@ -219,7 +219,7 @@ const Register = () => {
               </div>
               <p className="mb-10 -mt-8 gc-description">
                 {t("signUpRegistration.slaAgreement")}&nbsp;
```
craigzour, line 221: Just to avoid any confusion I would rename both `slaAgreement` and `slaAgreementLink` to `termsAgreement` and `termsAgreementLink` in the `register.tsx` and `signup.json` (EN and FR) files.

craigzour (changes requested):

PR #2084, opened 2023-05-10
Release/3.0.5

bryan-robitaille:
### Fixed

- Download API will not override a response status that is different than `New`. [2052](https://github.com/cds-snc/platform-forms-client/issues/2052)
- Language Toggle
- Button styling across the product
- Error message when a user tries to login and the service is unavailable
- Unsupported browser page / Java Script not enable page


craigzour (approved):
LGTM :)

thiessenp-cds:
LGTM also :)

Did some basic browser testing - some future thoughts:
- I would occasionally see a console error but it was hard to reproduce so it's probably related to my local setup "assertion failed: input argument is not an HTMLInput Element". If anyone else sees this in staging, might be worth looking into
- meta charset is not first item in head, it probably should be
- I saw a warning about using x-frames-options in the response headers and using CSP instead. I would think x-frames-options is more supported by browsers. I double checked with OWASP (https://cheatsheetseries.owasp.org/cheatsheets/Clickjacking_Defense_Cheat_Sheet.html) and I suspect a pen tester would have said something but noting encase anyone comes across this

bryan-robitaille:
In terms of x-frames-options we actually do allow our forms to be added as Iframes on external sites.

thiessenp-cds:
> @thiessenp-cds in terms of x-frames-options we actually do allow our forms to be added as Iframes on external sites. This was to meet a specific use case early on in the GCForms project however may need to remove in the future.

Oh I had no idea, good to know, thanks! :)

PR #2088, opened 2023-05-10
Release/3.0.5

bryan-robitaille:
### Fixed

- Download API will not override a response status that is different than `New`. [2052](https://github.com/cds-snc/platform-forms-client/issues/2052)
- Language Toggle
- Button styling across the product
- Error message when a user tries to login and the service is unavailable
- Unsupported browser page / Java Script not enable page


craigzour (approved):
LGTM :)

PR #2139, opened 2023-05-19
feat: added new API to request new 2FA verification code

craigzour:
# Summary | Résumé

- Added very basic API path to request new 2FA verification code. 

API path = `/api/auth/2fa/request-new-verification-code`
Request method = `POST`
Request body = `email: <user_email>`
Requires CSFR token


craigzour:
> Good for testing but the request new 2fa code will need to be thought out a little more. A potential attack vector would be to flood the API endpoint with email addresses and it would disrupt the valid authentication flow of users. We may need to send some sort of temporary value back to the client with the cognito credentials success to be used as an identifier when requesting a new code. So the identifier and email need to match before we regenerate a new 2FA code.

Oh yes for sure!

PR #2140, opened 2023-05-19
Clarify description for Open a form file

### LINKED ISSUE #2134: Open form file text is confusing
The text for opening form files is misleading to users, who mistakenly think they can upload files like PDFs and have them converted into an online form.

anikbrazeau:
**Fixes #2134** 

Attempt to clarify that uploading PDFs is not the intent, but rather that it's a file originating from GC Forms.
Tried to avoid using JSON file to not confuse our first-time user segment.


public/static/locales/fr/form-builder.json
```diff
@@ -552,7 +552,7 @@
   "startH2": "Créer un formulaire",
   "startH3": "Ouvrir un formulaire",
   "startP1": "Commencez à partir de zéro.",
-  "startP2": "Modifiez un formulaire téléversé de votre ordinateur.",
+  "startP2": "Modifiez un formulaire qui a été créé et téléchargé à partir de Formulaires GC",
```
anikbrazeau, line 555:
```suggestion
  "startP2": "Modifiez un formulaire qui a été créé et téléchargé à partir de Formulaires GC.",
```

bryan-robitaille (approved):

PR #2154, opened 2023-05-23
feat: introducing authentication flow token

craigzour:
# Summary | Résumé

- Introducing authentication flow token as an extra layer of security to our authentication APIs (especially the one that can request a new 2FA verification token)

### DIFF
--- a/lib/auth/cognito.ts
+++ b/lib/auth/cognito.ts
@@ -25,6 +25,8 @@ type DecodedCognitoToken = {
   email: string;
 };
 
+export type AuthenticationFlowToken = string;
+
 export enum Validate2FAVerificationCodeResultStatus {
   VALID,
   INVALID,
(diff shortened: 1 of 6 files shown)


craigzour:
Since 2FA code is not yet in develop and I had to change the Prisma schema, I decided to remove the initial migration script (that was created last week) in order to only have one associated to the 2FA feature. 
As a developer make sure you just run `yarn prisma:reset`.

cc @timarney

craigzour:
@timarney With this new security mechanism the first step in the authentication flow (Cognito authentication) will return `{"status":"success","challengeState":"MFA","authenticationFlowToken":"cli0kivbt0004d9wdlg760gim"}`. You will then have to pass the `authenticationFlowToken` to all authentication APIs when going through the 2FA steps (validate verification code + resend new verification code).

Let me know if you have any questions.

PR #2156, opened 2023-05-23
Consistent data removal length of time

### LINKED ISSUE #1982: Incorrect copy in the confirmation modal
## Context

> After you confirm receipt, responses will be available for 7 days then automatically removed from our database.

This copy is incorrect. Check with Devs to find out how this works, but if I confirm right away, the response is set to delete after 30 days, not sure how this works if the confirmation is delayed.

## Screenshots of the discrepancy 

See 7 days:
Screenshot 2023-05-23 at 2 51 23 PM

See 30 days:
Screenshot 2023-05-23 at 2 51 32 PM

anikbrazeau:
Fixes #1982


bryan-robitaille (approved):
👍

PR #2162, opened 2023-05-24
Added unit tests for new 2FA code

craigzour:
# Summary | Résumé

- Reduced number of allowed attempts from 10 to 3 for 2FA validation mechanism
- Added unit tests for new 2FA code

### DIFF
--- a/lib/tests/auth/cognito.test.ts
+++ b/lib/tests/auth/cognito.test.ts
@@ -1,0 +1,307 @@
+/**
+ * @jest-environment node
+ */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import Redis from "ioredis-mock";
+import { mockClient } from "aws-sdk-client-mock";
+import {
+  CognitoIdentityProviderClient,
+  AdminInitiateAuthCommand,
+  AdminInitiateAuthResponse,
+} from "@aws-sdk/client-cognito-identity-provider";
+import { prismaMock } from "@jestUtils";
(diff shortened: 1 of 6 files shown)


lib/tests/auth/cognito.test.ts
```diff
@@ -0,0 +1,313 @@
...
+  getRedisInstance: jest.fn(() => redis),
+}));
+
+const cognitoMock = mockClient(CognitoIdentityProviderClient);
+
+const mockSendEmail = {
+  sendEmail: jest.fn(() => Promise.resolve()),
+};
+jest.mock("notifications-node-client", () => ({
+  NotifyClient: jest.fn(() => mockSendEmail),
```
bryan-robitaille, line 35:
```suggestion
  NotifyClient: jest.fn(),
```
There's no need to go a level deeper to mock the `sendEmail` function unless you wanted to test that `sendEmail` has been called.  Jest mock will stub out the initialization class and all sub functions.

bryan-robitaille (changes requested):
Nice!
A small tweak to add the eslint ignore at the top of the test file and looks good to go.

lib/tests/auth/cognito.test.ts
craigzour: That is right! I think I was planning on checking if it was called but then decided to use `expect(mockSendVerificationCode).toHaveBeenCalled();` instead of `sendEmail`

PR #2167, opened 2023-05-24
Fixes ErrorPannel home link depending on auth status

thiessenp-cds:
# Summary | Résumé

Updates the error pages (ErrorPannel) home link depending on auth status. 

If the user is authenticated then the home link should be to the form builder, otherwise the link should be to the public forms site.


thiessenp-cds:
Note: there is a follow up question here https://github.com/cds-snc/platform-forms-client/issues/2169

bryan-robitaille (approved):

PR #2179, opened 2023-05-29
feat: renew 2FA verification code when beginning 2FA auth session even if one already exists

craigzour:
# Summary | Résumé

- New verification code is being generated and saved when calling `begin2FAAuthentication` even if a code already exists.


bryan-robitaille (approved):
👍  LGTM

PR #2190, opened 2023-05-31
chore: update develop with main

bryan-robitaille:
# Summary | Résumé
Brings Notify API logging commit from main to develop before release.


bryan-robitaille:
Opening a different PR because develop and main cannot be used as base branches

PR #2195, opened 2023-06-01
Feature/remove account confirmation api

craigzour:
# Summary | Résumé

- Removed account confirmation APIs (`/api/signup/confirm` and `/api/signup/resendconfirmation`)
- Moved `/api/signup/register` to `/api/account/register`
- Reworked unit test for 2FA verification code generation


lib/auth/2fa.ts
```diff
@@ -1,19 +1,14 @@
...
+
 const TEMPLATE_ID = process.env.TEMPLATE_ID;
 const NOTIFY_API_KEY = process.env.NOTIFY_API_KEY;
 
-export const generateVerificationCode = async () => {
-  // Temporary - for testing purposes
-  // Will create more robust code generation in the future
-  return await generateTokenCode(5);
-};
+export const generateVerificationCode = async () => await generateTokenCode(5);
```
bryan-robitaille, line 8:
```suggestion
export const generateVerificationCode = async () => generateTokenCode(5);
```
Small performance/memory gain by not rewrapping a resolved promise in another promise.
I'll add a eslint rule for this in another PR.

lib/tests/auth/cognito.test.ts
```diff
@@ -120,7 +120,7 @@ describe("Test Cognito library", () => {
         id: mockedId,
       });
 
-      mockGenerateVerificationCode.mockReturnValueOnce("a1é3_8");
```
bryan-robitaille, line 123: I know the value is purely contained within the test but I wonder if we should only use the characters available for an actual 2FA code?

bryan-robitaille (approved):
Looks good to me.
I added an index.ts file to `lib/auth` and fixed the downstream imports.  I thought trying to import from `@lib/auth/auth` was a little redundant.

lib/tests/auth/cognito.test.ts
craigzour: Sure I can change it :)

PR #2200, opened 2023-06-02
Put back formTitle for Google Analytics and GTM

daine:
# Summary | Résumé

Form titles are used for tracking success in google analytics
There's no other way to easily see this within the application, so putting it back as one of our data points in google analytics


craigzour (approved):
LGTM :)

PR #2202, opened 2023-06-02
fix: GC Notify callback API should only handle form submission type of email

craigzour:
# Summary | Résumé

Our GC Notify callback API cannot differentiate between all the types of emails we are sending (form submission, account confirmation, share a form, etc...) with our GC Notify account.
The form submission email does have a reference (submission ID we are manually passing when sending a form submission) that can be sent back to us through the callback while the other emails do not have any (`null`) because it would be irrelevant.


bryan-robitaille (commented):
Nice work figuring out the issue.  We should also add a logMessage.warn() for when the reference is set but the status is a permanent failure so we can track those form responses but not retry them.  The processed reference should be removed but not retried by adding them to the reprocess queue.

PR #2207, opened 2023-06-05
2fA Account Register

thiessenp-cds:
# Summary | Résumé

Test by registering a new user. (will need the auto confirmation work)

What's left to do:
 -  Add "new code" API call from Verify when ready
 - Unit Tests but I'd say to hold off until the feature is more complete/stable.
 - Refactoring. Step 1 was splitting up the useAuth hook into several hooks.  Step 2 has been moving some of the auth hooks that aren't shared back into the related component. Step 3 TODO would be re-thinking how the state is shared between the parent to child components


components/auth/Verify.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { TextInput, Label, Alert, Description } from "@components/forms";
+import { useTranslation } from "next-i18next";
+import * as Yup from "yup";
+import Link from "next/link";
+import { ErrorStatus } from "@components/forms/Alert/Alert";
+import { useAuthErrors } from "@lib/hooks/auth/useAuthErrors";
+import { Button } from "@components/globals";
+import { useVerify } from "@lib/hooks/auth";
+
+// TODO:
```
timarney, line 13: Let's remove this

pages/signup/account-created.tsx
```diff
@@ -30,31 +30,23 @@ export default function AccountCreated() {
       <div className="mt-20">
         <StyledLink
           href={`/${i18n.language}/unlock-publishing/`}
-          className={` 
-            no-underline visited:text-white-default mr-8
-            bg-blue-dark text-white-default border-black-default py-4 px-8 rounded-lg border-2 border-solid
-            hover:text-white-default hover:bg-blue-light active:text-white-default active:bg-blue-active active:top-0.5
-            focus:outline-[3px] focus:outline-blue-focus focus:outline focus:outline-offset-2 focus:bg-blue-focus focus:text-white-default disabled:cursor-not-allowed disabled:text-gray-500
-          `}
+          theme="primaryButton"
```
timarney, line 33: 👍

pages/signup/register.tsx
```diff
@@ -133,7 +187,7 @@ const Register = () => {
                 ) : undefined}
               </Alert>
             )}
-            {Object.keys(errors).length > 0 && !authErrorsState.isError && (
+            {/* {Object.keys(errors).length > 0 && !authErrorsState.isError && (
```
timarney, line 190: Same here as per above

components/auth/Verify.tsx
thiessenp-cds: Got it, will replace <Confirmation> with <Verification>

pages/signup/register.tsx
```diff
@@ -23,12 +21,69 @@ import { getServerSession } from "next-auth/next";
...
+      // Register the user. An error is thrown if the username exists etc.
+      await fetchWithCsrfToken("/api/account/register", {
+        username,
+        password,
+        name,
+      });
+
+      // TEMP - to remove once confirmation code removed
+      // Add a Breakpoint below, to give you time first confirm the account in cognito before login
+      logMessage.info("");
```
timarney, line 69: Can remove this :)

thiessenp-cds: True :) The problem though is without this there is no way to test the 2FA flow after the register step. We still need a delay to allow us to go into Cognito and manually confirm an account before the code moves on to the auto-login step.

But if we'd rather get the code ready for future 2FA testing when the manual confirm step is no longer needed, then it should be removed?

timarney: @bryan-robitaille we're looking to get setup to test without manually confirming correct?

thiessenp-cds: Sounds good, will remove :)

components/auth/Verify.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  // TODO
+  const handleReVerify = async () => {
+    setIsReverify(false);
+  };
+
+  const validationSchema = Yup.object().shape({
+    verificationCode: Yup.string()
+      .required(t("verify.fields.confirmationCode.error.notEmpty"))
+      .matches(/^[0-9a-z]{5}?$/i, t("verify.fields.confirmationCode.error.length")),
```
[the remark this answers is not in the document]
craigzour: I wonder if we should not just skip the validation here. No matter what the code is we will be validating it on the backend side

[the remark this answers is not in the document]
thiessenp-cds: Makes sense. The string validation is more complex than I realized and probably easy to get wrong. I'll only leave a not null and length of 5 characters check.

bryan-robitaille:
I pushed a few changes that implement the 'resend verification code' that was marked as 'todo'  in `@components/auth/Verify`

craigzour (approved):
LGTM :)

timarney:
Looks good - can make further adjustments as needed in a fresh PR.

thiessenp-cds:
FYI: I added back the Confirmation code related components. From what I understand the password reset will be using the "old" confirmation code and not our new verify code. So both the Confirmation and Verify components will exist in the repo. Mental note for the future to remove the confirmation related code if we replace it with our new verify code.

craigzour (approved):
Tested and approved!

Issue #2115, opened 2023-05-16
Update "Name your form" link destination


anikbrazeau:
# Issue/opportunity

Link to "Name your form" before sharing (in share menu) goes to form title field (ex: `GIve us feedback`) rather than form name field (ex: `Unnamed form file`)

# Screenshots

<img width="1169" alt="Screenshot 2023-05-16 at 12 27 02 PM" src="https://github.com/cds-snc/platform-forms-client/assets/38330843/3e38b928-3b50-4a21-b764-35883175e380">

<img width="1151" alt="Screenshot 2023-05-16 at 12 28 06 PM" src="https://github.com/cds-snc/platform-forms-client/assets/38330843/e6dbaeea-5da2-4975-b176-32ca89c9901b">

_Found on: Production, Mac, Chrome_

anikbrazeau:
Potentially related #2157

srtalbot:
Good to go, waiting on PR review.

PR #2213, opened 2023-06-05
Fix for "name your form" link

daine:
# Summary | Résumé

This PR fixes issue #2115 
- #2115

_For any UI changes please include screenshots at different breakpoints. It would also be helpful to include a
before and after comparison for each change i.e._

**New look** The placeholder for the filename "Unnamed Form FIle" used to be black. Not sure if this was intentional, so I changed it to look the same as the form title wherein it's grayed out until the user starts typing.

| Screenshot 2023-06-05 at 3 26 23 PM| Screenshot 2023-06-05 at 3 27 57 PM
 |Input on focus:Screenshot 2023-06-05 at 3 26 27 PM|Input on focus:Screenshot 2023-06-05 at 3 28 06 PM|

Using the form builder, leave the file name blank. Attempt to share the form by hitting the "Share" button at the top right. You will receive an error message saying "You must name your form". Click on the "name your form" link. It should make the cursor focus on the text input for naming the form, which has a placeholder "Unnamed form file"

- [x] Need to ask designers if changing the placeholder colour from black to gray is approved @samsadasivan @anikbrazeau @connorscarolyns 

# Pull Request Checklist


craigzour (dismissed):
Works well!
I see one unresolved question where you would like the approval of design team. Will approve and let you make sure you talk to them before merging ;)

bryan-robitaille (changes requested):
I'll try to provide some technical guidance on how we can move away from driving component state through url params after lunch but I noticed 2 things in testing.

1. If the form name is not completed after focus a user can still share the form.
save with no form name
2. If the form name is not completed and a user shifts focus elsewhere on the page the focus button will not return focus to the form name input a second time.
share state resets

craigzour:
> I'll try to provide some technical guidance on how we can move away from driving component state through url params after lunch but I noticed 2 things in testing.
> 
> 1. If the form name is not completed after focus a user can still share the form.
>    save with no form name
> 2. If the form name is not completed and a user shifts focus elsewhere on the page the focus button will not return focus to the form name input a second time.
>    share state resets

I think the reason why you can share even if the form name is empty is because you have put something in the form title. If I remember correctly we needed either one to be able to share a form.

daine:
@bryan-robitaille Thank you! Some thoughts:
> I'll try to provide some technical guidance on how we can move away from driving component state through url params after lunch but I noticed 2 things in testing.

Thanks! The changes in this PR follows the pattern and implementation of the old code so I can see why it needs to be updated

> > 1. If the form name is not completed after focus a user can still share the form.

> I think the reason why you can share even if the form name is empty is because you have put something in the form title. If I remember correctly we needed either one to be able to share a form. Good catch on the second issue!

I believe @craigzour is correct, I'm working on a separate (but related) issue to address the issues with form names and titles

> 2. If the form name is not completed and a user shifts focus elsewhere on the page the focus button will not return focus to the form name input a second time.

This is the same behaviour with the old one that I've found, so it's not a new bug but rather a leftover from the old behavior. I think it's definitely tied in with potentially just changing the previous implementation altogether!

samsadasivan:
Hey Daine! I like this - could we just switch the shade to gray-500 (#6b7280) for colour contrast for AA. slate-500 (#64748b) also works but may be too fancy hehe 


image

timarney:
> Let's work together to get this into a more 'React' state of mind.

@bryan-robitaille fyi if your thinking about refs it gets a little tricky given where the components get built. For the "most" of the form builder elements now have a refs provider which store the "els" in an array.   The "file name" is part of the header hard to get the ref over to the share modal --- the form builder also does quite a bit of focusing so if I remember correctly it was "stealing focus"  Perhaps you have some tricks for that sort of thing :)

timarney:
As far as the code goes this fixes the issue as described.

Before approving ... from the design end do we want to allow sharing an empty form that has been given a filename?

Related ....

I can't find an issue for it but the "file name" used to auto-populate when a "file title was filled out".    It does fill it out if you enter a "form title" and than navigate to another page.  Fixing that could help.

samsadasivan:
It definitely used to auto-populated in FR/EN

craigzour (approved):
LGTM :) Nice work!

components/form-builder/app/navigation/FileName.tsx
```diff
@@ -17,6 +19,33 @@ export const FileNameInput = () => {
...
+
+  // React Hook that lets you customize the handle exposed as a ref
+  // In this case we are only exposing the focus function
+  useImperativeHandle(
+    remoteRef,
+    () => {
+      return {
+        focus() {
+          fileNameInput.current?.focus();
+          logMessage.debug("Calling focus function on fileNameInput");
```
timarney, line 42: Let's remove the debug log here.

craigzour (dismissed):

lib/hooks/useRefStore.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+const RefStoreContext = createContext<RefStore | undefined>(undefined);
+
+export const RefStoreProvider = ({ children }: { children: React.ReactNode }) => {
+  const refStore = useRef(new Map<string, React.RefObject<HTMLElement>>());
+
+  const getRef = useCallback(
+    (key: string) => {
+      const value = refStore.current.get(key);
+      logMessage.debug(`getRef :key: ${key}, value: ${value}`);
```
timarney, line 17: Probably don't need this anymore?

lib/hooks/useRefStore.tsx
```diff
@@ -0,0 +1,47 @@
...
+      return value;
+    },
+    [refStore]
+  );
+
+  const setRef = useCallback(
+    (key: string, ref: React.RefObject<HTMLElement>) => {
+      const store = refStore.current;
+      store.set(key, ref);
+      logMessage.debug(`setRef :key: ${key}, ref: ${ref}`);
```
timarney, line 27: Probably don't need this anymore?

timarney (dismissed):
Approving --- just a couple of debug messages left

components/form-builder/app/navigation/FileName.tsx
```diff
@@ -17,6 +18,32 @@ export const FileNameInput = () => {
...
+    setRef("fileNameInput", remoteRef);
+
+    return () => {
+      removeRef("fileNameInput");
+    };
+  }, [remoteRef, setRef, removeRef]);
+
+  // React Hook that lets you customize the handle exposed as a ref
+  // In this case we are only exposing the focus function
+  useImperativeHandle(
```
thiessenp-cds, line 35: Nice!

thiessenp-cds (approved):
I took a quick look at the accessibility issue with the Share menu. The sub menu will need to add some keyboard navigation etc. 

I can confirm that this issue is an existing issue (develop branch). I think it makes sense to open a new issue about this, out of scope etc. I'll approve the work along side other approvals. Nice work! :)

PR #2225, opened 2023-06-07
fix: 2fa misc ui updates

thiessenp-cds:
# Summary | Résumé

Misc UI updates for the Login, Register, and Verify screens


components/auth/Verify.tsx
```diff
@@ -111,12 +111,12 @@ export const Verify = ({ username, authenticationFlowToken }: VerifyProps): Reac
                 >
                   {t("verify.fields.confirmationCode.label")}
                 </Label>
-                <Description className="text-p text-black-default" id={"verificationCode-hint"}>
```
thiessenp-cds, line 114: Slowly dropping the Description component to be more consistent. The Description component is specific to forms needs and the app needs more flexibility sometimes e.g. having a bullet list
This could be solved by moving e.g. the bullet list outside the Description but then a container div would be needed to setup the aria-describes reference. More work for no benefit

pages/auth/login.tsx
```diff
@@ -121,24 +121,22 @@ const LoginStep = ({
...
                   name={"username"}
                   required
+                  ariaDescribedBy="login-description"
                 />
               </div>
               <div className="focus-group">
                 <Label id={"label-password"} htmlFor={"password"} className="required" required>
                   {t("fields.password.label")}
                 </Label>
-                <Description id="password" className="text-p text-black-default">
```
thiessenp-cds, line 139: Removing the hint - not that helpful

components/auth/Verify.tsx
```diff
@@ -111,12 +111,12 @@ export const Verify = ({ username, authenticationFlowToken }: VerifyProps): Reac
...
+                <div className="text-p text-black-default mb-2" id={"verificationCode-hint"}>
                   {t("verify.fields.confirmationCode.description")}
-                </Description>
+                </div>
                 <TextInput
                   className="h-10 w-36 rounded"
-                  type="password"
```
thiessenp-cds, line 119: Design confirmed that the code should be visible to the user

PR #2231, opened 2023-06-08
fix: cypress test login page validation checks

bryan-robitaille:
# Summary | Résumé

Fix's cypress login page test:
- Ensures form is visible before starting any actions
- Checks for validation errors on code length and symbols
- Successfully logs in using verification code.

run cypress tests


craigzour (commented):
LGTM :) There is something weird with the commits though. It appears that the first commit has a different identifier but is the same you have already merged earlier today.

PR #2246, opened 2023-06-09
Fix/a11y misc

thiessenp-cds:
# Summary | Résumé

The changes include some minor accessibility fixes:
 - Fixes Verify "pages" not updating the HEAD title
 - Fixes Verify "pages" not focussing the page header on load
 - Fixes lost focus when going from Verify to ReVerify
 - Fixes Form Builder Share dropdown sub menu to be activate-able by keyboard when a form has no title

Test the Verify page changes:
- start VoiceOver (or similar)
- navigating to the Sign In page, Sign In
- when the Verify "page" loads, the header should be focussed and announced, the page title updated as well
- click the "Send a new code" link
- when the Re-Send code "page" loads, the header should be focussed and announced, the page title updated as well
- click the "re-send code" button
- when the Verify "page" loads, the header should be focussed and announced, the page title updated as well

Test the Share dropdown sub menu:
- navigate to Form-builder, Design a form
- using the keyboard tab to the Share menu in the header and key enter
- the sub menu should open, key enter again
- the Form title field should be focussed 
- (bonus: do the above running VoiceOver and when the sub menu is focussed the "Activate to move.." text should be announced alongside the visible text)


lib/hooks/useFocusIt.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Utility function to focus an element on load. This helps manage the users focus
+ * e.g. on a page update, set the focus to the heading.
+ *
+ * @param elRef React element to focus
+ * @param addTabindIndex Whether to add a tab-index attribute. If true (default), a
+ *    `tab-index=-1` attribute is is automatically added to the element if none exist.
+ */
+export const useFocusIt = ({
```
thiessenp-cds, line 11: Definitely not how Hooks were really intended (sharing state + stateful logic). Though I can't think of a better way to tap into the React life-cycle and do this as a util method

components/auth/ReVerify.tsx
```diff
@@ -84,7 +92,9 @@ export const ReVerify = ({
           ) : undefined}
         </Alert>
       )}
-      <h1 className="border-0 mt-6 mb-6">{t("reVerify.title")}</h1>
+      <h1 ref={headingRef} className="border-0 mt-6 mb-6">
```
thiessenp-cds, line 95: I completely forgot to add this. So when a sort of SPA model is being used to update "pages" the user should be notified in some way that the page has changed. This is accomplished by updating the HEAD TITLE el above but to make it really clear, focussing the main heading, really helps. Both are announced actually, this really gives clarity and also takes care of managing the user focus.
This could also be accomplished in other ways like wrapping the entire "page" in a live-region. The problem with this is that Assistive Technologies will often ignore this update because the diff in the accessibility tree (tree built from DOM with all the accessible info) is too big / lazy AT programmers? :).

lib/hooks/useFocusIt.tsx
```diff
@@ -0,0 +1,43 @@
...
+      if (addTabindIndex && !el.getAttribute("tabIndex")) {
+        el.setAttribute("tabIndex", "-1");
+      }
+
+      // Give the DOM a little time to update
+      setTimeout(() => {
+        el?.focus();
+      }, 40);
+    }
+    // eslint-disable-next-line
```
thiessenp-cds, line 41: ESLint doesn't like restructuring in a useEffect array but I think this is the only way to pass a dynamic set of any reactive values

components/form-builder/app/navigation/ShareDropdown.tsx
```diff
@@ -56,24 +55,24 @@ export const ShareDropdown = () => {
...
-                      // Timeout is currently required because menu steals focus on close.
-                      setTimeout(() => {
-                        getRef("fileNameInput")?.current?.focus();
-                      }, 50);
-                    }}
-                  >
-                    {t("share.missingName.message2")}
-                  </Button>{" "}
-                  {t("share.missingName.message3")}
+              <DropdownMenuPrimitive.Item
```
thiessenp-cds, line 58: This seemed the simplest fix.  @timarney if there is a better way, just let me know. The sub men needs to be "activate-able" by enter. Some way to let the user know the menu is "clickable" and what will happen when s/he clicks it would be helpful but not necessarily required

timarney:
PR looks good -- just need to fix-up the hover state.

Screenshot 2023-06-13 at 8 55 05 AM

thiessenp-cds:
> PR looks good -- just need to fix-up the hover state.
> 
> Screenshot 2023-06-13 at 8 55 05 AM

Good find. I updated the hover state of the "link" to match the rest of the text and not look so out of place

timarney:
@thiessenp-cds 

Wondering if the useFocusIt

Should return a callback instead of the useEffect

If we return a callback whatever component is using it could manage the dependancy array higher up --- which gets us back proper linting to avoid messing up the dependancy array.  Thoughts?

```
const setFocus = useCallback(() => {

    const el = elRef.current;
    if (el) {
      // Add a tabindex if one doesn't exist on the element. This is needed to focus an element,
      // unless it is a form control like an input or button
      if (addTabindIndex && !el.getAttribute("tabIndex")) {
        el.setAttribute("tabIndex", "-1");
      }

      // Give the DOM a little time to update
      setTimeout(() => {
        el?.focus();
      }, 40);
    }
  }, [elRef, addTabindIndex]);

  return setFocus;
  ```

File: public/static/locales/fr/form-builder.json [hunk 5]
```diff
@@ -539,7 +539,8 @@
     "missingName": {
       "message1": "Veuillez",
       "message2": "donner un nom à votre formulaire",
-      "message3": "avant de partager."
+      "message3": "avant de partager.",
+      "message4": "Activate to move focus to the form title field[FR]"
```
thiessenp-cds, line 543: @anikbrazeau A new string for translation. Feel free to replace the content with anything sounding better. The context:  the use is creating a new form, keyed the Share menu drop down and the menu dropdown currently has focus

anikbrazeau: Is it to the title of the form (En + Fr separate) or the name of the form file (unilingual)?

anikbrazeau, line 543:
```suggestion
      "message4": "Activez pour déplacer le focus sur le champ du titre du formulaire"
```

thiessenp-cds:
> @thiessenp-cds
> 
> Wondering if the useFocusIt
> 
> Should return a callback instead of the useEffect
> 
> If we return a callback whatever component is using it could manage the dependancy array higher up --- which gets us back proper linting to avoid messing up the dependancy array. Thoughts?
> 
> ```
> const setFocus = useCallback(() => {
> 
>     const el = elRef.current;
>     if (el) {
>       // Add a tabindex if one doesn't exist on the element. This is needed to focus an element,
>       // unless it is a form control like an input or button
>       if (addTabindIndex && !el.getAttribute("tabIndex")) {
>         el.setAttribute("tabIndex", "-1");
>       }
> 
>       // Give the DOM a little time to update
>       setTimeout(() => {
>         el?.focus();
>       }, 40);
>     }
>   }, [elRef, addTabindIndex]);
> 
>   return setFocus;
> ```

I originally stayed away from useCallback because I didn't think it would "respect" the React lifecycle of when the component was fully mounted.  From testing it looks like useCallback does and also removes the need for dependencies, it simply calls the function when the component is mounted. Exactly what I was trying to do with useEffect. This way also removes the need to assign a ref to a var and then pass it to the hook. A little cleaner :)

public/static/locales/fr/form-builder.json
thiessenp-cds: Oh right, good point. It should be the name of the form file. So "... the form name"

timarney:
Sorry Pete I mis-read the original code I thought useFocusIt was happening via a user interaction.

How about making a new component like this.

```
export const FocusHeading = ({ as, className, children }) => {

      const inputReference = useRef(null);
      
       useEffect(() => {
        inputReference.current.focus();
        }, []);


	const As = as || 'h2';
	return <As className={className}>{children}</As>;
};
```

<FocusHeading as="h2">your text</FocusHeading>

thiessenp-cds:
> Sorry Pete I mis-read the original code I thought useFocusIt was happening via a user interaction.
> 
> How about making a new component something like:
> 
> ```js
> export const FocusHeading = ({ as, className, children }) => {
> 
>       const inputReference = useRef(null);
>       
>        useEffect(() => {
>         inputReference.current.focus();
>         }, []);
> 
> 
> 	const As = as || 'h2';
> 	return <As ref={inputReference} className={className}>{children}</As>;
> };
> ```
> 
> ```js
> <FocusHeading as="h2">your text</FocusHeading>
> ```

Decided to go back to the useEffect way. I really like the idea of this util being a component. But I couldn't get this to work because of the lifecycle - component created outside/after/before :) the useEffect or maybe the useEffect applies to the container..

timarney (approved):

PR #2278, opened 2023-06-13
Fix/login error message

thiessenp-cds:
# Summary | Résumé

Updates the login page to handle how errors are received. Only the logic was changed, not the error messages etc.

Try an invalid login. An auth related error should show.
You could try programmatically throwing different errors from the API to see the right error message displayed in the UI. Maybe there is a better way though.

Try creating a new user. Probably the main thing to test is that this still works.


lib/hooks/auth/useLogin.ts
```diff
@@ -37,24 +37,36 @@ export const useLogin = () => {
         timeout: process.env.NODE_ENV === "production" ? 60000 : 0,
       });
 
-      if (data?.error) {
-        const error = data.error;
-        if (hasError("UserNotConfirmedException", error)) {
+      // TODO: are there other challengeState/states to check for? It looks like the rest are errors (API responses)
```
thiessenp-cds, line 40: @craigzour from what I can see in [...nextauth].ts this is the only success message. Would you know if this is right?

craigzour: And yes the `challengeState` can only have the `MFA` value associated to it for now.

thiessenp-cds: Got it - thanks!

cypress.config.ts
```diff
@@ -14,4 +14,12 @@ export default defineConfig({
       bundler: "webpack",
     },
   },
+  retries: {
```
thiessenp-cds, line 17: @bryan-robitaille thanks this helps!

lib/hooks/auth/useLogin.ts
```diff
@@ -37,24 +37,42 @@ export const useLogin = () => {
...
+    } catch (err: unknown) {
+      logMessage.error(err);
+
+      // 40X Client errors
+      if (axios.isAxiosError(err) && err.response && err.response.data) {
+        const { reason } = err.response.data;
+
+        // Handle slightly different response objects with the 400 check
+        if (
+          hasError(["UserNotFoundException", "NotAuthorizedException"], reason) ||
```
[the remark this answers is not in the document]
thiessenp-cds: I like it, definitely cleaner

timarney (approved):
From my testing works as advertised.

PR #2281, opened 2023-06-14
feature: deactivate user account

timarney:
# Summary | Résumé

 **Updates "Administration" top nav**

Screenshot 2023-07-11 at 7 35 34 AM

<hr>

**Adds Administration homepage**

Screenshot 2023-07-11 at 7 33 59 AM

<hr>

**Updates Accounts screen**

- Adds filters for "active" & "deactivated" users
- Adds toggle button for "locking / unlocking" publishing access
- Adds dropdown menu for "manage permissions" + "deactivate"
- Adds deactivate checks (when a form is published or has existing responses)

Screenshot 2023-07-11 at 7 36 50 AM

- Adds email for account that was deactivated 

Screenshot 2023-07-11 at 7 48 10 AM

<hr>

**Manage Forms**

- Adds manage forms page with links to responses + manage ownerships
- Adds dropdown "more menu" with delete + confirm delete modal

Screenshot 2023-07-11 at 7 40 25 AM

<hr>

**Form ownership**

- Moves form ownership under settings
- Adds check for at least one form owner

Screenshot 2023-07-11 at 7 43 45 AM

Adds emails for form ownership transfer

Screenshot 2023-07-11 at 7 49 12 AM

<hr>

**Manage permissions**

- Updates Manage permissions page

Screenshot 2023-07-11 at 7 46 38 AM

**Adds account deactivated screen**

- Adds page / check for deactivated when logging in
Screenshot 2023-07-11 at 7 50 37 AM

**Restricted access**

- Adds restricted access checks and page

Screenshot 2023-07-11 at 7 52 11 AM

**Download blocking**

- Adds checks and blocks for responses page when responses are over X days

Screenshot 2023-07-11 at 7 56 37 AM

<hr>

- Updates user schema to add an "active" status

- Adds restricted access screen 
- Adds check when logging in for deactivated user
- Adds API endpoint for activating  / deactivating a user

**My Forms**

Adds action required message under my forms

Screenshot 2023-07-11 at 7 53 46 AM

- Refer to test plan

- We'll need to determine if after being deactivated if the user is "auto logged out" / can only access the deactivated screen immediately or if that will happen upon the next login.

# Pull Request Checklist


components/admin/Users/ConfirmDeactivateModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+}: {
+  handleAddType?: (type?: FormElementTypes) => void;
+  handleClose: () => void;
+  user: DBUser;
+}) => {
+  const { t } = useTranslation("admin-users");
+  const dialog = useDialogRef();
+  const { refreshData } = useRefresh();
+
+  const { data, isLoading, error } = useSWR(`/api/account/${user.id}/forms`, fetcher);
```
bryan-robitaille, line 37: I'm wondering if the use of SWR (stale-while-revalidate) might cause unintended issues.
Thinking if a user loads the modal and get's a message that the user has existing forms.  The user changes ownership of forms in a different tab or by navigating where useSWR isn't called again.  This modal could potentially fetch stale data that a user still have forms while in the database fresh data has no forms associated.

components/globals/AdminNav.tsx
```diff
@@ -16,44 +16,50 @@ const AdminNav = (props: AdminNavProps): React.ReactElement => {
...
-          {t("title", { ns: "common" })}
+        <div className="mt-3 box-border block h-[40px] px-2 py-1 text-base font-bold">
+          {t("title", { ns: "admin-login" })}
         </div>
       </div>
       <nav className="justify-self-end">
         <>
-          <ul className="mt-2 px-0 flex text-base list-none">
+          <ul className="mt-2 flex list-none px-0 text-base">
             {user && user.name && (
```
bryan-robitaille, line 39: User will always be defined according to props.  No need to check for existance.
We should either modify the props to show that `user` can be undefined or simpifly the logic.


```suggestion
            {user.name && (
```

components/globals/AdminNav.tsx
```diff
@@ -16,44 +16,50 @@ const AdminNav = (props: AdminNavProps): React.ReactElement => {
...
-              <li className="mr-2 tablet:mr-4 py-2 text-sm pt-3">
+              <li className="mr-2 py-2 pt-3 text-sm tablet:mr-4">
                 {t("logged-in", { ns: "admin-login" })}: <span>{user.email}</span>
               </li>
             )}
 
+            {user && user.name && (
```
bryan-robitaille, line 45: User will always be defined according to props.  No need to check for existance.
We should either modify the props to show that `user` can be undefined or simpifly the logic.


```suggestion
            {user.name && (
```

components/globals/AdminNav.tsx
```diff
@@ -16,44 +16,50 @@ const AdminNav = (props: AdminNavProps): React.ReactElement => {
...
+              <li className="mr-2 py-2 text-base tablet:mr-4">
+                <Link href="/admin">{t("adminNav.administration")}</Link>
+              </li>
+            )}
+
             {(!user || !user.name) && (
```
bryan-robitaille, line 51: Impossible logic if the `user` is always defined as per prop typing.

lib/auth/cognito.ts
```diff
@@ -217,22 +231,29 @@ export const validate2FAVerificationCode = async (
       },
     });
 
+    // ensure the user account is active
+    const prismaUser = await prisma.user.findUnique({
+      where: {
+        email: email,
```
bryan-robitaille, line 237: This should probably be using the `sanitizedEmail` variable for consistency

lib/auth/cognito.ts
```diff
@@ -117,6 +117,20 @@ export const begin2FAAuthentication = async ({
   email,
   token,
 }: CognitoToken): Promise<AuthenticationFlowToken> => {
+  // ensure the user account is active
+  const prismaUser = await prisma.user.findUnique({
+    where: {
+      email: email,
```
bryan-robitaille, line 123: Should probably use the sanitizedEmail variable (declared immiediatley after) for consistency.

lib/types/form-types.ts
```diff
@@ -145,4 +145,7 @@ export type FormRecord = {
 };
 
 // defines the fields for the form record that is available to unauthenticated users
-export type PublicFormRecord = TypeOmit<FormRecord, "name" | "deliveryOption" | "bearerToken">;
+export type PublicFormRecord = TypeOmit<
+  FormRecord,
+  "name" | "deliveryOption" | "bearerToken" | "email"
```
bryan-robitaille, line 150: `email` does not exist on FormRecord and is a part of `deliveryOption` which is already ommited.

components/globals/AdminNav.tsx
timarney: Updated

components/globals/AdminNav.tsx
timarney: updated.

components/admin/Users/ConfirmDeactivateModal.tsx
timarney: @dsamojlenko can you validate this one?

lib/types/user-types.ts
```diff
@@ -1,5 +1,16 @@
+import { Privilege } from "@prisma/client";
+
 export interface FormOwner {
   id: string;
   email: string;
+  name?: string;
+  active: boolean;
+}
+
+export interface DBUser {
```
bryan-robitaille, line 10: FYI it is also possible to import types from Prisma for the defined models.
Would the Prisma User type meet your needs?
`import { User } from "@prisma/client";`

timarney: Almost I updated the code to extend the prisma type

```
export interface DBUser extends User {
  privileges: Privilege[];
}
```

components/admin/Users/ConfirmDeactivateModal.tsx
dsamojlenko: Screen Recording 2023-07-11 at 1 10 03 PM

Does this cover the scenario you're describing?

bryan-robitaille: It does.  Interesting, I guess multiple renders are causing the data to update.

timarney: @dsamojlenko is that running in Dev mode ? 

Just want to confirm same for a prod version.

lib/templates.ts
```diff
@@ -325,6 +328,67 @@ export async function getAllTemplates(
...
+ * Get all form templates for a specific User.
+ * @returns An array of Form Records
+ */
+export async function getAllTemplatesForUser(
+  ability: UserAbility,
+  userID: string,
+  isPublished?: boolean
+): Promise<Array<FormRecord>> {
+  try {
+    checkPrivileges(ability, [{ action: "view", subject: "FormRecord" }]);
```
bryan-robitaille, line 341: Note to myself to check into further because I have a gut feeling this doesn't do what we may intend it to check for.

lib/types/form-types.ts
timarney: Removed  `email`

lib/users.ts
```diff
@@ -96,12 +101,58 @@ export const getOrCreateUser = async ({
           name: true,
           email: true,
           privileges: true,
+          active: true,
```
bryan-robitaille, line 104: The active propery value only seems to be returned for a user who exists but did not have base privileges associated.  A few things we might want to do which this fn.  Update the return type to include `active: boolean` and update the other prisma calls to include `active` in the selection set.

lib/users.ts
```diff
@@ -96,12 +101,58 @@ export const getOrCreateUser = async ({
...
       .catch((e) => prismaErrors(e, null));
   }
 };
 
+/**
+ * Get User by id
+ * @returns User if found
+ */
+export const getUser = async (id: string, ability: UserAbility) => {
+  try {
```
bryan-robitaille, line 116: We should define the fn return type as it is currently set to any.

lib/users.ts
```diff
@@ -140,3 +192,99 @@ export const getUsers = async (ability: UserAbility) => {
...
   }
 };
+
+/**
+ * Update and overwrite existing User active status
+ * @param userID id of the user to be updated
+ * @param active activate or deactivate user
+ * @returns User
+ */
+export const updateActiveStatus = async (ability: UserAbility, userID: string, active: boolean) => {
```
bryan-robitaille, line 202: We should define the fn return type as it is currently set to `Promise<any>`.

lib/users.ts
```diff
@@ -116,6 +167,7 @@ export const getUsers = async (ability: UserAbility) => {
           id: true,
           name: true,
           email: true,
+          active: true,
```
bryan-robitaille, line 170: I know it's somewhat out of scope but since we're updating the function we could maybe also update the fn return type?

lib/users.ts
```diff
@@ -140,3 +192,99 @@ export const getUsers = async (ability: UserAbility) => {
...
+
+        const unprocessed = await detectOldUnprocessedSubmissions(allSubmissions.submissions);
+
+        if (unprocessed.level > 0) {
+          overdue[template.id] = unprocessed;
+        }
+      })
+    );
+  } catch (error) {
+    // noop
```
bryan-robitaille, line 286: Did we mean to add handling here?

lib/users.ts
timarney: Perhaps there's a better way to define the SelectedUser type but added now.

pages/admin/accounts/[id]/manage-forms.tsx
```diff
@@ -0,0 +1,237 @@
...
+    <AdminNavLayout user={page.props.user} backLink={<BackToAccounts />}>
+      <TemplateStoreProvider {...{ locale: page.props.locale }}>{page}</TemplateStoreProvider>
+    </AdminNavLayout>
+  );
+};
+
+export const getServerSideProps = requireAuthentication(
+  async ({ locale, params, user: { ability } }) => {
+    const id = params?.id || null;
+    checkPrivileges(ability, [{ action: "view", subject: "Setting" }]);
```
Comment (bryan-robitaille) [inline pages/admin/accounts/[id]/manage-forms.tsx:194]: To scope it down to the context of what's being retrieved in terms of information we could instead use:

```js
    checkPrivileges(ability, [
      { action: "view", subject: "User" },
      {
        action: "view",
        subject: {
          type: "FormRecord",
          // Passing an empty object here just to force CASL evaluate the condition part of a permission.
          // Will only allow users who have privilege of Manage All Forms
          object: {},
        },
      },
    ]);
```

lib/users.ts
```diff
@@ -96,17 +105,77 @@ export const getOrCreateUser = async ({
...
   }
 };
 
+/**
+ * Get User by id
+ * @returns User if found
+ */
+export const getUser = async (
+  id: string,
+  ability: UserAbility
```
bryan-robitaille, line 121: FYI we try to keep 'ability' as the firt argument as a generic pattern for our functions that require ability checks.

components/admin/Users/ConfirmDeactivateModal.tsx
dsamojlenko: Works same in PR Review environment

pages/api/account/[id]/forms.ts
```diff
@@ -0,0 +1,33 @@
...
+const getPublishedForms = async (
+  req: NextApiRequest,
+  res: NextApiResponse,
+  props: MiddlewareProps
+) => {
+  try {
+    const { session } = props as WithRequired<MiddlewareProps, "session">;
+    const ability = createAbility(session);
+
+    const accountId = req.query.id as string;
```
Comment (bryan-robitaille) [inline pages/api/account/[id]/forms.ts:17]: Instead of force typing, I know it's a pattern we've done elsewhere, we can also do a quick type check that throws an error instead.

```suggestion
    if (Array.isArray(req.query.id) || !req.query.id) {
      return res.status(400).json({ error: "Bad request" });
    }
    const accountId = req.query.id;
```

pages/api/account/submissions/overdue.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    Object.entries(overdue).forEach(([, value]) => {
+      if (value.level > 2) {
+        return res.status(200).json({ hasOverdueSubmissions: true });
+      }
+    });
+
+    return res.status(200).json({ hasOverdueSubmissions: false });
+  } catch (err) {
+    if (err instanceof AccessControlError) return res.status(403).json({ error: "Forbidden" });
```
bryan-robitaille, line 33: The access control error will not be caught because `getUnprocessedSubmissionsForUser` that has the various privilege checks does not bubble up the access control error.

lib/users.ts
```diff
@@ -140,3 +210,100 @@ export const getUsers = async (ability: UserAbility) => {
...
+      },
+    });
+
+    if (!active && user.email) {
+      sendDeactivationEmail(user.email);
+    }
+
+    return user;
+  } catch (error) {
+    logMessage.error(error as Error);
```
bryan-robitaille, line 246: Need to add a logEvent for the AccessControlError and rethrow (error) to bubble up the scope.

components/admin/Users/ConfirmDeactivateModal.tsx
bryan-robitaille: I think I know what's going on, but it doesn't matter we can leave as is.
If you're curious though...
The tab switching is causing the Next Auth session context to update which rerenders all children including the modal.  Next Auth updates that Session Context Provider when it detects tab switching to ensure the auth status of a user is consistent across tabs.

pages/form-builder/responses/[[...params]].tsx
```diff
@@ -333,13 +336,9 @@ export const getServerSideProps: GetServerSideProps = async ({
       FormbuilderParams.initialForm = initialForm;
       vaultSubmissions.push(...allSubmissions.submissions);
 
-      const isNagwareEnabled = await checkOne("nagware");
```
Comment (bryan-robitaille) [inline pages/form-builder/responses/[[...params]].tsx:336]: If we're remove the flag check we should also remove the "nagware" flag from initialization.

Comment (bryan-robitaille) [reply pages/form-builder/responses/[[...params]].tsx:336]: Sorry, yes I meant we should remove the flag completely from the app.

.eslint-tailwindcss.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
     ],
     settings: {
       tailwindcss: {
-        whitelist: ["(gc\\-).*"]
+        whitelist: ["(gc\\-).*", "form-builder", "page-container", "visually-hidden"]
```
bryan-robitaille, line 19: Does Tailwinds built in class of `sr-only` not meet the need for `visually-hidden`?

timarney: This update isn't really in the scope of this PR but the update is more around getting rid of the lint issue vs needing the class.
Screenshot 2023-07-12 at 11 11 38 AM

pages/api/account/submissions/overdue.ts
timarney: @dsamojlenko is going to write some tests / fix this up as needed.

pages/form-builder/responses/[[...params]].tsx
Comment (timarney) [reply pages/form-builder/responses/[[...params]].tsx:336]: Removed.
https://github.com/cds-snc/platform-forms-client/pull/2281/commits/4a97f1c9f6f7c37df0e19999d3aa60998dc77fe6

.eslint-tailwindcss.js
bryan-robitaille: Fair point.  I'll open a seperate issue that we can tackle later on.

lib/templates.ts
bryan-robitaille: Since we're allowing templates to be pulled for any user and not just the user identifed on the Ability we should also check that the user has the "View User" privilege.

bryan-robitaille (approved):
👍

PR #2285, opened 2023-06-14
fix: manually handle logout redirect for PR Review environments

dsamojlenko:
# Summary | Résumé

To save ourselves from having to add all PR Review environment URLs to Cognito as allowed Callback URLs dynamically, we just use the Staging URL for the NEXTAUTH_URL in the Review environments. 

For the most part, this doesn't affect anything visibly until you try to Logout, when NextAuth will redirect you using the NEXTAUTH_URL from the environment as the baseUrl, which would be the Staging URL.

This PR modifies the NextAuth `redirects` callback to return a relative path without baseUrl, so it will work in any environment.

# Pull Request Checklist

### DIFF
--- a/components/auth/LoginMenu.tsx
+++ b/components/auth/LoginMenu.tsx
@@ -14,7 +14,7 @@ const LoginMenu = ({ isAuthenticated }: LoginMenuProp) => {
   const handleClick = () => {
     clearTemplateStore();
     signOut({ redirect: false, callbackUrl: "/" });
-    window.location.href = "/${i18n.language}/auth/logout";
+    window.location.href = `/${i18n.language}/auth/logout`;
   };
 
   return (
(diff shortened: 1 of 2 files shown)


components/auth/LoginMenu.tsx
```diff
@@ -13,7 +13,8 @@ const LoginMenu = ({ isAuthenticated }: LoginMenuProp) => {
   const { i18n, t } = useTranslation("common");
   const handleClick = () => {
     clearTemplateStore();
-    signOut({ callbackUrl: `/${i18n.language}/auth/logout` });
+    signOut({ redirect: false, callbackUrl: "/" });
```
timarney, line 16: Can we use `${window.location.origin}/${i18n.language}/auth/logout` ?

dsamojlenko: Sadly, that didn't work ... nextAuth removes the domain prefix and replaces it with the base url set in NEXTAUTH_URL in the environment

timarney: Just adding this here per Slack discussion https://next-auth.js.org/configuration/callbacks#redirect-callback

timarney (approved):

PR #2290, opened 2023-06-15
ci: Update token on release generator

bryan-robitaille:
# Summary | Résumé
Updates the token used on the Release Generator action to personal token so that CI tests and other actions downstream can run.  
See https://github.com/google-github-actions/release-please-action#github-credentials for details


thiessenp-cds (approved):
LGTM :) Would the test be doing a test release? I'll approve regardless

PR #2294, opened 2023-06-16
feat: add tailwindcss eslint config

dsamojlenko:
# Summary | Résumé

Adds an ESLint package and config to help managing tailwindcss classes. Currently the config sets most of the violations as warnings so they won't block commits. The exception being conflicting classes - ie, if you have `mb-2 mb-4` you will get an error due to conflicting classes.

You can [read about the rules on the plugin docs](https://github.com/francoismassart/eslint-plugin-tailwindcss/tree/master#supported-rules)

Running this lint against the entire site obviously turns up a *lot* of issues, so will keep these as warnings and hopefully we can address the issues gradually over time.

To make it easier, this is being added as a separate custom eslint config, so nothing changes with the regular `yarn lint` - it will just run our original rules without the noise generated by this.

### To use
Pull this down and run `yarn install` to get the new dependency, then if you want to integrate with VS Code, follow the instructions below.

### Husky pre-commit hook
This new config has been added to our pre-commit hook, but since most are warnings they won't block the commit. If you have other linting errors, then you will see the warnings along with them.

The one rule that is set to Error is conflicting classes, ie `mb-2 mb-3` will throw such an error and block the commit.

### VS Code
Highly recommend adding this new eslint config to your VS Code Workspace settings, so you get the warning highlights and the "Quick fix" option (see screen recording below).

You will need to have the [VS Code ESLint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) installed.

You can configure ESLint to use the tailwindcss linting config by creating (or updating) your `.vscode/settings.json` file (this is .gitignored as devs can have their own customizations) with the following entry:

```
{
    "eslint.options": { "overrideConfigFile": ".eslint-tailwindcss.js" }
}
```

Screen Recording 2023-06-16 at 11 08 09 AM

Note that in the screen recording, the final state shows a warning that the `gc-input-text` is not a Tailwindcss class. That has since been fixed by adding an exception to that rule for any `gc-`-prefixed classes.

### Yarn command
To run this linter on the command line, a new Yarn command has been added to make it easier: 
`yarn lint:tailwindcss [DIR | FILE] [ARGS]`

Since the tailwindcss lint config extends the base config, this will run all of our linters against whatever file(s) or directories you've specified.

To use it, make sure you specify one or more directories or files to scan:

```
yarn lint:tailwindcss .
yarn lint:tailwindcss components/admin/
yarn lint:tailwindcss components/admin/FormAccess/FormAccess.tsx
```

This command will also pass arguments through to ESlint, so you can for example, auto fix errors:

```
yarn lint:tailwindcss components/admin/FormAccess/FormAccess.tsx --fix
```


timarney:
Tested locally works well -- just need to ensure the eslint extension is installed --- I didn't have it installed.

timarney (approved):

PR #2309, opened 2023-06-20
docs: add info about local testing

daine:
# Summary | Résumé
Updated the readme to include information about local dev testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,31 @@ There are 2 ways to connect to the database. Either directly using PGAdmin or a 
 - A table will appear. Find your username in the list and double-click on the value under the `role` column to modify to "ADMINISTRATOR".
 - Click on "Save Change" button in the top menu bar once completed.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,6 +108,31 @@ There are 2 ways to connect to the database. Either directly using PGAdmin or a
...
+```
+
+If a test is failing, you can run the test tool:
+```
+yarn dev:test
+yarn cypress
+# A chrome instance starts, then manually start and watch that test running
+```
+
+The application also uses Jest for unit testing. To run the tests, run `yarn test:unit`.
```
bryan-robitaille, line 134: A little mix-up in commands, it should be just `yarn test`.

bryan-robitaille (changes requested):
Small little change for the command to run jest tests.

bryan-robitaille (approved):
:+1

PR #2328, opened 2023-06-23
Adds initial work for restricting downloads

thiessenp-cds:
# Summary | Résumé

This work adds the Client side logic to restrict a download. A submission download is restricted if the user account is in a "escalated" state and the submission marked as new.

## Testing

IMPORTANT: because a placeholder value is use for "escalated" and set to true, the Download table will always show a warning and block any new downloads.

At this point I think the only thing that can be tested is whether non-new submissions are not blocked in the Client.

## Future Work

The warning text may need to be updated and eventually the French translation added.

This work relies on a placeholder value for the account "escalated" state. This will need to be done.

The download API should also use the account "escalated" state to block a submission download at the API level. This can be down once the variable is created (or now :).


timarney (approved):
Approving -- we'll adjust further once we get more pieces in place.

PR #2367, opened 2023-07-07
feat: Delete test responses on publish

### LINKED ISSUE #2337: As soon as we connect to a database, and the form is going from Draft to Publish -> empty the vault.
**User story:**
As a forms administrator
I need to clear all test responses as soon as I publish my form
So that I don't get an email reminding me to clear old test data

**Story context**
When users are testing their form, they are able to create test responses/test submissions using the "Preview" tab. These submissions are test data when the form is still in draft or unpublished.

**_Unresolved_**:
Question: How do we want to communicate this action to our users? See red text below:
Screenshot 2023-06-28 at 11 17 32 AM

1) I wonder if we could or should update the Responses tab for unpublished forms to something like "Preview responses" or "View Test Responses?

2) We can probably add a logic check if there are any responses and only show the navigation menu on the left (the Responses tab) if there are any. But thinking about this more, I think it's best left alone as is?

3) For messaging about responses, I wonder if we would want to do a logic check to only show the warning message? There's already one warning message on the screen and I wonder if a user will be more inclined to ignore warnings if there are too many of them

4) Is there a preferred design for displaying this new warning message?

daine:
# Summary | Résumé

fixes #2337 
**New UI for warning:**
|                                          | Screenshot                                        |
| English |Screenshot 2023-07-07 at 3 16 00 PM|
| French |Screenshot 2023-07-07 at 3 16 14 PM |

Submit test responses
Publish form
Responses page should not have anything on it

- [ ] Confirm: New design incorporates rounded corners and a new UI component for "What you should know"
- [ ] Tablet designs
- [ ] A11y considerations

# Pull Request Checklist


lib/vault.ts
```diff
@@ -184,3 +311,125 @@ export async function numberOfUnprocessedSubmissions(
...
+      throw new Error(`Failed to delete form responses from DynamoDB.`);
+    }
+  }
+  return { responsesDeleted };
+}
+
+export async function deleteDraftFormTestResponses(ability: UserAbility, formID: string) {
+  try {
+    // TODO: Verify that the form status is not Published (must be draft)
+    // const templateWithAssociatedUsers = await _unprotectedGetTemplateWithAssociatedUsers(formID);
```
bryan-robitaille, line 406: Even though it will result in another call to the Database it would be a good idea to check.  Instead of using a more complex call to functions in `Templates` a simple prisma call to the db could be used instead.
For example:
```js
  const isPublished = await prisma.template
    .findUnique({
      where: {
        id: formID,
      },
      select: {
        isPublished: true
        },
      },
    })
    .catch((e) => prismaErrors(e, true));
```

lib/vault.ts
```diff
@@ -163,6 +191,105 @@ export async function listAllSubmissions(
...
+        },
+        ProjectionExpression:
+          // "FormID,#status,SecurityAttribute,#name,CreatedAt,LastDownloadedBy,ConfirmTimestamp,DownloadedAt,RemovalDate",
+          "FormID,#status,#name,ConfirmationCode",
+      };
+      const queryCommand = new QueryCommand(getItemsDbParams);
+      // eslint-disable-next-line no-await-in-loop
+      const response = await documentClient.send(queryCommand);
+
+      if (response.Items?.length) {
```
bryan-robitaille, line 240: This parsing function will need to be modified to only address the information you're returning from the dynamodb call. (FormID, Status, Name, ConfirmationCode)

lib/vault.ts
```diff
@@ -163,6 +191,105 @@ export async function listAllSubmissions(
...
+              confirmedAt: confirmedAt ?? null,
+              downloadedAt: downloadedAt ?? null,
+              removedAt: removedAt ?? null,
+              confirmationCode: confirmationCode ?? null,
+            })
+          )
+        );
+      }
+      // We either manually stop the paginated request when we have 10 or more items or we let it finish on its own
+      if (accumulatedResponses.length >= 500) {
```
bryan-robitaille, line 270: As mentioned above the limit should be removed.

bryan-robitaille (commented):
I did a quick review after seeing you had posted the PR in Slack.  Noticing now that it's still in draft and there's quite a few cleanup items to address.  Hopefully these comments can provide some guidance while you continue to refactor.

components/form-builder/store/useTemplateStore.tsx
```diff
@@ -377,6 +378,10 @@ const createTemplateStore = (initProps?: Partial<InitialTemplateStoreProps>) =>
             getIsPublished: () => get().isPublished,
             getName: () => get().name,
             getDeliveryOption: () => get().deliveryOption,
+            isDeliveryOptionVault: () => {
```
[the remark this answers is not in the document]
daine: I was looking elsewhere and the only way we've been checking if the delivery option is via email or sent to the vault is doing a logic check every single time (i.e. if it's set, it's an object containing the email info, if it's undefined that means it's sent to the vault). I was hoping to reduce errors by having a helper functionality in the store itself. 

I don't mind renaming it to `getIsDeliveryOption` but it sounds a little off considering it's not an actual data attribute in the model itself so we're not doing a get command to retrieve the value of an attribute. What do you think?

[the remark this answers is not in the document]
bryan-robitaille: @timarney any thoughts on including helper functions in the store vs separate hooks?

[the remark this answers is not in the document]
timarney: What about adding a utility function where you can pass the value from the store vs adding the function to the store itself.

parseDeliveryOption(deliveryOption){
 // check for  emailAddress
 return DeliveryOption.email 
 
...
 
 return DeliveryOption.vault
 
}

This could be re-used in the /components/form-builder/app/SetResponseDelivery.tsx

timarney:
|  current |  design |  
|---|---|
|  No rounded corners Screenshot 2023-07-12 at 7 25 08 AM | Rounded corners Screenshot 2023-07-12 at 7 23 41 AM  |  

Looks like we don't have tablet designs ?

Screenshot 2023-07-12 at 7 22 24 AM

components/form-builder/app/Publish.tsx
```diff
@@ -85,83 +87,102 @@ export const Publish = () => {
...
+              <InfoCard title={t("whatYouNeedToKnow")}>
+                <div className="mb-5 bg-gray-50 p-1.5">
+                  <p className="mb-1 font-bold pb-0">{t("publishingDisablesEditing")}</p>
+                  <p className="text-sm">{t("publishingDisablesEditingDescription")}</p>
+                </div>
+                <div className="mb-5 bg-gray-50 p-1.5">
+                  <p className="mb-1 font-bold pb-0">{t("publishingLocksSettings")}</p>
+                  <p className="text-sm">{t("publishingLocksSettingsDescription")}</p>
+                </div>
+                {isVaultDelivery() && (
```
timarney, line 168: Looks like your not passing the delivery option here so this check is not working when email delivery is set.

components/form-builder/util.ts
```diff
@@ -251,3 +251,11 @@ export const allowedTemplates = [
   FormElementTypes.firstMiddleLastName,
   FormElementTypes.contact,
 ] as const;
+
+export const isVaultDelivery = (deliveryOption?: { emailAddress?: string } | null) => {
```
timarney, line 255: deliveryOption should not be an optional parameter --- without it your not doing any sort of check.

lib/templates.ts
```diff
@@ -563,6 +564,11 @@ export async function updateIsPublishedForTemplate(
           field: "isPublished",
         },
       ]);
+
+      if (templateWithAssociatedUsers.formRecord.isPublished)
```
timarney, line 568: Nit: I don't think we drop the {} for if statements elsewhere but could be wrong
 
 ```javascript
 if (templateWithAssociatedUsers.formRecord.isPublished) {
        throw new TemplateAlreadyPublishedError();
 }
 ```

components/form-builder/util.ts
daine: good catch!

lib/templates.ts
daine: It occurs on line 739 under the existing `removeDeliveryOption` function but happy to change it to always have {}. I don't think the linter catches it.

components/form-builder/app/Publish.tsx
daine: fixed

components/form-builder/util.ts
daine: Changed it to 
```javascript
export const isVaultDelivery = (deliveryOption: DeliveryOption | undefined)
```

let me know if that's better or not @timarney

craigzour (changes requested):
@daine I have been looking the changes and it does look good to me outside of a few similar things that will find below. I am going to test the feature tomorrow morning.

lib/vault.ts
```diff
@@ -184,3 +283,131 @@ export async function numberOfUnprocessedSubmissions(
...
 }
+
+/**
+ * This method deletes all responses for a given form
+ * @param ability
+ * @param dynamoDb - DynamoDB Document Client
+ * @param formResponses - List of form submissions and confirmations
+ * @param formID
+ */
+export async function deleteResponses(
```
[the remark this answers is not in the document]
daine: Yeah I don't mind, I was thinking it could be used somewhere else in the future but I can just easily remove the export for now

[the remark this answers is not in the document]
craigzour: The main reason is to avoid having a function that should be used very carefully be available for any developer to call while it was originally created for a specific feature.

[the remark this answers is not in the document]
daine: I think this function is imported and used by the test suite. Would you be okay with keeping the export for this function and removing it for the others? @craigzour

lib/vault.ts
```diff
@@ -35,9 +39,51 @@ async function getUsersForForm(formID: string) {
   return templateOwners?.users;
 }
 
+export async function checkAbilityToAccessSubmissions(ability: UserAbility, formID: string) {
```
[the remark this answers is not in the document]
daine: done

lib/vault.ts
```diff
@@ -162,6 +191,76 @@ export async function listAllSubmissions(
...
   }
 }
+/**
+ * This method returns a list of all form submission and confirmations records.
+ * The list does not contain the actual submission data, only attributes
+ * @param ability - The user ability object
+ * @param formID - The form ID from which to retrieve responses
+ * @returns {Promise<{submissions: VaultSubmissionAndConfirmationList[]}>} - The list of submissions and confirmations
+ */
+export async function listAllSubmissionsAndConfirmations(
```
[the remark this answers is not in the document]
daine: done

craigzour (approved):
Completed testing using my local environment.

PR #2402, opened 2023-07-13
Account page e2e tests

thiessenp-cds:
# Summary | Résumé

Adds cypress testing util functions to login (2fa) a user. Also adds a test admin user `testadmin.user@cds-snc.ca` and DB session related clean up functions. e.g.
```
cy.login(true, true);
...
cy.resetAll();
```

Some basic cypress e2e tests are also added for the Accounts page.

## Test

Run the cypress e2e tests. Should succeed and also not slow down the tests.

## Future Work

Replace the Account buttons to filter active/deactivated accounts with a tab interface. The current implementation works but it would be clearer for assistive technology users as tabs. With tabs you get a semantic list of button controls, the expected behavior for activating a tab button etc.
I put this off because the current Tabs component will need to be refactored to be more reusable e.g. use the Button component.


timarney:
Merged updates from the deactivate account so possibly not passing now but test is failing.

Screenshot 2023-07-14 at 7 52 00 AM

Outside of that 💯 .

thiessenp-cds:
> I merged updates from the deactivate account so possibly test is failing now because of that but seeing test fail.
> 
> Screenshot 2023-07-14 at 7 52 00 AM
> Outside of that 💯 .

Actually that makes sense. Will fix. Thanks for testing that out!

timarney (approved):
@thiessenp-cds going to approve this so you can merge and either fix up here or in the deactivate branch --- CI doesn't run on this branch.

PR #2410, opened 2023-07-17
Stop sending emails using Notify when running in the test environment

thiessenp-cds:
# Summary | Résumé

Our cypress tests were sending a lot of emails using notify and our test accounts were marked as spam/blocked by notify. This is fixed by adding a wrapper around the notify client to return a mock if the app is being run from the test environment.

### Verify no emails are being sent when in a testing environment 
Run the unit tests and the cypress tests. 
Then login to the Notify dashboard. Currently the test.user account is flagged as a deactivated account.  So click on the failed tab. No new, emails should have been blocked from test.user
Screenshot 2023-07-20 at 2 21 05 PM

### Verify emails are still being sent from the app 
From a browser, try a few actions that require reaching notify, like 2FA, account creation etc.


lib/auth/2fa.ts
```diff
@@ -9,7 +9,10 @@ export const generateVerificationCode = async () => generateTokenCode(5);
 
 export const sendVerificationCode = async (email: string, verificationCode: string) => {
   try {
-    const notify = new NotifyClient("https://api.notification.canada.ca", NOTIFY_API_KEY);
+    // Avoids test accounts being blocked by Notify
+    const notifyUrl =
```
timarney, line 13: Just a thought ...

What if we wrap the Notify Client 



So instead of pulling in

```
import { NotifyClient } from "notifications-node-client";
```


We 
import { NotifyClient } from "notify";

```
// notify.ts

const mockNotify = {
   sendEmail: () =>{
    log("email sent ...")
   }
}

export const  NotifyClient = () =>{

  if(process.env.APP_ENV === "test"){
   ... return  mockNotify;
  }
  
  else 
  .. 
  
  return  new NotifyClient("https://api.notification.canada.ca", NOTIFY_API_KEY);
  
  
  
}

const notify = new NotifyClient("https://api.notification.canada.ca", NOTIFY_API_KEY);
```

thiessenp-cds: Looks good IMO. That would remove all the if-statements and keep our code a bit cleaner. We'd just need to remember to use the mock in tests

timarney: >  mock in tests

If we update the import to use out wrapper it should figure itself out based in the env

thiessenp-cds: Oh I see, notify.js would be used in place globally, got it. It could also be a convenient place to add any notify specific customization in the future

thiessenp-cds:
Running the unit tests locally works. Running the cypress tests, 2 fail but I don't think the fail is email related. Screenshots below if it helps
Screenshot 2023-07-20 at 2 14 47 PM
Screenshot 2023-07-20 at 2 15 50 PM

thiessenp-cds:
This PR seems to be having problems after the base branch was changed (a few times).  I'm a bit worried other changes snuck in. I'm going to recreate this PR on a new branch.

Issue #2412, opened 2023-07-18
Security Questions - Add Account Dropdown to header navigation


dsamojlenko:
<img width="619" alt="image" src="https://github.com/cds-snc/platform-forms-client/assets/1187115/28ccef7a-b9f8-4e11-81de-4dade8b65e46">

<img width="674" alt="image" src="https://github.com/cds-snc/platform-forms-client/assets/1187115/c84b3e21-04df-46fb-8e88-a3219a6ec2ce">

https://www.figma.com/file/89fRce4gmy64UIArSZoAM8/v3.-GC-Forms-(builder-and-vault)?type=design&node-id=271-33265&mode=design&t=P2sX4NNyOlzHMlD5-0

samsadasivan:
Thank you Dave!

timarney:
Merged.

PR #2427, opened 2023-07-20
Combine admin/formBuilder Headers and add Your Account dropdown

dsamojlenko:
# Summary | Résumé

See issue: #2412 

Adds a YourAccount dropdown to the Header.
Also combines the old `form-builder/app/navigation/Header.tsx` and `globals/AdminNav.tsx` into a new `globals/Header.tsx` for simplicity since the navs are largely the same now.

The new `globals/Header.tsx` accepts an optional `context` param that determines if the nav is appearing in one of: "admin", "formBuilder", or "default" modes.

Also includes a reversion to a previous version of the Radix Dropdown component. The latest point release is causing issues.

| Logged in as Admin                                          | Logged in as other                                        |
| image | image |

| Signed in Default context                                         |
| ----------------------------------------------- |
| image |

| Signed out Default context                                         |
| ----------------------------------------------- |
| image |

| Signed in FormBuilder context                                         |
| ----------------------------------------------- |
| image |

| Signed out FormBuilder context                                         |
| ----------------------------------------------- |
| image |

| Admin context                                         |
| ----------------------------------------------- |
| image |


timarney:
Let's set a min width for the dropdown to bring it inline with Figma where we can see it's wider vs the "your account" trigger.

254314145-28ccef7a-b9f8-4e11-81de-4dade8b65e46

timarney (approved):

PR #2460, opened 2023-07-27
Setup security questions

thiessenp-cds:
# Summary | Résumé

Adds UI for setting up security questions.

# Test

Not much to test yet other than whether the basic form validation works.

# Future work

 - Add the api calls
 - Add test for new Select component and more work on this and similar components (probably during the portage time)


timarney (approved):
Approving -- had to remove the redirect to the page as that code is causing a bug but that's outside of this issue.

PR #2484, opened 2023-08-01
Adds unknown error to profile dialog plus loader

thiessenp-cds:
# Summary | Résumé

Adds an "unknown error" to the profile dialog. 

Also adds the start of a "loader" for the create security questions page. I was noticing a small delay, even locally, when submitting the form. Currently this just disables the submit button as an indicator when submitting.

## Future work

Consider a "submitting" (previously called "loading") pattern for all/most of our forms.

### DIFF
--- a/components/admin/Profile/EditSecurityQuestionModal.tsx
+++ b/components/admin/Profile/EditSecurityQuestionModal.tsx
@@ -11,10 +11,6 @@ import { getCsrfToken } from "next-auth/react";
 import { useRouter } from "next/router";
 import debounce from "lodash.debounce";
 
-// TODO: Dialog component currently takes actions to control the dialog behavior. Would be nice to
-// be able to wire with a form element to work with onSubmit and button type=submit to get form
-// behaviors like keying enter in an input to submit the form.
-
 const updateSecurityQuestion = async (
   oldQuestionId: string,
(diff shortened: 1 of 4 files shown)


samsadasivan:
Will start thinking of a "loading" pattern for forms

timarney:
Thinking in a case where questions don't exist 

We should just display the error and not the form -- given you shouldn't be able to submit without questions.

Currently you can end up in this state.
Screenshot 2023-08-01 at 3 02 32 PM

thiessenp-cds:
> Thinking in a case where questions don't exist
> 
> We should just display the error and not the form -- given you shouldn't be able to submit without questions.
> 
> Currently you can end up in this state. Screenshot 2023-08-01 at 3 02 32 PM

Good point. Instead, I'll have it show a dialog with just the title and error

thiessenp-cds:
> > Thinking in a case where questions don't exist
> > We should just display the error and not the form -- given you shouldn't be able to submit without questions.
> > Currently you can end up in this state. Screenshot 2023-08-01 at 3 02 32 PM
> 
> Good point. Instead, I'll have it show a dialog with just the title and error
Screenshot 2023-08-01 at 3 21 29 PM

PR #2485, opened 2023-08-02
Adds on blur behavior to security questions

thiessenp-cds:
# Summary | Résumé

Adds on blur behaviour to the Setup Security Questions form. Now validation will run when tabbing or clicking out of a field and only update the validation for that one field.

## Test

Go to the setup security form and try clicking on a select or input. Then click outside that field and only that field's validation should be updated.
Screenshot 2023-08-02 at 8 42 52 AM

### DIFF
--- a/components/globals/Input/Input.tsx
+++ b/components/globals/Input/Input.tsx
@@ -1,0 +1,97 @@
+import React, { useState } from "react";
+import classnames from "classnames";
+import { useField } from "formik";
+import { ErrorMessage } from "@components/forms";
+import { InputFieldProps, CharacterCountMessages, HTMLTextInputTypeAttribute } from "@lib/types";
+
+/**
+ * Note: Copy+Past and edit from forms/Input. This is to allow us to customize inputs wihout worry
+ * of modifying the look and field of existing form-forms.
+ *
+ * TODO: clean up below, add a unit test, and then remove this and above comment :)
+ */
+
+export interface InputProps extends InputFieldProps {
+  type: HTMLTextInputTypeAttribute;
+  characterCountMessages?: CharacterCountMessages;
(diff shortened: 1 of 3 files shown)


components/globals/Input/Input.tsx
```diff
@@ -0,0 +1,97 @@
+import React, { useState } from "react";
```
timarney, line 1: Is this a straight copy or?

Asking as I would rather make these types of edits outside of the current PR.

thiessenp-cds: Yup, the only edit is on line 76 adding the meta.touched. Makes sense, though first, here was my logic. 
I'd like to add on blur validation. I need to add a dirty (touched) check to the form components. I need to edit the Input component. I'm afraid to modify the forms-form components. I guess I should "clone" the input and start customizing it. Maybe I should do this later? Nah, I need it now :)

But yes, this small change could be added to the forms-form TextInput component instead if you'd prefer?

timarney: Wondering if we use the existing TextInput and update the `ErrorMessage` to take a meta prop

We could check for meta.touched at that level and possibly do something with the `validateOnBlur`

thiessenp-cds: Sounds good. I think I'm over my fear of touching forms-form components now. Adding the meta.touched field should really only impact form behavior if the validateOnBlur is set to true.

Though in the future, maybe over the Portage. I'd (/whoever) really like to add basic elements like Select / Input / .. that we can customize for the app. I'm worried one day we'll accidentally break the forms-form components

thiessenp-cds:
Moving work to Portage time since it is not needed to complete the security questions work.

PR #2496, opened 2023-08-08
Verify email before showing security questions

timarney:
# Summary | Résumé

Updates the security questions flow to include a link to verify by email before showing security questions 

- Security questions reset flow

# Pull Request Checklist


lib/auth/passwordReset.ts
```diff
@@ -0,0 +1,104 @@
...
+  });
+
+  if (doesUserExist === null) throw new Error("UserDoesNotExist");
+
+  const generatedToken = await generateVerificationCode();
+
+  const dateIn15Minutes = new Date(Date.now() + 900000); // 15 minutes (= 900000 ms)
+
+  try {
+    await prisma.magicLink.upsert({
```
bryan-robitaille, line 26: Just noting so we can create an issue and potenially take on as part of tech debt.  There is a small chance that the unique contraint is thrown if a duplicate token exists for another user.  In the future we should probably handle that case in the code instead of breaking the user flow by generating an error.

timarney: @craigzour let's create an issue for this now so it doesn't get lost.

craigzour: Sure I will do it

lib/auth/passwordReset.ts
```diff
@@ -0,0 +1,104 @@
...
+export const sendPasswordResetLink = async (email: string): Promise<void> => {
+  const sanitizedEmail = sanitizeEmailAddressForCognito(email);
+
+  const doesUserExist = await prisma.user.findUnique({
+    where: {
+      email: sanitizedEmail,
+    },
+  });
+
+  if (doesUserExist === null) throw new Error("UserDoesNotExist");
```
bryan-robitaille, line 19: Noticed that this results in throwing a 500 back to the client.  I'm wondering if we should just log that a password reset was requested with a wrong email address and do an early return?

craigzour: I could create a custom error to be treated differently on the API side. I did it for the other lib function because I knew we would need to handle it on the front end side (expired or invalid link).
Was not sure it would be necessary because the UI would not need to know exactly what happened. Maybe we want to show something special when the `UserDoesNotExist ` error happen though.

lib/auth/passwordReset.ts
craigzour: Will add some context but here it is https://github.com/cds-snc/platform-forms-client/issues/2498

lib/auth/passwordReset.ts
bryan-robitaille: Similar to our login we should try to keep information on whether or not a user exists internal as much as possible.  I wouldn't do any special handling... just a graceful fail and return a 200 like everything is normal.

craigzour: Sure thing! Will change it

bryan-robitaille (approved):

Issue #2476, opened 2023-07-31
Updating the "Contact Us" Form


Abi-Nada:
### Description:

Update the contact us form to improve information quality collected via it.

### Acceptance Criteria:
- Add the following fields:
    - [x] Department
    - [x] Branch
    - [x] Job title
- [x] Update the corresponding French form
~~- [ ] Update the link to the Support Form and the Contact form link within it in tandem~~
- [x] Update link on the 'Contact Us' button in marketing page
~~- [ ] Ensure all links are updated across the product~~

connorscarolyns:
The link also needs to be updated on the get support form, and once that is updated, that form will have a new link and need to be updated on the contact us form... (it's a vicious cycle) :
![image.png](https://images.zenhubusercontent.com/620427f3846ebb620d78e582/d8043422-d23b-4ba3-a920-3c197ddab8ac)
![image.png](https://images.zenhubusercontent.com/620427f3846ebb620d78e582/505811f9-4bf4-4c21-b79f-b2c0dfde7b7d)

jose-cds:
Updated version of the form. Sent to Anik for feedback and review

jose-cds:
```
{
  "titleEn": "Contact us",
  "titleFr": "Contactez l'équipe de Formulaires GC",
  "introduction": {
    "descriptionEn": "Use this form to learn more about GC Forms and how it can work in your department.\n\nWe provide services to people who work in the government. We cannot give general advice to the public about the forms that we host. ",
    "descriptionFr": "Utilisez ce formulaire pour en savoir plus sur Formulaires GC et sur la façon dont votre département peut utiliser ce service.\n\nNous fournissons des services aux personnes qui travaillent au sein du gouvernement. Nous ne pouvons pas donner de conseils généraux au public sur les formulaires que nous hébergeons. Pour toute autre question ou pour tout problème technique, [veuillez contacter notre équipe de soutien.](https://forms-formulaires.alpha.canada.ca/fr/form-builder/support)"
  },
  "privacyPolicy": {
    "descriptionEn": "Learn about your [privacy and the Canadian Digital Service](https://digital.canada.ca/legal/privacy/).",
    "descriptionFr": "En savoir plus sur votre [vie privée et le Service numérique canadien](https://numerique.canada.ca/transparence/confidentialite/)."
  },
  "confirmation": {
    "descriptionEn": "Thank you for contacting the GC Forms team. We will reply to you via email within the next 3 business days.",
[...]

PR #2504, opened 2023-08-09
Fix/contact us form

thiessenp-cds:
# Summary | Résumé

Adds several new fields to the Contact Us and Support forms. The related API is also updated with the new fields.

For both the support and contact us forms:
 - Verify the new fields show in the form and are required: Department, Branch, Job Title
 - Verify the new fields show in the related email

Note: the Marketing contact us link update will be done outside this PR


__tests__/api/request/support.test.ts
```diff
@@ -172,6 +178,9 @@ function runEmailAPITests() {
         supportType: "support",
         name: "name",
         email: "email@email.com",
+        department: "Department",
+        branch: "Branch",
+        jobtitle: "Job Title",
```
daine, line 183: A bit of a nit: the same two word variable above (`supportType`) uses camel case format and `jobtitle` is all lowercase

pages/api/request/support.ts
```diff
@@ -5,9 +5,9 @@ import { logMessage } from "@lib/logger";
 
 // Allows an authenticated or unauthenticated user to send an email requesting help
 const requestSupport = async (req: NextApiRequest, res: NextApiResponse) => {
-  const { supportType, name, email, request, description } = req.body;
+  const { supportType, name, email, request, description, department, branch, jobtitle } = req.body;
 
-  if (!name || !email || !request || !description) {
+  if (!name || !email || !request || !description || !department || !branch || !jobtitle) {
     return res.status(404).json({ error: "Malformed request" });
```
daine, line 11: Not an issue with your PR, but is there a reason we're doing 404s for this error? Seems like 400 would be more appropriate, no?

timarney:
Posted the link here for review outside the devs

https://github.com/cds-snc/platform-forms-client/issues/2476#issuecomment-1673069766

pages/api/request/support.ts
thiessenp-cds: Missed that, good point

__tests__/api/request/support.test.ts
thiessenp-cds: Sure, sounds good

__tests__/api/request/support.test.ts
```diff
@@ -105,6 +105,9 @@ function runEmailAPITests() {
         supportType: "support",
         name: "name",
         email: "email@email.com",
+        department: "Department",
+        branch: "Branch",
+        title: "Job Title",
```
daine, line 110: Would it be helpful if we put back the old name/variable but just did it in camel case? 
```suggestion
        jobTitle: "Job Title",
```

public/static/locales/fr/form-builder.json
```diff
@@ -203,6 +203,9 @@
       "title": "Plus d'informations"
     },
     "email": "Votre courriel",
+    "department": "Ministère ou organisme",
+    "branch": "Direction ou secteur",
+    "jobtitle": "Titre de poste",
```
daine, line 208: Wondering if we could use camelcase for this. The issue with just using `title` is that there is already a field for that, which is the title of the page.

```suggestion
    "jobTitle": "Titre de poste",
```

pages/form-builder/support/[[...supportType]].tsx
```diff
@@ -116,6 +122,29 @@ export default function Contactus() {
...
+        </div>
+        <div className="focus-group mt-14">
+          <Label id={"label-branch"} htmlFor={"branch"} className="required" required>
+            {t("contactus.branch")}
+          </Label>
+          <TextInput type={"text"} id={"branch"} name={"branch"} className="required w-[34rem]" />
+        </div>
+        <div className="focus-group mt-14">
+          <Label id={"label-title"} htmlFor={"title"} className="required" required>
+            {t("contactus.title")}
```
Comment (daine) [inline pages/form-builder/support/[[...supportType]].tsx:144]: I think this change from `contactus.jobtitle` to this - `contactus.title` has resulted in a bug. It's now showing up as "Contact us" instead of "Job Title"


Screenshot 2023-08-10 at 9 49 30 AM

Could we just do (completely up to you @thiessenp-cds )
```suggestion
            {t("contactus.jobTitle")}
```

daine (commented):
Thanks for making the change to the names/variable, however I think it may have cause an issue going from `jobtitle` to simply `title`.

pages/form-builder/support/[[...supportType]].tsx
Comment (thiessenp-cds) [reply pages/form-builder/support/[[...supportType]].tsx]: Thanks for taking another look. From the start I went back and forth on title vs jobTitle vs job_title vs job-title. I went with jobtitle because it felt like a middle ground. But your point about consistency makes the most sense

thiessenp-cds:
I'll go ahead an merge this and do the intro content change in a separate PR

PR #2520, opened 2023-08-11
Fix/update contactus page

thiessenp-cds:
# Summary | Résumé

Adds new fields *only* to the Contact Us page. The support page should be unchanged. Also updates the intro content.

Note the contact us email was updated to: `assistance+forms-formulaires@cds-snc.ca`

Screenshot 2023-08-15 at 8 50 39 AM

# Testing

Try submitting a contact us request. The new fields should be on the form and show up in the email.

Try submitting a support request. The behavior and email should be unchanged.


pages/form-builder/support/[[...supportType]].tsx
```diff
@@ -33,400 +34,457 @@ export default function Contactus() {
...
-                </Button>
-              </form>
-            </>
-          )}
-        </Formik>
-      )}
       {isSuccessScreen && (
         <>
-          <h1>{t("requestSuccess.title")}</h1>
+          <h1 ref={headingSuccessRef}>{t("requestSuccess.title")}</h1>
```
Comment (thiessenp-cds) [inline pages/form-builder/support/[[...supportType]].tsx:487]: Focusing the heading vs a live region around the "page" is a bit cleaner to let an assistive technology user know the page content has changed.

pages/form-builder/support/[[...supportType]].tsx
```diff
@@ -1,7 +1,7 @@
 import { serverSideTranslations } from "next-i18next/serverSideTranslations";
```
Comment (thiessenp-cds) [inline pages/form-builder/support/[[...supportType]].tsx:1]: The form structure changed between the two forms. IMO separating the forms like the below makes the most sense vs. a lot of if-statements  in a single form that can be hard to follow. This also takes the forms one step closer to becoming separate pages/files that would be easier to maintain IMO.

pages/form-builder/support/[[...supportType]].tsx
```diff
@@ -33,400 +34,457 @@ export default function Contactus() {
...
-      },
-      data: { supportType, name, email, department, branch, jobTitle, request, description },
-      // If development mode disable timeout
-      timeout: process.env.NODE_ENV === "production" ? 60000 : 0,
-    }).catch((err) => {
-      logMessage.error(err);
-      setErrorMessage(t("contactus.errors.submissionError"));
-    });
-  };
+  useFocusIt({ elRef: headingSuccessRef, dependencies: [isSuccessScreen] });
```
Comment (thiessenp-cds) [inline pages/form-builder/support/[[...supportType]].tsx:39]: Since we have an easy way to focus a header I went with that vs. using a live region around an entire page update. At the time it was a quick solution, technically correct, but the header focus to indicate a new page loading is cleaner

pages/form-builder/support/[[...supportType]].tsx
```diff
@@ -487,7 +486,7 @@ export default function Contactus() {
           <h1 ref={headingSuccessRef}>{t("requestSuccess.title")}</h1>
           <p className="mb-16 mt-[-2rem] font-bold">{t("requestSuccess.weWillRespond")}</p>
           <div className="mb-16">
-            <StyledLink href={`/myforms`} className="gc-button--blue">
+            <StyledLink href={`/myforms`} theme="primaryButton">
```
Comment (timarney) [inline pages/form-builder/support/[[...supportType]].tsx:489]: Should be able to use <LinkButton.Primary here

Comment (thiessenp-cds) [reply pages/form-builder/support/[[...supportType]].tsx]: Got it.  So StyledLink is so out and LinkButton.Primary is so hot right now :D

timarney (approved):

PR #2536, opened 2023-08-14
fix: display missing privilege in admin panel

craigzour:
# Summary | Résumé

- Added missing privilege in admin panel


timarney (approved):
Perfect

PR #2539, opened 2023-08-15
chore: adding new gtm tag for unlocking publishing (request)

daine:
# Summary | Résumé

Adds new data-* class selector for google tag manager (gtm) to trigger on Unlock Publishing requests

# Pull Request Checklist


components/forms/Button/Button.tsx
```diff
@@ -48,6 +52,7 @@ export const Button = (
       data-testid={testid ? testid : "button"}
       disabled={disabled}
       {...ariaProps}
+      {...(dataAttrName && dataAttrValue ? { [`data-${dataAttrName}`]: dataAttrValue } : "")}
```
timarney, line 55: I would prefer we use `...props` at this point vs adding more props to the args.

We'll be looking to re-factor the Button component in the future to remove some of these "extra" props i.e. shape that have been added.

We'll also look to remove standard props that a html button should just handle like data attributes, disabled etc...

daine: I think there are 2 Button components (there might be a purpose to having two? not sure), the other one also had this [multiple extra prop issue](https://github.com/cds-snc/platform-forms-client/blob/develop/components/globals/Buttons/Button.tsx).

I modified the PR to remove extras, let me know if that works or if it needs changing. Thanks @timarney!

timarney:
Will review and get into the release upcoming on Mon.

components/forms/Button/Button.tsx
```diff
@@ -12,23 +11,14 @@ interface ButtonProps {
   destructive?: boolean;
   testid?: string;
   ariaProps?: React.AriaAttributes;
+  dataAttrName?: string;
```
timarney, line 14: Given the `...props` we should be able to remove the `dataAttrName` and  `dataAttrValue`

components/forms/Button/Button.tsx
```diff
@@ -12,23 +11,14 @@ interface ButtonProps {
   destructive?: boolean;
   testid?: string;
   ariaProps?: React.AriaAttributes;
```
timarney, line 13: Looks like `ariaProps` was removed but we still have it here.

components/forms/Button/Button.tsx
timarney: Right --- yes there are 2 buttons components 

The "original" and the one being used on the page your modifying is the button for "public forms" vs the other one that we don't need to modify for this PR is a button we created for use within the form builder / admin.  At the time we didn't want to alter anything with "public forms".  At some point once we get some styling sorted out we can re-combine but that's way outside the scope of this PR :)

components/forms/Button/Button.tsx
daine: Removed!

components/forms/Button/Button.tsx
```diff
@@ -11,24 +10,12 @@ interface ButtonProps {
...
-    secondary,
-    base,
-    onClick,
-    className,
-    testid,
-    destructive,
-    disabled,
-    ariaProps,
-  } = props;
+  const { children, secondary, base, onClick, className, testid, destructive } = props;
```
timarney, line 18: Looking at the failing test ... 
 
 we should use ...rest 
 
 { children, secondary, base, onClick, className, testid, destructive, ...rest } = props;
 
 ```
 <button
      className={classes}
      onClick={onClick}
      data-testid={testid ? testid : "button"}
      {...rest}
    >
 ```

timarney (approved):

PR #2553, opened 2023-08-16
chore: Updates left nav to be more semantic

thiessenp-cds:
# Summary | Résumé

Adds more semantics around where a user is for "page level" nav lists. Using aria-current="page" tells an AT user that this item in the list is the page s/he's on.

As a bonus, the nav lists are updated to list elements. Why a UL/OL list? AT users have more utility when multiple items that form a list are structured in a OL/UL. Like the number of items in the list and which item in the order of list items they're in. UL was chosen because order number doesn't seem implicit in the items.

## Test

The behavior and look of the lists should be unchanged. The only change should be that the lists are now in UL elements and the aria-current="page" should only be added to the current active item that is the displaying "page"

Nav Lists updated:
 - /form-builder/edit
 - /en/myforms/drafts
 - /admin/accounts


components/myforms/LeftNav/index.ts
thiessenp-cds, line 1: Since there is now only 1 file in the dir, having an index seems overkill IMO

components/form-builder/app/navigation/LeftNavLink.tsx
```diff
@@ -27,6 +27,7 @@ export const LeftNavLink = ({ children, href }: { children: ReactElement; href:
         className={`${
           active ? "font-bold" : ""
         } group no-underline rounded block w-38 py-1 mb-2 pl-2 pr-0 laptop:pr-2 text-black-default hover:text-blue-hover visited:text-black-default focus:text-white-default focus:bg-blue-hover active:no-underline active:bg-blue-hover active:text-white-default !shadow-none`}
+        {...(active && { "aria-current": "page" })}
```
thiessenp-cds, line 30: AT support for On/Off with ARIA attributes generally works more consistently by removing the attribute for the Off/false state. For aria-current, our use is only "page" or false, so it's either on the element or just removed altogether

timarney (approved):

PR #2632, opened 2023-09-12
Adds a description to the Report form

thiessenp-cds:
# Summary | Résumé

Adds a description to the Report form.

**I started refactoring this also but the number of changed files was around 20+. I figured it would be easier to review with just the needed changes and then follow up with a refactoring PR.**

First let Jimmy know you'll be doing a test.

Go to a form with responses and:
1. Copy a form name 
2. Open the report problem dialog
3. Enter the form name and try to submit
4. An error should show about the description not being empty
5. Enter something in the description field and try to submit
6. The dialog should close, the table updated and a freshdesk ticket created

Do the same test but in a different language.


components/form-builder/app/responses/DownloadTableDialog.tsx
```diff
@@ -50,6 +56,7 @@ export const DownloadTableDialog = ({
   unknownErrorTitle,
   unknownErrorDescription,
   unknownErrorDescriptionLink,
+  tempType,
```
thiessenp-cds, line 59: This will be refactored into two dialogs in the next PR along with the below code.

components/form-builder/app/responses/DownloadTableDialog.tsx
```diff
@@ -207,6 +233,42 @@ export const DownloadTableDialog = ({
               setStatus={setStatus}
             ></LineItemEntries>
```
thiessenp-cds, line 235: As you can see, not separating these dialogs gets uglier and uglier.. definitely time to refactor soon

components/form-builder/app/responses/DownloadTableDialog.tsx
```diff
@@ -102,6 +113,16 @@ export const DownloadTableDialog = ({
       return;
     }
 
+    // TEMP
+    if (tempType === "report" && descriptionRef.current === "") {
```
timarney, line 117: What is `tempType` here?

thiessenp-cds: It acts to identify which dialog to show/hide related content. Named temp to make it very clear that the type and related is just temporary. The refactor will have two dialogs etc.

timarney (approved):
Tested via the preview URL ✔️

PR #2636, opened 2023-09-13
Cleanup unused GITHUB_SHA vars in Dockerfile and workflows

dsamojlenko:
# Summary | Résumé

The GITHUB_SHA env var in the Dockerfile is not used. It was copied over from a previous project that included the GITHUB_SHA in the compiled application. We do not do that here, so it is unnecessary.

The only place we need the GITHUB_SHA is in the Staging workflows that build a container as the Staging container gets tagged with the GITHUB_SHA.

Also removes an unused PRODUCTION_ENV var from production workflow container builds.


bryan-robitaille (dismissed):

bryan-robitaille (approved):
👍

PR #2638, opened 2023-09-13
refactor: Responses Dialogs

thiessenp-cds:
# Summary | Résumé

Refactors the Responses Dialogs to be easier to maintain. The behaviour and design should be unchanged.

Give Jimmy a heads up that you'll be testing "report a problem"

Go to a form with responses and try:
 - confirming a response
 - report a problem for a response
 
Also since it was changed, test updating a security question (in Profile). Trying to enter an empty profile question should show an error.


components/form-builder/app/responses/Dialogs/DialogStates.ts
```diff
@@ -0,0 +1,12 @@
+export enum DialogStates {
```
thiessenp-cds, line 1: This is breaking the "interface" segregation principle :). I couldn't come up with a good way to extend an enum in TS. Searching around it seems common to use a type with a union. But then an enum would no longer be being used.

components/form-builder/app/responses/Dialogs/line-item-entries/LineItemEntries.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useEffect, useRef } from "react";
 import { LineItems } from "./LineItems";
```
thiessenp-cds, line 2: Moved the LinItem and related into the Dialogs directory since they are specific to the Dialogs and not really reusable.

components/form-builder/app/responses/Dialogs/ConfirmDialog.tsx
```diff
@@ -0,0 +1,215 @@
+import React, { useState } from "react";
```
thiessenp-cds, line 1: Copy+paste from DownloadTableDialog and removes anything related to the ReportDialog

components/form-builder/app/responses/Dialogs/ReportDialog.tsx
```diff
@@ -0,0 +1,263 @@
+import React, { useRef, useState } from "react";
```
thiessenp-cds, line 1: Copy+paste from DownloadTableDialog and removes anything related to the Confirm behavior

components/globals/ValidationMessage/ValidationMessage.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
```
thiessenp-cds, line 1: In the future success/* can be added

components/form-builder/app/responses/Dialogs/DialogStates.ts
timarney: Curious if this would work

Noting it might not but curious as I said :)

```
export const DialogStates = {
  EDITTING: "EDITTING",
  SENDING: "SENDING",
  SENT: "SENT",
  MIN_ERROR: "MIN_ERROR",
  MAX_ERROR: "MAX_ERROR",
  FORMAT_ERROR: "FORMAT_ERROR",
  DESCRIPTION_EMPTY_ERROR: "DESCRIPTION_EMPTY_ERROR",
  FAILED_ERROR: "FAILED_ERROR",
  UNKNOWN_ERROR: "UNKNOWN_ERROR",
} as const;

export const ResponsesStates = {
  ...DialogStates,
  SOMETHING_ELSE: "SOMETHING_ELSE",
} as const;

export type DialogStatesType = (typeof DialogStates)[keyof typeof DialogStates];

export type ResponsesStatesType = (typeof ResponsesStates)[keyof typeof ResponsesStates];
```

pages/form-builder/responses/[[...params]].tsx
```diff
@@ -196,113 +196,20 @@ const Responses: NextPageWithLayout<ResponsesProps> = ({
...
-            "downloadResponsesModals.reportProblemsDialog.errors.unknown.description"
-          )}
-          unknownErrorDescriptionLink={t(
-            "downloadResponsesModals.reportProblemsDialog.errors.unknown.descriptionLink"
-          )}
-          tempType="report"
-          setIsServerError={setIsServerError}
-        />
-      )}
+      <ConfirmDialog
```
Comment (timarney) [inline pages/form-builder/responses/[[...params]].tsx:199]: Nice -- this cleans things up a lot.

timarney:
Tested on review environment - working as expected ... one  comment if you want to give that a try 

https://github.com/cds-snc/platform-forms-client/pull/2638#discussion_r1326413968

components/form-builder/app/responses/Dialogs/DialogStates.ts
thiessenp-cds: When I try this  e.g.
`
const [status, setStatus] = useState<ResponsesStatesType>(ResponsesStatesType.EDITTING);
`
I keep seeing a TS error about
`"'ResponsesStatesType' only refers to a type, but is being used as a value here."`

I could just drop using an enum here.. hmm..

timarney: Will approve as is and we can play around further with it later.

timarney (approved):

PR #2643, opened 2023-09-14
fix: Cypress test for Share Modal and refactor for helpers

bryan-robitaille:
# Summary | Résumé

## Adds a new cypress test helper function that types into a field and then waits to ensure that the component has rendered the text before continuing.

This stems from an issue I've noticed when running testing on slower systems or in dev:test mode.  Often the failed screenshots will show the typed field being blank when trying to submit a form or other action.

New Cypress helper:
```js
/**
* Helper to type in input and other fields
* @param field jQuery selector for the field to type in
* @param typeText Text to type into the field.  Can contain special commands in { }
* @param overrideText This can be used to override the text to verify against.  Helpful when adding to an already existing string.
*/
cy.typeInField(field: string, typeText: string, overrideText? :string)
```

## Leverages existing helpers I had built previously
```js
/** Helper to visit and ensure page is rendered before start assertions
* @param page Url of page to visit
*/
cy.visitPage(page: string)
```

## Fix race condition when reseeding the database between tests

The cypress config task to teardown and seed the database was not awaiting completion before starting the test

## Modifed assertions for "exists"

Although the element existed in the DOM the children were often not yet rendered causing race conditions in tests.  Where possible "exists" has been swapped out for "be.visible".


timarney (approved):
Nice

PR #2657, opened 2023-09-18
Fixes missing translation key

thiessenp-cds:
# Summary | Résumé

Adds missing translation key for Address in the form-builder.

## Test

Try setting the language to French. Then in the form-builder menu dialog, the Address block label should be in French (Addresse)


anikbrazeau (approved):
This looks good in the preview environment! Thanks Pete!

PR #2659, opened 2023-09-18
Fix for CSV file download as UTF8

thiessenp-cds:
# Summary | Résumé

Only in Windows, the translated version of a form with French characters was being saved in ANSI (windows default). So the French characters would not display correctly when imported into Excel. The fix sets the default to UTF8.

1. From Windows (Tablet? VM?) open Explorer/* and go to GC forms
2. In form-builder create a form with French characters (e.g.  é / à / ô)
3. Export the form to the local desktop (should create a .csv file)
4. From Excel open the .csv file
5. The French characters should display correctly

Refactor file download as a shared component/* once this is tested and shown to work :)


components/form-builder/app/translate/DownloadCSV.tsx
```diff
@@ -90,7 +90,11 @@ export const DownloadCSV = () => {
 
     const csv = data.map((row) => row.join(",")).join("\n");
 
-    const blob = new Blob([csv], { type: "application/csv" });
+    // Windows saves CSV files by default as ANSI. This foces UTF-8.
```
timarney, line 93:
```suggestion
    // Windows saves CSV files by default as ANSI. This forces UTF-8.
```

timarney (approved):

PR #2663, opened 2023-09-19
fix: Fixes html download file missing French

thiessenp-cds:
# Summary | Résumé

The responses HTML file was displaying English text when French text should have been displayed. This was a bit tricky because of how the related JavaScript was written inline in the page. The fix was to set the localization string to French manually. 

1. Go to a form with responses
2. Download a response
3. Open that response and try clicking the "Copier le code" button
4. The French translation "Copié dans le presse-papiers" should be displayed 
5. Click the other buttons as well to make sure the text is displayed in the right language

N/A


timarney (approved):
Tested via preview url 👍

PR #2674, opened 2023-09-20
Fix/refactors file download

thiessenp-cds:
# Summary | Résumé

Adds a util function to download a file in the browser. Also updates related code to call this function.

I eventually gave up trying to get a Blob() to respect the mime type I sent it. Looking at the debugger I could see the object always had the type "text/xml". Windows/Mac seem to go with whatever file extension is on the filename so it probably doesn't matter. I guess this comment is just a rant :)

Try downloading a file in the related areas:
- Responses
- Forms Card
- Form-builder translate tool


thiessenp-cds:
I'm not sure why the tests are failing off and on.

thiessenp-cds:
This PR has been sitting around for a bit and with Friday and the weekend coming up, I know I'll forget about it. It's not assigned work so I'll just close this PR and come back to it if ever needed

PR #2748, opened 2023-10-06
Fix/noindex only in staging

craigzour:
# Summary | Résumé

- Fixed issue where the logic to add the `noindex` tag in Staging was not working properly


thiessenp-cds (approved):
Interesting. Sure, I'll test this on staging also. We should probably add a note somewhere to test this when released to production. But LGTM :)

PR #2750, opened 2023-10-06
fix: add noindex meta tag in Staging on all pages including root index

craigzour:
# Summary | Résumé

- Attempting to fix `noindex` meta tag not being included on root index page

### DIFF
--- a/pages/_document.tsx
+++ b/pages/_document.tsx
@@ -208,14 +208,17 @@ class StrictStaticCSP extends Head {
 // Actual main Document being rendered
 
 const CustomHead = process.env.NODE_ENV === "production" ? StrictStaticCSP : Head;
+
+const NoIndexMetaTag = process.env.NEXTAUTH_URL?.includes("staging") ? (
+  <meta name="robots" content="noindex,nofollow" />
+) : null;
+
 class MyDocument extends Document {
   render() {
     return (
       <Html>
         <CustomHead>
-          {process.env.NEXTAUTH_URL?.includes("staging") && (
-            <meta name="robots" content="noindex,nofollow" />
-          )}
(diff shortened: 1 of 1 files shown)


timarney:
Thoughts on generating a robots.txt file?

https://mmazzarolo.com/blog/2021-04-27-nextjs-robots-txt/#generating-a-robotstxt-in-the-build-process

craigzour:
> Thoughts on generating a robots.txt file?
> 
> Can alter for prod
> 
> ```
> const robotsTxt =
>     process.env.VERCEL_ENV === "production"
>       ? crawlableRobotsTxt
>       : uncrawlableRobotsTxt;
> ```
> 
> https://mmazzarolo.com/blog/2021-04-27-nextjs-robots-txt/#generating-a-robotstxt-in-the-build-process

I used the `noindex` meta tag approach because we want to make sure we are not included in search results. 

According to https://aioseo.com/docs/when-to-use-noindex-or-the-robots-txt/ this is the right solution for what we are trying to achieve.

`So if you want content not to be included in search results, then use NOINDEX. If you want to stop search engines crawling a directory on your server because it contains nothing they need to see, then use “Disallow” directive in your robots.txt file.`

thiessenp-cds (commented):
Oh interesting, so it looks like it was a context problem. 
Nice, LGTM :) but I'll leave the PR open for @timarney to look at as well

timarney (approved):

craigzour:
Also got confirmation in the Google documentation in the intro of https://developers.google.com/search/docs/crawling-indexing/robots/intro that the `robots.txt` file would not keep us out of search results. cc @thiessenp-cds @timarney

PR #2800, opened 2023-10-24
chore: Prep for response filtering and pagination

bryan-robitaille:
# Summary | Résumé

Uses the GSI Index based on Form ID and Response Status to provide the ability to retrieve responses based on status without performance overhead.

The functionality is currently not leveraged in the application but it sets up the response table pagination feature.

Prerequisite:
Local
- create more than 500 responses in the vault for a form using the util script
- download some responses so that statuses change

in `pages/form-builder/responses` in `getServerSideProps` line 275 add a filter to the `listAllSubmissions` function by adding a VaultStatus arg.
```js
listAllSubmissions(ability, formID, VaultStatus.New)
```
Only New responses should be rendered on the response table.  Any responses that were status of `downloaded` or `confirmed` should have made way for responses of status `New`.

```js
listAllSubmissions(ability, formID, VaultStatus.Downloaded)
```
Only Downloaded responses should be rendered on the response table.

To enable pagination a new API will be required.  This API will also allow the form response table to refresh it's contents without requiring serverSideProps to run on a page refresh.

# Pull Request Checklist


lib/vault.ts
```diff
@@ -157,19 +161,30 @@ export async function listAllSubmissions(
...
-    const numOfUnprocessedSubmissions = accumulatedResponses.filter((submission) =>
-      [VaultStatus.NEW, VaultStatus.DOWNLOADED, VaultStatus.PROBLEM].includes(submission.status)
-    ).length;
-
-    await numberOfUnprocessedSubmissionsCachePut(formID, numOfUnprocessedSubmissions);
+    if (!status) {
+      // Only update value when we are not filtering by status
+      // Will need to rework logic once filtering is implemented on client side
+      const numOfUnprocessedSubmissions = accumulatedResponses.filter((submission) =>
+        [VaultStatus.NEW, VaultStatus.DOWNLOADED, VaultStatus.PROBLEM].includes(submission.status)
```
thiessenp-cds, line 171: Would adding `VaultStatus.CONFIRMED` be helpful also?

bryan-robitaille: `VaultStatus.Confirmed` would not be an unprocessed Submission.  This function is the existing one that tries to calculate the number of unconfirmed responses.
The logic of this will all be rewritten once we implement the filtering and pagination in the client side.

thiessenp-cds (approved):
LGTM :)

PR #2829, opened 2023-11-07
Wire up download button to new downloadable formats API

dsamojlenko:
# Summary | Résumé

- Wires up the current Download button to the new Downloadable Formats API endpoint.
- Removes the old HTML output components, which were refactored in the previous PR.
- Updates the response handler for HTML to loop over the json data, extract the html, and generate the rendered document for downloading
- Adds a 200ms delay to the loop/download trigger to make downloading large numbers of individual responses more reliable
- Adds the ability to configure max number of response downloads in Settings (Defaults to 20)

## To test
- Login to the [PR Review Environment](https://hfuf274k5nohp7ht6cnk75r2c40pcmiw.lambda-url.ca-central-1.on.aws/)
- Go to Forms
- Find a form with many responses (try Dave's Form)
- Open browser network inspector
- Select up to 20 responses to download

In your browser network inspector, you should see a single request to the new api endpoint:
`/api/id/[FormID]/submissions/download?format=html`

You should receive the number of HTML files you've requested.

If you would like to test with a higher number of downloads, go to Administration/Settings and set a new value for `Limit the number of responses that can be downloaded at once`

### Compare to Staging
If you perform the same action on Staging, you will see many requests in your network inspector (as many requests as files requested), and you will likely not receive all of them if you have selected more than 15.


pages/form-builder/responses/[[...params]].tsx
```diff
@@ -295,11 +299,14 @@ export const getServerSideProps: GetServerSideProps = async ({
     }
   }
 
+  const responseDownloadLimit = Number(await getAppSetting("responseDownloadLimit")) || 20;
```
Comment (thiessenp-cds) [inline pages/form-builder/responses/[[...params]].tsx:302]: Thanks for completing that TODO :)

thiessenp-cds (approved):
LGTM :)

We can do more testing around many files downloading etc. later on staging

Issue #2821, opened 2023-11-02
Unable to save form if field contains HTML


jprince-cds:
The form was unable to be saved because it contained < > brackets that look similar to HTML tags. Unfortunately, the error is non-descriptive to the clients so they don't know what is happening or what needs to be fixed.

Would it be possible to HTML-encode problematic characters to avoid this error?

Freshdesk Ticket #15927

jprince-cds:
The form was unable to be saved because it contained < > brackets that look similar to HTML tags. Unfortunately, the error is non-descriptive to the clients so they don't know what is happening or what needs to be fixed. 

Would it be possible to HTML-encode problematic characters to avoid this error? 

Freshdesk Ticket #15927

timarney:
@samsadasivan @connorscarolyns wondering how we should proceed here for the design end.

We've seen several support tickets where this has come up.

@dsamojlenko and I looked into what we can do on the dev end and have determined that we can pinpoint the field or fields where the markup or what appears to the editor as markup appears.  

Should we look to remove the markup upon when trying to save --- which should fix the save problem or display am alert indicating the problems(s) -- and possibly link back to the specific text area.  If we go this direction I think we should move the error outside of the save button and look to display it in a more prominent location (and larger i.e. a full alert).

timarney:
Ref: https://github.com/cds-snc/platform-forms-client/pull/2832

connorscarolyns:
I think in French, < kjjkkj > is used as a quotation mark? Telling people they need to remove it could be problematic. Is there a way er can escape these characters when the template is saved?

timarney:
Notes from some some further testing

« Je voudrais du thé. C'est combien ? »


<img width="880" alt="Screenshot 2023-11-08 at 9 35 29 AM" src="https://github.com/cds-snc/platform-forms-client/assets/62242/b9d35e28-665e-40ef-add4-235e93077cd6">

This also passes 
```
< a >
```

This fails
```
<a>
```


Copying and pasting links also works

<img width="400" alt="Screenshot 2023-11-08 at 9 38 59 AM" src="https://github.com/cds-snc/platform-forms-client/assets/62242/c4006cdb-9a71-4d57-9056-7180a6d035f9">

---


We can look into converting the < and >

timarney:
Another support ticket 

https://cds-snc.freshdesk.com/a/tickets/15993

This time outside of the Richtext area


```
[...]

PR #2832, opened 2023-11-07
fix: Pad angle brackets in user uploaded json or text input to avoid html exception

dsamojlenko:
# Summary | Résumé

This PR addresses an issue where a user could either: 
- upload a form file; or 
- enter text into a question or description input

...that includes HTML or text that looks like HTML (ie `<text that has angle brackets on either end>`, but saving would throw an error and fail to save due to invalid input. This usually results in the user needing to contact support to resolve.

This PR addresses the issue on two ends- on JSON upload, and on input/save, by "escaping" any angle brackets found by adding a space after an opening bracket and a space before a closing bracket.

The matching algorithm only matches strings that contain both an opening and closing bracket, and ignores them if there are already spaces. For example:

- `<something>` becomes `< something >`
- `<something > becomes `< something >`
- `< something>` is valid and remains `< something>`
- `<something` is valid and remains `<something`

The text between the angle brackets can contain any number of alphanumeric characters, spaces, punctuation, or special characters.

Fixes: https://github.com/cds-snc/platform-forms-client/issues/2821


components/form-builder/util.ts
```diff
@@ -259,3 +259,31 @@ export const isVaultDelivery = (deliveryOption: DeliveryOption | undefined) => {
...
+      } else if (typeof obj[key] === "string") {
+        obj[key] = escapeAngleBrackets(obj[key]);
+      }
+    }
+
+    return obj;
+  }
+
+  return input;
+};
```
dmethvin-gov, line 289: A recursive type can avoid the `any`. I think the code below should work and avoid the `in` operator which is flakey for some objects and arrays.

```suggestion
type Cleanable = string | Cleanable[] | { [key: string]: Cleanable } | unknown;

export const cleanInput = (input: Cleanable) => {
    if (typeof input === "string") {
        return escapeAngleBrackets(input);
    }
    if ( Array.isArray(input) ) {
        return input.map((elem) => cleanInput(elem));
    }
    if ( input instanceof Object ) {
        return Object.fromEntries(
            Object.entries(input).map(
                ([key, value]) => [key, cleanInput(value)]
            )
        )
    }
    return input;
};
};
```

timarney: Thanks @dmethvin-gov you beat us to cleaning up the for this :)

That's a nice great improvement on the recursion code 🎉 .

timarney:
@dsamojlenko I fixed up the backspace issue.

Let's get CI fixed up and write some tests for this PR.

timarney (approved):
There's still room to iterate on this for sure but I think this solves the immediate issue.

Specifically some of the inputs live update while other don't leading to some inconsistency as to what's happening with the fields.  

Will approve `as is` as it should cut down on the incoming support request re:failed to save message.

PR #2845, opened 2023-11-09
Add "Select all" header icon to DownloadTable

dsamojlenko:
# Summary | Résumé

Adds an icon to the header of DownloadTable to enable selecting all visible records.
Icon toggles between 3 states:
- All selected
- None selected
- Indeterminate (some selected)

Screen Recording 2023-11-09 at 4 12 30 PM


thiessenp-cds (commented):
Thanks for reaching out about accessibility. For the "select all" icon, just imagine that instead of an icon you had to explain the change with a hidden text label. The text should be announced to a screen reader though.

thiessenp-cds (approved):
LGTM :)

PR #2848, opened 2023-11-10
Responses DownloadTable More menu and single item download

dsamojlenko:
# Summary | Résumé

Adds a "More" menu to the Responses download table with two options:
- Download single response
- Delete response (placeholder - to be implemented in future)

Screen Recording 2023-11-10 at 2 50 38 PM


thiessenp-cds (approved):
Nice work, LTM :)

The download table is becoming more interactive. If more cells get actions in the future, the semantics should change from a table to a grid which is quite a bit more complex. Though that's tomorrow's problem and only if we add more controls :)

PR #2859, opened 2023-11-15
Add Download/Confirm panel to Responses

dsamojlenko:
# Summary | Résumé

Adds the Download/Confirm panel to the bottom of the Responses screen and adds a RESET action to the Reducer to handle when switching Status filter.

This PR just adds the panel and uses the existing Download/Delete buttons and actions.

Screen Recording 2023-11-16 at 11 46 02 AM


thiessenp-cds (approved):
Nice work, LGTM :)

We'll probably need to add some relationships/semantics between the download/selected buttons and the response checkboxes. But user testing will show if users are confused with this

timarney (approved):

PR #2867, opened 2023-11-16
Add confirm button and dialog to the Downloaded Responses page

dsamojlenko:
# Summary | Résumé

Restores the Confirm dialog and button to the Downloaded Responses page and updates the content of the dialog. Also uses responseDownloadLimit for the maximum number of codes to input. This may need its own setting eventually.

image

image

### DIFF
--- a/components/form-builder/app/responses/Dialogs/DownloadDialog.tsx
+++ b/components/form-builder/app/responses/Dialogs/DownloadDialog.tsx
@@ -1,5 +1,0 @@
-import React from "react";
-
-export const DownloadDialog = () => {
-  return <div>DownloadDialog</div>;
-};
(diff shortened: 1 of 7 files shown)


components/form-builder/app/responses/Dialogs/DownloadDialog.tsx
```diff
@@ -0,0 +1,5 @@
+import React from "react";
+
+export const DownloadDialog = () => {
+  return <div>DownloadDialog</div>;
```
thiessenp-cds, line 4: Assuming here this is just a placeholder

thiessenp-cds (approved):
Nice work, LGTM :)

components/form-builder/app/responses/Dialogs/DownloadDialog.tsx
dsamojlenko: oh yeah that's an accidental commit, will remove before merge good catch

PR #2869, opened 2023-11-16
Hide delete on confirmed

dsamojlenko:
# Summary | Résumé

Hides the inline Delete option on the More menu on the Confirmed Responses screen.

| image | image |


thiessenp-cds (approved):
LGTM :)

PR #2874, opened 2023-11-17
Use responseDownloadLimit setting for listAllSubmissions query

dsamojlenko:
# Summary | Résumé

The responseDownloadLimit setting was added to limit the number of concurrent HTML file downloads. 

This PR applies that same limit to listAllSubmissions (the max number of submissions that can be displayed)


bryan-robitaille (approved):
👍

PR #2875, opened 2023-11-17
Cleanup error messages that are no longer needed

dsamojlenko:
# Summary | Résumé

Just a bit of cleanup. These error messages are no longer needed as it's not possible to click the Download button if there are no responses selected, and it's not possible to select more responses than the responseDownloadLimit because we will only retrieve/display up to that limit.


craigzour (approved):
LGTM

PR #2880, opened 2023-11-21
Add number shown to "more responses" alert

dsamojlenko:
# Summary | Résumé

When displaying responses in the DownloadTable, an Alert has been added to the bottom to indicate there are more responses available. This just adds the current number of responses shown to that message.

Note, this messaging should be reviewed by Content/Design.

image

### DIFF
--- a/public/static/locales/en/form-builder-responses.json
+++ b/public/static/locales/en/form-builder-responses.json
@@ -74,6 +74,8 @@
       "errorDownloadingFilesHeader": "Error downloading the selected files.",
       "errorDownloadingFiles": "Try again in a few minutes. If the problem persists, ",
       "errorDownloadingFilesLink": "contact Support",
+      "remainingResponses": "There are remaining responses on the server",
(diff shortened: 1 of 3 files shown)


public/static/locales/en/form-builder-responses.json
```diff
@@ -74,6 +74,8 @@
       "errorDownloadingFilesHeader": "Error downloading the selected files.",
       "errorDownloadingFiles": "Try again in a few minutes. If the problem persists, ",
       "errorDownloadingFilesLink": "contact Support",
+      "remainingResponses": "There are remaining responses on the server",
+      "remainingResponsesBody": "Showing {{max}} responses. Not all responses can be shown, please download responses to see more.",
```
anikbrazeau, line 78:
```suggestion
      "remainingResponses": "There are remaining responses on the server",
      "remainingResponsesBody": "Showing {{max}} responses. Not all responses can be shown, download responses to see more.",
```

PR #2893, opened 2023-11-22
Responses table updates

thiessenp-cds:
# Summary | Résumé

Implements the figma designs for the new responses table. The bulk of the work is in the "Next step" column and that is where I'd say the focus of the PR should be (checking that logic).


thiessenp-cds:
Moving work to a different branch https://github.com/cds-snc/platform-forms-client/pull/2896

PR #2900, opened 2023-11-22
Remove limit on confirmation codes, fix retrieval for large numbers of responses

dsamojlenko:
# Summary | Résumé

Updates the Confirm dialog to remove the limit of confirmation codes
Fixes the retrieveSubmissions vault query to chunk the request properly and return the full data set


pages/api/id/[form]/submission/confirm.ts
```diff
@@ -12,12 +12,13 @@ import { connectToDynamo } from "@lib/integration/dynamodbConnector";
 import { AccessControlError, createAbility } from "@lib/privileges";
 import { checkUserHasTemplateOwnership } from "@lib/templates";
 import { logEvent } from "@lib/auditLogs";
-
-// TODO: move to an app setting variable
-const MAXIMUM_CONFIRMATION_CODES_PER_REQUEST = 20;
```
Comment (craigzour) [inline pages/api/id/[form]/submission/confirm.ts:17]: If I remember correctly, one of the reason why we had such limitation was because the AWS API we use to confirm responses has a maximum number of operations and also a payload size limitation. Just to make sure the code can handle any kind of `MAXIMUM_CONFIRMATION_CODES_PER_REQUEST` we decide in the future, you should split the confirmation operations in chunks of 20.

Comment (bryan-robitaille) [reply pages/api/id/[form]/submission/confirm.ts:17]: Looking at the API documentation again I can see that the BatchGet and TransactWrite commands both have limits of 100 items.  The confirmation function creates 2 items for each confirmation code in the TransactWrite command so in reality we can only accept 50 confirmation codes from the client at once in it's current state.
I like the idea of splitting the call on the client side into smaller batches.  This would help with error handling in the interim because all the confirmation codes pasted in wouldn't fail, only the batch that contained the the troublesome code would.
Thoughts?

Comment (craigzour) [reply pages/api/id/[form]/submission/confirm.ts:17]: Yep you are right! Will definitely improve the user experience for version 1

Comment (timarney) [reply pages/api/id/[form]/submission/confirm.ts:17]: Can we create a queue on our end to handle or will that really complicate things?

Take in as many as we can from the UI -- and send in batches of X to the api.

Comment (bryan-robitaille) [reply pages/api/id/[form]/submission/confirm.ts:17]: Yes, we'll be batching them client side so it's transparent to the user.  I'm testing it locally now and will hopefully update this PR later today.

craigzour (approved):
LGTM :)

dsamojlenko:
Beauty, works great!

PR #2904, opened 2023-11-24
Bugfix: Fix date format helper

dsamojlenko:
# Summary | Résumé

In some environments the test for the formatDate helper method started to fail. This is because Node 19 included a change to the `toLocaleDateString("en-CA"...` date format, so dates started coming out as `mm/dd/yyyy` instead of `yyyy-mm-dd`. See: https://github.com/nodejs/node/issues/45945

This PR updates the helper to just use the ISO8601 date, splitting off the time part.

### DIFF
--- a/lib/clientHelpers.ts
+++ b/lib/clientHelpers.ts
@@ -290,11 +290,7 @@ export const formatDate = (date: Date): string => {
     return "Unknown";
   }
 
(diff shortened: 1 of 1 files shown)


thiessenp-cds (approved):
Good catch. Works as expected and tests pass. LGTM :)

PR #2905, opened 2023-11-24
DownloadTable table row + button semantics

dsamojlenko:
# Summary | Résumé

Split out the "Download single file" button from the MoreMenu
Save the MoreMenu component for future
Add some table row / download button semantics for a11y

Following the example here: https://stackoverflow.com/questions/72231079/accessibility-how-to-handle-buttons-in-a-table
- Added a row header using the response id column for this purpose
- Added aria-labelledBy to the download button referencing the header
- Added some visually-hidden text to the row header and the button itself as well

### DIFF
--- a/components/form-builder/app/responses/DownloadTable.tsx
+++ b/components/form-builder/app/responses/DownloadTable.tsx
@@ -19,13 +19,13 @@ import { getDaysPassed, isStatus } from "@lib/clientHelpers";
 import { Alert } from "@components/globals";
 import { CheckAll } from "./CheckAll";
 import { DownloadButton } from "./DownloadButton";
-import { MoreMenu } from "./MoreMenu";
 import { ActionsPanel } from "./ActionsPanel";
 import { DeleteButton } from "./DeleteButton";
 import { ConfirmDeleteNewDialog } from "./Dialogs/ConfirmDeleteNewDialog";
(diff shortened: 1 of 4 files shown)


components/form-builder/app/responses/DownloadTable.tsx
```diff
@@ -250,7 +257,7 @@ export const DownloadTable = ({
                         }
                       }}
                       setDownloadError={setDownloadError}
-                      setShowConfirmNewDialog={setShowConfirmNewDialog}
+                      ariaLabelledBy={submission.name}
```
thiessenp-cds, line 260: Currently this is announced as Label + Accessible name or e.g. 16-11-0bd7, button

I wonder if adding the word download in there would make it more obvious? Maybe something like:
` ariaLabelledBy={submission.name + " download}`
which a screen reader would announce as: 16-11-0bd7 download, button

Though the table relationship exists so a user could/should be able to figure it out. Both work

thiessenp-cds (approved):
Nice work as is or with that minor change. LGTM :)

components/form-builder/app/responses/DownloadTable.tsx
thiessenp-cds: Wops, that should have been an array of elements

Issue #2914, opened 2023-11-27
Heading of empty states not coded as headings.


connorscarolyns:
H1's on empty state pages are hidden, and the empty state heading is coded as a paragraph. This may confuse users of screen readers; they will scan the page and hear that new responses are available when there are none. The same condition exists on each of the other tabs as well.

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM0VPQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--66a3c037f5d1ac55c8639ed63f491f5dd7877fc8/image.png)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM1lPQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--58159c3cca0fb97a443221fd720f155cdb1a715d/image.png)

thiessenp-cds:
Fixed/For review in PR https://github.com/cds-snc/platform-forms-client/pull/2928

Issue #2913, opened 2023-11-27
Broken Skip the response table link


connorscarolyns:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMVlPQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8b2a44a46d110e1ec89d22c98084fa6f52ab9464/image.png)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMXdPQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5f7af364650452176efda69c957a8eea98145a96/image.png)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMWdPQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ca19beef8ab8ac740d4bf68b4e7e553b15fae539/image.png)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMVVPQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1ca9e317b131cb4e4c3645b041561262d36f1822/image.png)

thiessenp-cds:
Fixed/For review in PR https://github.com/cds-snc/platform-forms-client/pull/2928

PR #2928, opened 2023-11-28
Accessibility Updates for the New File Formats Work

thiessenp-cds:
# Summary | Résumé

I'll be using this PR to track accessibility issues.

Accessibility Issues
- [x] Skip the response table doesn’t jump anywhere #2913 
- [x] Heading of empty states not coded as headings. #2914
- [ ] Download responses dialog, radio button/group on initial entry missing a focus indication
- [ ] Download responses dialog, checkbox button is controllable but missing keyboard focus indicator
- [x] Download responses dialog, checkbox not associated with label correctly (wrong Id?)
- [x] Skip link broken (minimal change - see notes below)

Browser/AT Quirks?
- Official receipt: Windows Chrome/Edge/FF+NVDA/Jaws not announcing “Copied to clipboard” 
	- macOS Chrome/FF announces correctly

Notes/Future Thought
- On closing Download response dialog, the browser takes focus with it's own download dialog so there isn't much we can do here. So it's difficult to say where/how to return focus to an element.
- On download/signing-off all submissions in the download table, all the submissions will be removed. The default for aria-live=polite is to have additions/removals announced. This makes sense but in this case it can be quite "noisy" for the a screen reader user. Imagine 50+ announcements that an item has been removed. Probably most users would just skip all the announcements in the queue. But still something to keep in mind
-  Skip link broken - More work to be done for pages with a hidden H1 e.g. /form-builder. It's not possible to focus a hidden h1 so we'll probably need to come with another solution/design that includes consistent visible h1's across the app

### DIFF
--- a/components/globals/card/Card.tsx
+++ b/components/globals/card/Card.tsx
@@ -1,14 +1,44 @@
 import React from "react";
 
+export enum HeadingLevel {
+  H1 = "h1",
+  H2 = "h2",
+  H3 = "h3",
+  H4 = "h4",
+  H5 = "h5",
+  H6 = "h6",
+}
+
 interface CardProps {
   children?: React.ReactNode;
   icon?: JSX.Element;
   title?: string;
   content?: string;
+  heading?: HeadingLevel;
+  headingStyle?: string;
 }
 
 export const Card = (props: CardProps) => {
-  const { children, icon, title, content } = props;
+  const { children, icon, title, content, heading, headingStyle } = props;
+
+  function createTitle(
+    heading: HeadingLevel | undefined,
+    content: string | undefined
+  ): React.JSX.Element {
+    if (heading && content) {
+      return React.createElement(
+        heading,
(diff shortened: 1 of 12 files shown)


components/globals/card/Card.tsx
```diff
@@ -1,14 +1,44 @@
...
 export const Card = (props: CardProps) => {
-  const { children, icon, title, content } = props;
+  const { children, icon, title, content, heading, headingStyle } = props;
+
+  function createTitle(
+    heading: HeadingLevel | undefined,
+    content: string | undefined
+  ): React.JSX.Element {
+    if (heading && content) {
+      return React.createElement(
```
thiessenp-cds, line 29: `createElement` is part of the [React legacy API](https://react.dev/reference/react/createElement) but as far as I know these API functions aren't going anyways

thiessenp-cds:
Ran into some Git weirdness - recreating on a new PR

PR #2974, opened 2023-11-30
(fix) Close DownloadDialog on error

### LINKED ISSUE #2922: Error while downloading appears behind download modal
# Issue / opportunity

When trying to download hundreds of responses at once, there seems to be a lag and eventually an error that appears on screen. However this shows up behind the modal which remains open and does not provide feedback to users on what's happening.

# Screenshot

Screenshot 2023-11-28 at 10 51 49 AM

dsamojlenko:
# Summary | Résumé

Fixes #2922 
When there is an error while downloading, an error is bubbled up to the DownloadTable and rendered there, but the Dialog remains open covering the error message.

This PR closes the dialog on error after setting the error state.


thiessenp-cds (approved):
LGTM :)

PR #2988, opened 2023-12-01
fix: Download receipt first

### LINKED ISSUE #2933: Receipt shows up at the bottom of the list f files, it should show up at the top
Because the response files start with a date, and the receipt starts with R alphabetically, it shows up below all the responses in the file folder. Perhaps add an asterisk to the name or other character so that it appears above all the responses in the folder.

dsamojlenko:
# Summary | Résumé

Fixes #2933 
When downloading the Individual HTML Files with receipt, add a `_` prefix to the receipt filename so it appears at the top of the downloaded files.

When downloading CSV or JSON formats, ensure the receipt is delivered first, in case the multiple-file-download fails.

### DIFF
--- a/components/form-builder/app/responses/Dialogs/DownloadDialog.tsx
+++ b/components/form-builder/app/responses/Dialogs/DownloadDialog.tsx
@@ -97,7 +97,7 @@ export const DownloadDialog = ({
         });
 
         const zip = new JSZip();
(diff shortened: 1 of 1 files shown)


thiessenp-cds (approved):
Works great. LGTM :)

thiessenp-cds (approved):
LGTM :)

PR #2989, opened 2023-12-01
fix: run promises synchronously when making requests to DynamoDB

craigzour:
# Summary | Résumé

- make DynamoDB requests run synchronously


bryan-robitaille (approved):
👍

bryan-robitaille (approved):

PR #2993, opened 2023-12-01
feat: Log GA event when downloading to record selected format

dsamojlenko:
# Summary | Résumé

Adds a GA logging event when downloading to log the selected format, form id, number of records, and a timestamp.


thiessenp-cds (approved):
Can confirm the code logic. LGTM :)

To test whether the event is fired though, someone with access to the Google analytics account will need to test this.

PR #3002, opened 2023-12-06
feat: Render Responses page navigation as tabs

dsamojlenko:
# Summary | Résumé

Updates the styling of the navigation on the Responses page to appear more like tabs.

image


thiessenp-cds (approved):
This is great, works as expected. LGTM :)

In another PR we can add aria roles like role=tab and role=tabpanel and key controls etc.

PR #3005, opened 2023-12-07
feat: Add pagination to Responses

dsamojlenko:
# Summary | Résumé

Adds Pagination to Responses table. A few things to note:

Responses are ordered Newest->Oldest, so if newer responses come in after you've navigated forward, when you navigate back to the first page, it will reset the page count and new responses will appear at top.

It might be easier to see what's going on if you set the DownloadLimit to a lower number.

The `pages` query param is a Base64-encoded list of ResponseIDs that correspond to the `lastEvaluatedKey` of each "page" or responses. If this query param is modified so that it is invalid, we'll just redirect you back to the start.

### Other changes
- Fixed alignment of "Check all" header checkbox
- Fixed colour of selected rows
- Removed "You have more than xx ... responses" message if there is only one page of results
- Only show dateCreated on all tabs in DownloadTable

~~**Known bug:** If your last page has exactly the same number of responses on it as the maxDownloadLimit, we have no way of knowing that you're on the last page, so the `Next` link still appears, and when you click it you get taken to the "No responses" page. Looking for a solution.~~ bug squashed

image


components/form-builder/app/responses/DownloadTableReducer.ts
```diff
@@ -33,7 +32,7 @@ export const initialTableItemsState = (
   return {
     checkedItems: new Map(),
     statusItems: new Map(vaultSubmissions.map((submission) => [submission.name, false])),
-    sortedItems: sortVaultSubmission(vaultSubmissions),
+    sortedItems: vaultSubmissions,
```
timarney, line 35: Does it make sense to still call this sortedItems?

components/form-builder/app/responses/Pagination.tsx
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  // Keep track of the start point for each of our pages.
+  // To help when determining whether we're at the beginning or end,
+  // the first item in the array is always "start" and the last item is always "end"
+  const [pages, setPages] = React.useState<string[]>(["start"]);
+
+  // Update our pages state when the query changes
+  useEffect(() => {
+    setPages(router.query.pages ? String(router.query.pages).split(",") : ["start"]);
```
timarney, line 39: Might be easier to read if we pull pages off into a variable as your using it multiple times.

components/form-builder/app/responses/Pagination.tsx
```diff
@@ -0,0 +1,110 @@
...
+  // Need statusQuery when building up the prev/next links
+  let statusQuery;
+  if (router.query.params) {
+    [, statusQuery] = router.query.params;
+  }
+
+  // Keep track of the start point for each of our pages.
+  // To help when determining whether we're at the beginning or end,
+  // the first item in the array is always "start" and the last item is always "end"
+  const [pages, setPages] = React.useState<string[]>(["start"]);
```
timarney, line 35: Should this be called something other than pages ?  Your keeping track of codes correct?  Vs what I would typically think of as a page.

lib/vault.ts
```diff
@@ -121,10 +121,15 @@ const submissionTypeExists = async (ability: UserAbility, formID: string, status
...
   formID: string,
-  status?: VaultStatus
-): Promise<{ submissions: VaultSubmissionList[]; submissionsRemaining: boolean }> {
+  status?: VaultStatus,
+  lastEvaluatedKey: Record<string, any> | null | undefined = null,
+  scanForward = true
+): Promise<{
+  submissions: VaultSubmissionList[];
+  submissionsRemaining: boolean;
+  lastEvaluatedKey: Record<string, any> | null | undefined;
```
timarney, line 130: Does we need to use `any` here --- it's a string isn't it?

components/form-builder/app/responses/DownloadTableReducer.ts
dsamojlenko: yeah not really anymore, will update

components/form-builder/app/responses/Pagination.tsx
dsamojlenko: good q ... thinking on a new name 🤔

components/form-builder/app/responses/DownloadTableReducer.ts
dsamojlenko: `allItems`

components/form-builder/app/responses/Pagination.tsx
timarney: pageKeys ?

dsamojlenko: renamed to keys

timarney (approved):

PR #3007, opened 2023-12-08
Adds initial tabs semantics

thiessenp-cds:
# Summary | Résumé

To break up the work for creating a reusable tabs component/pattern here is one way to break down the work:

### 1. Semantics (this PR)
This PR will be about adding the base ARIA attributes to follow the tab pattern.

### 2. Behaviour
Add the keyboard controls

### 3. Performance
Improve the loading, async or server-side. Probably start with some measurements.

### 4. Make it a pattern
Start looking at places in the app where this tabs component/pattern can be reused. Refactor the tabs as needed and updated the app


thiessenp-cds:
Will come back to this later

PR #3019, opened 2023-12-11
chore: with the new infra update LOCAL_LAMBDA_ENDPOINT is now replaced by LOCAL_AWS_ENDPOINT

craigzour:
# Summary | Résumé

- With the new infra update we can now only use specific AWS endpoint for the whole application


thiessenp-cds (approved):
LGTM :)

PR #3075, opened 2024-01-04
feat: Add Element Dialog Refresh

dsamojlenko:
# Summary | Résumé

Refreshed Add Element Dialog, including new Filters + improved examples.

Includes a new package to handle "roving index" navigation, `react-roving-tabindex`. In future we may consider using this package in other places where we've created custom roving indexes, for example: the richText Toolbar, the listbox in the addElement dialog, etc.

| image | image |


components/form-builder/app/edit/elements/element-dialog/tests/ElementDialog.test.js
```diff
@@ -26,15 +26,15 @@ describe("Element dialog", () => {
 
     expect(button).toHaveTextContent("addElement");
 
-    await user.click(button);
+    await user.click(button); // huzzah
```
timarney, line 29: :) remove 

// huzzah

timarney:
Couple of things:

Looks like "Page Text" "element" reads "Large text" for the heading ... seems a little confusing maybe?

Repeating sets --- the heading styling doesn't seem to match the other examples

components/form-builder/app/edit/elements/element-dialog/tests/ElementDialog.test.js
dsamojlenko: lol

thiessenp-cds:
I did some quick testing with screen readers. IMO this could be difficult to fix and could be left as a known bug? If so, @connorscarolyns is there somewhere we could note the below? We could reference it as a known bug if asked about it?

Inconsistent behaviour between screen readers when opening the add element dialog and the initial focussed element.
**Windows**
- **_Edge+Jaws initial focus is on the All filter text (not button) so must tab to focus button etc._**
- Edge+NVDA initial focus on All button filter 
- FireFox+NVDA initial focus on list blox element
- FireFox+Jaws initial focus on list blox element

**MacOs**
Chrome+VoiceOver initial focus on list block element

timarney (approved):
Approving - but hold for #ally discussion

thiessenp-cds:
> > I did some quick testing with screen readers. IMO this could be difficult to fix and could be left as a known bug? If so, @connorscarolyns is there somewhere we could note the below? We could reference it as a known bug if asked about it?
> > Inconsistent behaviour between screen readers when opening the add element dialog and the initial focussed element. **Windows**
> > 
> > * **_Edge+Jaws initial focus is on the All filter text (not button) so must tab to focus button etc._**
> > * Edge+NVDA initial focus on All button filter
> > * FireFox+NVDA initial focus on list blox element
> > * FireFox+Jaws initial focus on list blox element
> > 
> > **MacOs** Chrome+VoiceOver initial focus on list block element
> 
> Oh, interesting finding @thiessenp-cds . Do you have any suggestions on an approach for this? We can fix this at a later time if you don't feel like it's too much of a blocker for AT users.

 I think it's an important bug but not a blocker. Jaws is the most popular screen reader so this is definitely important. But AT users are use to working around issues like this and I'm pretty sure could find a way around it e.g. just tabbing once to get the context of where they are. So agreed, I think we can fix this at a later time.

thiessenp-cds (dismissed):
Nice work! I really like the clean design and implementation. LGTM also :)

timarney (approved):

PR #3104, opened 2024-01-09
Adds fix for removing street-address

thiessenp-cds:
# Summary | Résumé

Based on the original bug report, the form-builder street-address was updated. Previously `autoselect="street-address"` was used for a Street Address input field. To be valid HTML a `street-address` must be on a textarea (multi-line element). The fix is probably temporary since we may want to have a multi-line  (textarea) in the future.

TODO
 - fix tests probably failing
 - test loading old form templates in the form builder that have the old `street-address`

### DIFF
--- a/components/form-builder/util.ts
+++ b/components/form-builder/util.ts
@@ -200,7 +200,6 @@ export const autoCompleteFields = [
   "honorific-prefix",
   "honorific-suffix",
   "organization-title",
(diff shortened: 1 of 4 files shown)


craigzour (dismissed):
LGTM :)

dsamojlenko (approved):

Issue #3112, opened 2024-01-11
Changes to Locked Blocks doesn't prompt "Save Draft"


Abi-Nada:
### Description

If I were to update the description field, privacy notice, or confirmation message while I build my form, it doesn't allow me to save those changes.

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNVc3QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0f71fb609aef2f82b0c86fb5b69f0fd9828bcdb9/Screen%20Recording%202024-01-11%20at%2010.00.18%20AM.mov)

PR #3121, opened 2024-01-16
fix: update template subscribe

timarney:
# Summary | Résumé

Fixes https://github.com/cds-snc/platform-forms-client/issues/3112

Update to monitor changes to description fields and force provider to re-render to make the save button available.


thiessenp-cds (approved):
Nice, my first thought was to test to see if translations were also updating the "save" button. That's working well and the updates are still being announced for a screen reader well. My only thought looking through the code was that if the state compare logic becomes more complicated a reducer might help. But right, LGTM :)

PR #3126, opened 2024-01-18
chore: question selector dropdown fix

timarney:
# Summary | Résumé

- Fixes question selector for modal rules

1) Fix for Richtext field to use description vs title
2) Fix to use correct language key
3) Fix to filter out currently selected question.


thiessenp-cds (approved):
LGTM :)

PR #3139, opened 2024-01-24
feat: Managed data components + Combobox

dsamojlenko:
# Summary | Résumé

## "Managed data" components.
Managed Data components are list-based components (ie, dropdown) where the data is loaded from a managed list. The current managedData lib simply loads a local .json file, but in future could pull in data from an external API or other data source.

Managed data components rely on a new Template Schema property `managedChoices`.  When `managedChoices` is set, the choices array is populated from the `managedData` lib which can parse local json files and return them as an array of `PropertyChoices`.

## Searchable list component
This PR includes a new frontend form component called "Searchable list" - an "accessible-autocomplete" or "typeahead" style combobox/listbox component that allows searching through a list of options.

This new combobox component relies on the [Downshift](https://www.downshift-js.com/use-combobox/) useCombobox hook.

## Departments component
For the first Managed data component, we've combined a pre-populated list of Departments with the new "Searchable list" component.

This component makes use of the new managed data concept, the Searchable list component, and uses the existing `form-builder-templates` to tie it all together.

There is currently no automated mechanism to update the departments.json, it's just a straight export from the [gc-organizations repo](https://github.com/cds-snc/gc-organisations). This particular list does not update often, but this could be an area for future development.

## Component tests
Had a difficult time figuring out the interactivity in the Jest test, so kept that basic and added a more detailed Cypress Component test.

| Searchable list                                          | Departments                                        |
| image | image |


connorscarolyns:
There are still a few copy adjustments before this can go live. The text in the add modal does not reflect the text in the template. The department list should display departments. The hint text should match. I would also like to have a different icon for departments. An office building icon would work.

lib/managedData/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  return departments
+    .filter((item) => {
+      return item.type === type;
+    })
+    .map((item) => ({
+      en: item.name_eng,
+      fr: item.name_fra,
+    }));
+};
+
```
thiessenp-cds, line 20: Looks like a candidate for a reducer - could be a future todo

components/clientComponents/form-builder/app/edit/elements/element-dialog/descriptions/Combobox.tsx
```diff
@@ -0,0 +1,29 @@
+"use client";
+import React from "react";
+import { useTranslation } from "@i18n/client";
+import { ExampleWrapper } from "./ExampleWrapper";
+import { Combobox as ComboboxComponent, Description, Label } from "@clientComponents/forms";
+
```
thiessenp-cds, line 6: FYI Note: The dropdown can be a little verbose announcing the entire list on each added/removed character. But I think this is necessary to both announce all dynamic updates and keep the user informed of all the choices at each state. We could think about ways to reduce the verbosity if user testing shows this overwhelms users.

Also, at first glance it looks like the generated reference from the input to the label is broken e.g. `downshift-:r18:-label` vs `label-1`. Even though that is an invalid query selector (I think?) from the library, this still works.

thiessenp-cds (approved):
This tests really well with screen readers. The label and description being announced when focusing the input really help make it clear where you are. Also the dropdown list updates being announced are very helpful. Plus it follows ARIA patters/BPs. I also added a few minor comments. An excellent component!
LGTM :)

components/clientComponents/forms/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,75 @@
...
+        );
+      },
+      items,
+      onSelectedItemChange({ selectedItem }) {
+        setValue(selectedItem);
+      },
+    });
+
+  return (
+    <>
```
timarney, line 39: We should be able to remove this fragment yes?

lib/formBuilder.tsx
```diff
@@ -222,6 +231,19 @@ function _buildForm(element: FormElement, lang: string): ReactElement {
         />
       );
     }
+    case FormElementTypes.combobox:
+      return (
+        <div className="focus-group">
+          {labelComponent}
+          {description && <Description id={`${id}`}>{description}</Description>}
```
timarney, line 238: For a bunch of these does this not work? vs the string wrapping?

```
 id={id}
 ```

File: lib/managedData/index.ts [hunk 5]
```diff
@@ -0,0 +1,32 @@
...
+  CROWN_CORPORATIONS = "crownCorporations",
+  PROVINCIAL_TERRITORIAL = "provincialTerritorial",
+}
+
+const filterDepartmentsData = (type: string) => {
+  return departments
+    .filter((item) => {
+      return item.type === type;
+    })
+    .map((item) => ({
```
timarney, line 15: ({name_eng,name_fra,}) bit clean if we destructure here ?

timarney:
Any reason not to make this wider or maybe add an ...
Screenshot 2024-02-08 at 12 50 39 PM

timarney:
Screenshot 2024-02-08 at 12 53 07 PM

Bit of an edge case but there's some weirdness when toggling the language.

lib/formBuilder.tsx
dsamojlenko: eh, yeah... think that was some codepilot completion... will fix

dsamojlenko: image
er

components/clientComponents/forms/Combobox/Combobox.tsx
dsamojlenko: yup

lib/managedData/organizations.ts
```diff
@@ -0,0 +1,42 @@
...
+  PTM = "PTM",
+}
+
+export const organizations: ManagedDataSet<Organization[]> = {
+  values: organizationsData,
+  filters: {
+    departments: (values: Organization[]) =>
+      values
+        .filter((item) => item.type === OrganizationType.GCDepartment)
+        .map((item) => ({
```
timarney, line 23: Maybe try this

```javascript
.map(({ name_eng: en, name_fra: fr }) => ({ en, fr })),
```

dsamojlenko: yeah, that's nicer

timarney (approved):
Great work on this @dsamojlenko

PR #3140, opened 2024-01-24
fix: update page text rule indicator

timarney:
# Summary | Résumé

Small update to fix "page text" indicator

Before:

After:

Screenshot 2024-01-24 at 12 22 44 PM


craigzour (approved):
LGTM!

PR #3146, opened 2024-01-25
chore: update support error message logs

timarney:
# Summary | Résumé

Add more detailed log message for support ticket 400 errors

### DIFF
--- a/pages/api/id/[form]/submission/report.ts
+++ b/pages/api/id/[form]/submission/report.ts
@@ -278,7 +278,7 @@ async function notifySupport(
     language,
   });
 
-  if (result.status >= 400) {
+  if (result?.status >= 400) {
     throw new Error(
       `Freshdesk error: ${JSON.stringify(result)} - ${userEmailAddress} - ${description}`
     );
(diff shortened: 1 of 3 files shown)


pages/api/id/[form]/submission/report.ts
```diff
@@ -278,8 +278,10 @@ async function notifySupport(
     language,
   });
 
-  if (result && result?.status >= 400) {
-    throw new Error(`Freshdesk error: ${result.status}`);
+  if (result.status >= 400) {
```
Comment (wmoussa-gc) [inline pages/api/id/[form]/submission/report.ts:281]: It would be beneficial to create a variable named "BAD_REQUEST" for the HTTP status code 400. In the case of a bad request, should we log the request?

wmoussa-gc (dismissed):

pages/api/id/[form]/submission/report.ts
Comment (timarney) [reply pages/api/id/[form]/submission/report.ts]: According to https://support.freshdesk.com/en/support/solutions/articles/225440-possible-error-codes-and-its-reasons

Any errors should be 400+

These are already logging but we're not getting enough info 
Screenshot 2024-01-25 at 9 52 16 AM

wmoussa-gc (approved):

PR #3183, opened 2024-01-30
chore: Update branding options

timarney:
# Summary | Résumé

- Adds branding option for Elections Canada
- Adds branding option for of Canada Secretariat


craigzour (dismissed):
Found a small typo but the rest looks good :)

craigzour (approved):

PR #3205, opened 2024-01-31
fix: validation when conditional rules array is empty

timarney:
# Summary | Résumé

Fixes an issue where when using conditional rules ... elements with empty rules arrays we're bypassing validation checks.


craigzour (approved):
LGTM!

wmoussa-gc (approved):

PR #3208, opened 2024-01-31
fix: Sign-in flow not working properly when getting the Restricted access to responses dialog

wmoussa-gc:


craigzour (approved):
It works!

PR #3214, opened 2024-02-01
Adds comments noting apis no longer used

thiessenp-cds:
# Summary | Résumé

Comments unused apis

### DIFF
--- a/app/api/id/[form]/apiusers/apiusers.ts
+++ b/app/api/id/[form]/apiusers/apiusers.ts
@@ -1,3 +1,10 @@
+/**
+ * This is part of our old API calls when we used to support API downloading of the responses.
+ * It was left in for historical purposes.
+ *
+ * TODO: potentially remove.
+ */
+
 // import { NextApiRequest, NextApiResponse } from "next";
 // import { prisma, prismaErrors } from "@lib/integration/prismaConnector";
 // import { Prisma } from "@prisma/client";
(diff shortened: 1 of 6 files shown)


craigzour:
I wonder if we should not just delete them to be honest. What do you think @bryan-robitaille ? 
No matter what we would have to rewrite them in order to use the new way APIs are defined.

bryan-robitaille:
> I wonder if we should not just delete them to be honest. What do you think @bryan-robitaille ? No matter what we would have to rewrite them in order to use the new way APIs are defined.

Let's note it and create an issue for a follow up PR after the big merge.  The scope is rather large where we'll want to ensure that we remove not only the endpoints but all components that were also unique to that functionality.  We'd also need a DB migration to drop the API Users table and associated referenced fields.

craigzour:
> > I wonder if we should not just delete them to be honest. What do you think @bryan-robitaille ? No matter what we would have to rewrite them in order to use the new way APIs are defined.
> 
> Let's note it and create an issue for a follow up PR after the big merge. The scope is rather large where we'll want to ensure that we remove not only the endpoints but all components that were also unique to that functionality. We'd also need a DB migration to drop the API Users table and associated referenced fields.

Sounds good! I will create a ticket. cc @thiessenp-cds

craigzour (approved):

bryan-robitaille:
API removed in PR #3503

Issue #3199, opened 2024-01-31
Warning: Failed prop type: Invalid prop `children` of type `array` supplied to `AdminNavLayout`, expected `object`.     at AdminNavLayout


wmoussa-gc:
## Description
While debugging locally found in the log
```Warning: Failed prop type: Invalid prop `children` of type `array` supplied to `AdminNavLayout`, expected `object`.
    at AdminNavLayout```

## Steps to reproduce
1. Sign In 
2. Click on Your Account
3. Scroll down to Administration
4. See error in console

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**:  local
- **Language**: English

## Expected behaviour
No warning error

## Screenshots or videos
```
DEBUG: Using Cached User Status for auth:active:clqwslxzg0002pxyslnjd61li
DEBUG: JWT refreshed for user wissam.moussa@cds-snc.ca
DEBUG: Using Cached Privileges for auth:privileges:clqwslxzg0002pxyslnjd61li
DEBUG: Session refreshed for user wissam.moussa@cds-snc.ca
DEBUG: Privilege Check PASS: Can view on User 
Warning: Failed prop type: Invalid prop `children` of type `array` supplied to `AdminNavLayout`, expected `object`.
    at AdminNavLayout

```

timarney:
@wmoussa-gc are you still seeing this error I don't seem to be able to re-create

wmoussa-gc:
It comes and goes; not sure what's happening under React's hood. If you keep navigating from one admin screen to another, it sometimes shows up and then disappears.

``` ○ Compiling /[locale]/admin/settings ...
 ✓ Compiled /[locale]/admin/settings in 1465ms (5470 modules)
(node:93461) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 exit listeners added to [process]. Use emitter.setMaxListeners() to increase limit
(Use `node --trace-warnings ...` to show where the warning was created)
WARN: NextAuth warning - Code: debug-enabled
DEBUG: Using Cached User Status for auth:active:clqwslxzg0002pxyslnjd61li
DEBUG: JWT refreshed for user wissam.moussa@cds-snc.ca
DEBUG: Using Cached Privileges for auth:privileges:clqwslxzg0002pxyslnjd61li
DEBUG: Session refreshed for user wissam.moussa@cds-snc.ca
DEBUG: Privilege Check PASS: Can view on Setting 
[...]

Issue #3173, opened 2024-01-29
French-first editing when using French interface


anikbrazeau:
## Description
When using the application in English, editing in English is the norm and expected. When using the application in French, it might reasonably be expected that the working language is French and for features to be French-first to match. Interface language shoudl match the language default of app features.

## Steps to reproduce
1. Sign in using French interface ("Se connecter")
2. Go to "Modifier" (Edit) in left nav
3. Land on form building interface
4. See "Modifier en: **anglais**" toggle at the top of the screen
5. Click "Ajouter un élément" to add a block
6. Select "Noms individuels" for separate names
7. The element is added in English

## Expected behaviour
Expected to start editing in French.

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: App router PR
- **Language**: French

## Screenshot
<img width="1410" alt="Screenshot 2024-01-29 at 2 48 54 PM" src="https://github.com/cds-snc/platform-forms-client/assets/38330843/31598f3b-5d40-4d7f-8232-a72de359bdf1">

wmoussa-gc:
fixed..Ill wait for Anik to confirm

anikbrazeau:
Fixed! :D Thank you!

Issue #3160, opened 2024-01-29
Deleting a form fails


thiessenp-cds:
Attempting to delete a form results in an error. I suspect this is a CSP related error.

https://github.com/cds-snc/platform-forms-client/assets/107579368/c2d6e0eb-4f45-4da3-ab8f-7602bda35ac6

timarney:
Closing this as the PR to fix is merged --- can re-open as needed.

PR #3221, opened 2024-02-01
fix: Refactor logic to download form schema from the ConfirmFormDeleteDialog

wmoussa-gc:
# Summary | Résumé
Closes

- https://github.com/cds-snc/platform-forms-client/issues/3160 
- https://github.com/cds-snc/platform-forms-client/issues/3199 
- https://github.com/cds-snc/platform-forms-client/issues/3173

The ConfirmFormDeleteDialog was using the DownloadFileButton, which relies on the Template Store context. 

image

Presently, this context is injected only in the Form-Builder layout, not in the My Forms or Admin Layout configuration. And, the ConfirmFormDeleteDialog is exclusively used in the last two groups (tbc).

Solution: Altered the logic in the dialog by incorporating an inline button to download the form schema from the backend instead of relying on the store.

Find a form to delete, the dialog and download a copy should work.


timarney (approved):

dsamojlenko:
Noting - this fixes that "Missing Template Store Provider" error, but there seems to be another (unrelated?) bug- after the dialog closes, the /forms screen does not refresh so it appears as if the form wasn't deleted, have to do a hard refresh to clear it.

wmoussa-gc:
@dsamojlenko ill merge this PR and open a new issue

PR #3245, opened 2024-02-07
fix: Cypress tests for Server Side Rendering 

bryan-robitaille:
# Summary | Résumé

- Updates Cypress tests for current App Router build.
- Remove flag for FormTimer
- Remove flag for ReCaptcha
- Update yarn to 4.1.0
- Add pre-stage git hook to format prettier
- Add client side accessible env var (NEXT_PUBLIC_APP_ENV) to disable form timer in tests
- Adds missing locale to some internal links
- Disable link prefetching on Language switcher (no need to prerender the exact page in different language)
- Adds some debug statements to help figure out what is redirecting a user.  Will remove during big refactor coming up.
- Wrap scripts injected in Head with Suspense.  This is due to a Cypress issue injecting it's own and causing react SSR mismatch hydration errors.
- Add a new package to help debug Cypress tests.  It allows to show browser console messages in the terminal when running Cypress headless.  `yarn cypress:run`


app/layout.tsx
```diff
@@ -52,19 +53,34 @@ export default async function Layout({ children }: { children: React.ReactNode }
...
-        <Script
-          nonce={nonce}
-          async
-          type="text/javascript"
-          src="/static/scripts/form-polyfills.js"
-        />
-        {NoIndexMetaTag}
-        <Script id="GoogleTagManager" nonce={nonce} async type="text/javascript">
-          {googleTagManager}
-        </Script>
```
thiessenp-cds, line 67: I spent days trying to find a work around for this. Awesome.

thiessenp-cds (commented):
FYI:
I did a quick manual test (happy paths) and noticed an error. "ERROR for site owner:
Invalid domain for site key". Searching for that string online, looks like it may be a Captcha related error. If so, looking at layout.tsx I would think this is just an error related to `RECAPTCHA_V3_SITE_KEY ` not matching prod or similar. Either way not a big deal, especially with the future refactoring.
Screenshot 2024-02-12 at 3 20 39 PM

thiessenp-cds (approved):
This is truly impressive and awesome you got cypress running without some weird hack etc. and even most the tests to past!!! LGTM :)

bryan-robitaille:
@thiessenp-cds that error is because you don't have the recaptcha key in your .env file.  I thought I put logic in there to not load recaptcha if the key wasn't available.  I'll double check tomorrow.

Issue #3246, opened 2024-02-08
String ID showing up rather than text itself


anikbrazeau:
## Description
In the "More" modal for a question, the hint text heading shows up as `inputDescription` rather than `Hint text` even though the string appears to be defined in the json locale file?

## Steps to reproduce
1. Go to GC Forms
2. Click on Create a form > Edit > Add form element
3. When the question is selected, click "More"
4. See missing string

## Details
- **Browser**: Edge
- **Operating system**: Windows
- **Environment**: Staging
- **Language**: English AND French
- **Viewport**: Tablet

## Expected behaviour
"Hint text" in English
"Texte explicatif" in French

## Screenshots or videos

![inputdescription](https://github.com/cds-snc/platform-forms-client/assets/38330843/09622c98-df07-4a4e-80fa-e8ccf18209e8)

<img width="1347" alt="Screenshot 2024-02-08 at 2 48 06 PM" src="https://github.com/cds-snc/platform-forms-client/assets/38330843/36a35af5-5fd8-4ce3-98af-316c2a738469">

PR #3247, opened 2024-02-09
Fix: String ID showing up rather than text itself

timarney:
# Summary | Résumé

Fixes: #3246 

Fixes typo


craigzour (approved):
You won!

bryan-robitaille (approved):

PR #3250, opened 2024-02-13
feature: Administration Pages Layouts

bryan-robitaille:
# Summary | Résumé

Moves Administration pages to dedicated layouts and converts some global components to server components.
Updates left nav to new planned designs

image

image

image

image


thiessenp-cds (approved):
I know the layouts are a bit "busy" with more code but I find it a lot easier to follow vs. the deeply nested components.
I gave this a quick test also. The happy paths look good. Nice! LGTM :)

Issue #3264, opened 2024-02-20
Audit Log - User Sign In event using wrong user identifier


bryan-robitaille:
## Description
A User Sign in event is storing the sub returned from Cognito as the user id instead of the user id from the User table.

## Steps to reproduce
1. Sign into the App
2. Check the logs in the AuditLog DynamoDB
3. User ID partition key is different for User Sign in Action vs other actions after signin

## Expected behaviour
Audit Log should use the User ID from the User table in the application.

## Screenshots or videos
![Screenshot 2024-02-20 at 10 26 50 AM](https://github.com/cds-snc/platform-forms-client/assets/25329319/8533840d-0368-4e6c-a4db-7dc8d8c1a971)

## Additional context
Any other information about the problem here.

PR #3266, opened 2024-02-21
fix: retrieve internal user id instead and pass it to the UserSignIn audit log instead of using the Cognito sub id

craigzour:
# Summary | Résumé

closes https://github.com/cds-snc/platform-forms-client/issues/3264

- Fixed issue where the wrong user id was used in the `UserSignIn` audit log. It used to be the Cognito sub id and it will now be our internal user id from PostgreSQL


pages/api/auth/[...nextauth].ts
```diff
@@ -115,7 +115,22 @@ export const authOptions: NextAuthOptions = {
...
     async signIn({ user }) {
-      logEvent(user.id, { type: "User", id: user.id }, "UserSignIn");
+      if (!user.name || !user.email)
+        throw new Error(
+          "Could not produce the UserSignIn audit log because of missing name and/or email"
+        );
+
+      const internalUser = await getOrCreateUser(user.name, user.email, user.image);
+
+      if (internalUser === null)
```
Comment (wmoussa-gc) [inline pages/api/auth/[...nextauth].ts:125]: we would expect that if calling getOrCREATEUser we would never get a null..

lib/users.ts
```diff
@@ -16,11 +16,11 @@ import { activeStatusUpdate } from "@lib/cache/userActiveStatus";
  * Get or Create a user if a record does not exist
  * @returns A User Object
  */
-export const getOrCreateUser = async ({
-  name,
-  email,
-  picture,
-}: JWT): Promise<{
```
wmoussa-gc, line 23: to clean up the import

pages/api/auth/[...nextauth].ts
Comment (craigzour) [reply pages/api/auth/[...nextauth].ts:125]: The main reason why it can return a `null` value is because of that `prismaErrors` function that catches specific database internal errors and may end up returning a default value which in our case is a `null`.
I can not remember why we did it this way instead of using a throw/catch pattern.
cc @bryan-robitaille

bryan-robitaille (changes requested):
Let's remove the getOrCreateUser and replace it with a direct primsa call that fetches the user ID based on email.  This ensures that the scope of the logging function can't create a user and can only read a user.

wmoussa-gc (approved):

bryan-robitaille (approved):

PR #3268, opened 2024-02-21
chore: added new migration script to make existing audit logs archivable

craigzour:
# Summary | Résumé

- Added new migration script to be executed in both Staging and Production to make existing audit logs archivable.

# Testing

- This script can be tested locally. Just make sure you create audit logs without that new `Status` field first.

The AWS DynamoDB console gives us an idea about the number of audit logs that we currently have in our table and that we will most likely have to migrate:
- Staging: 437,030
- Production: 357,090


bryan-robitaille (approved):
Looks good to me.

PR #3296, opened 2024-02-28
Template api to server action

dsamojlenko:
# Summary | Résumé

Converts useTemplateApi to server actions and updates related components.


timarney (approved):
Nice work.

PR #3302, opened 2024-03-04
Feature/support pages

thiessenp-cds:
# Summary | Résumé

Refactors pages Contact, Support, and Unlock-Publishing to use server actions.

A followup PR will investigate using Zod for form validation.

For testing, I'd wait for the next PR with the potential Zod updates to reduce the number of test tickets for the support team to see.


app/(gcforms)/[locale]/(support)/support/actions.ts
```diff
@@ -0,0 +1,66 @@
...
+  request?: string;
+  description?: string;
+  language?: string;
+}
+
+export async function support({ name, email, request, description, language = "en" }: Submission) {
+  // No auth etc. checking since this is a public endpoint
+
+  //Mandatory fields
+  if (!name || !email || !request || !description) {
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(support)/support/actions.ts:17]: server actions is intelligent enough that it will throw back an error to the calling function if a required field is not present.  If you modify the type to remove the undefined status from known required fields we won't need to do a "malformed request" check.

app/(gcforms)/[locale]/(support)/support/page.tsx
```diff
@@ -14,6 +16,39 @@ export async function generateMetadata({
...
+          <p className="mb-14">
+            {t("support.gcFormsTeamPart1")}{" "}
+            <Link href={`https://www.canada.ca/${language}/contact.html`}>
+              {t("support.gcFormsTeamLink")}
+            </Link>{" "}
+            {t("support.gcFormsTeamPart2")}
+          </p>
+          <Alert.Warning title={t("support.lookingForADemo")} role="note">
+            <p>
+              {t("support.ifYouWouldLike")} <Link href={`/contact`}>{t("support.contactUs")}</Link>.
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(support)/support/page.tsx:44]: Please include the language in the href

app/(gcforms)/[locale]/(support)/unlock-publishing/actions.ts
```diff
@@ -0,0 +1,63 @@
...
+  managerEmail?: string;
+  department?: string;
+  goals?: string;
+  language?: string;
+}) {
+  const session = await auth();
+  if (!session) throw new Error("No session");
+
+  //Mandatory fields
+  if (!managerEmail || !department || !goals) {
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(support)/unlock-publishing/actions.ts:21]: See comment about the need for checking if fields are present in other server action comments.

app/(gcforms)/[locale]/(support)/unlock-publishing/actions.ts
```diff
@@ -0,0 +1,63 @@
...
+  <br/>
+  Adresse email du responsable: ${managerEmail} .<br/>
+  `;
+
+  if (!session.user.name || !session.user.email) {
+    throw new Error("User name or email not found");
+  }
+
+  try {
+    const result = await createTicket({
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(support)/unlock-publishing/actions.ts:46]: Similar to the others, `result.status` will always be 200 in this case.

app/(gcforms)/[locale]/(support)/unlock-publishing/components/client/UnlockPublishingForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+export const UnlockPublishingForm = () => {
+  const {
+    t,
+    i18n: { language },
+  } = useTranslation("unlock-publishing");
+  const router = useRouter();
+  const [errorState, setErrorState] = useState({ message: "" });
+  const [submitting, setSubmitting] = useState(false);
+  const { data: session } = useSession();
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(support)/unlock-publishing/components/client/UnlockPublishingForm.tsx:32]: Instead of calling `useSession` for the current users email address, the value could potentially be passed in from `page.tsx` where there is already a server call to get the user.

bryan-robitaille (changes requested):
A few small tweaks and then it should be stable enough to start exploring Zod.

app/(gcforms)/[locale]/(support)/support/actions.ts
Comment (thiessenp-cds) [reply app/(gcforms)/[locale]/(support)/support/actions.ts:17]: Sounds good. Some time we should talk about "client" and "server" validation and how these patterns have probably changed.
My thinking was to do the main validation on the client and leave this check on the server just for form sake. A message could be returned but a thrown error with a generic error screen seemed good enough - since this should never happen

bryan-robitaille (approved):
Nice!

PR #3314, opened 2024-03-11
Fix: Conditional Indicator

timarney:
# Summary | Résumé

Fixes issue where Question indicators show incorrect question #.

Given a scenario where parent elements have been moved the question indicator was displaying the wrong question #.

This was because the elements being passed to the get the question number wasn't being sorted by layout 

```
 "layout": [
    47,
    35,
    48,
    44,
    2,
    3,
```

The PR ensures the elements are sorted by layout prior to getting the question #.

**Ticket:**
https://cds-snc.freshdesk.com/a/tickets/16890


bryan-robitaille (approved):
👍

Issue #3293, opened 2024-02-28
File Upload Scanner: File validation component


Abi-Nada:
### User Story
As a developer,
I want to fix the file validation component and enhance our file scanning feature,
So that the application can reliably identify file types and ensure that all uploaded files are free from malware, corruption, or malicious content.

### Context:
Initially with this work, we were uploading files and storing them in S3 buckets. These weren't being scanned because we were relying on email delivery/Outlook to perform these checks for us. But now that we're using the Response Database as a delivery response destination, we need to ensure that these files are being scanned to validate that they're clean and not corrupted/malicious.

#### Technical Obstacles
- We currently try to validate the file type during submission however the `mmmagic` library that were previously relied on does not interact well with the newer versions of Node.  

### Out of Scope
Any type of file retrieval of uploaded files is currently out of scope and will be handled in a separate Issue.
### Acceptance Criteria:
- [x] Implement, if possible,  a new file validation function to replace `mmmagic`.  If not possible then we'll need to document and come up with some mitigation strategies.  (Can be performed on the App Router or Develop branch)
- [x] Document where the attribute can be found on an uploaded file that determines if it has succeeded or failed the malicious file scanning process.
- [x] Use latest scan files module (v9.2.3)

craigzour:
WIP How to retrieve and use file scan results https://docs.google.com/document/d/1tUTzEafp3V5hNiD1ghVU1OXRbMlq421x9dyTNoxrHlc

PR #3321, opened 2024-03-12
chore: rework file validation

craigzour:
# Summary | Résumé

part of the work related to https://github.com/cds-snc/platform-forms-client/issues/3293

- Reworked file validation for the file input component. Two types of validation are applied at different stages (client and server)

- Create a form with a file input component
- Try to upload a file with a size larger than 8 MB
- Try to upload a file with an extension that is not allowed (e.g. .zip)
- Try to upload a file with a modified extension (e.g. rename a .zip to .csv)

You can also call the submit API directly to try to bypass the client side validation. Here is an example where I tried to upload a file larger than 8 MB by passing a fake size value:
```
POST
http://localhost:3000/api/submit

HEADERS
X-CSRF-Token: <insert_valid_token>

DATA:
{
    "formID": "<insert_form_id_with_a_single_component_that_is_file_input>",
    "securityAttribute": "Protected A",
    "1": {
        "name": "myfile.txt",
        "size": 100,
        "based64EncodedFile": "<insert_a_base64_encoded_file_that_is_larger_than_8_mb>"
    }
}
```

I used https://base64.guru/converter/encode/file to convert a file to a base64 string (I could not upload the result here in the comments because it is too big).

### DIFF
--- a/__utils__/mocks/file-type/index.ts
+++ b/__utils__/mocks/file-type/index.ts
@@ -1,0 +1,1 @@
+export const fileTypeFromBuffer = (buffer: Uint8Array | ArrayBuffer) => Promise.resolve(undefined);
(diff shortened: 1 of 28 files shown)


bryan-robitaille (approved):
Tested and works and described!
We can continue to refactor as we modify app router as well.

PR #3334, opened 2024-03-14
Feature/support pages

thiessenp-cds:
# Summary | Résumé

Refactors the support pages to use useFormState, useFormStatus and Valibot for validation. Updated pages:
- /support
- /contactus
- /unlock-publishing


tsconfig.json
```diff
@@ -51,7 +51,6 @@
     "cypress",
     "cypress.config.ts",
     "**/*.cy.ts",
-    "**/*.cy.tsx",
-    "__tests__"
```
thiessenp-cds, line 55: This line was causing linting to fail for me. If it works for you, I can add it back

app/(gcforms)/[locale]/(support)/contact/actions.ts
```diff
@@ -1,33 +1,68 @@
...
+  const result = await validate(language, rawFormData);
+
+  if (!result.success) {
+    return {
+      validationErrors: result.issues.map((issue) => ({
+        fieldKey: issue.path?.[0].key as string,
+        fieldValue: issue.message,
+      })),
+    };
   }
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(support)/contact/actions.ts:57]: Another way to do this would be to replace lines 58-64 with line 46

app/(gcforms)/[locale]/(support)/components/client/Checkbox.tsx
```diff
@@ -0,0 +1,32 @@
+"use client";
```
Comment (timarney) [inline app/(gcforms)/[locale]/(support)/components/client/Checkbox.tsx:1]: Are these files a straight copy of existing components --- or what changed.

Comment (thiessenp-cds) [reply app/(gcforms)/[locale]/(support)/components/client/Checkbox.tsx:1]: Any Forms-form component that used Formik state had to be copy+paste+updated. So in the case of Checkbox.tsx, it used the Field component that tied into Formik state e.g. useField().
The idea was to slowly move away from Formik with useFormState by testing on the form-builder side of the app. And lastly updating forms-form

timarney (approved):
Going to approve and we can look to capture issues as they come up from user testing.

With the move to more server action type forms I think there's going to be more tweaks here but this moves things along nicely.

PR #3354, opened 2024-03-18
Convert form submission API to server action

dsamojlenko:
# Summary | Résumé

Converts the `/api/submit` endpoint to a server action.
Refactors a bunch of the code from `/api/submit/route.ts` and `lib/client/clientHelpers.ts` into smaller discrete files in a lib folder next to the new `actions.ts`

Most of this code is unchanged from original, except cases where a generic Error was being thrown, we're now throwing specific custom exceptions.

Currently these custom exceptions bubble up to the Form.tsx component's handleSubmit where we catch, log, and set the form error state.

Note this all functions as before the AppRouter conversion, but without the API call and CSRF token issue.

A future PR could convert the Form/Action to use useFormState to handle passing messages back from the action.


timarney (approved):
Approving to unblock manual testing.

PR #3361, opened 2024-03-19
Fix my form background

timarney:
# Summary | Résumé

Fixes bg color.


thiessenp-cds (approved):
Yup that fixes the issue

PR #3378, opened 2024-03-20
Sorts forms by updated on Forms page

thiessenp-cds:
# Summary | Résumé

This updates the templates lib call to have an optional sort by updated_at time vs. previous created_at time. This helps a user more easily see the template(s) s/he recently updated by being first in the forms cards "list".

The lib templates call was moved out of the component and into its Page file. This was done to avoid the component being cached when a user navigates back to the page. Otherwise, after a user updates a form template and navigates back to the Forms page, s/he would see no order change in the form cards 'list" (component not re-rendered).


app/(gcforms)/[locale]/(form administration)/forms/[[...path]]/page.tsx
```diff
@@ -38,7 +52,36 @@ export default async function Page({
       redirect: true,
     });
 
-    const { t } = await serverTranslation("my-forms");
+    const {
+      t,
+      i18n: { language },
+    } = await serverTranslation("my-forms");
+
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(form administration)/forms/[[...path]]/page.tsx:59]: I added the below comment because I know in the future, if I were to see this, I'd be very tempted to move this call into the Cards component. It just feels cleaner to me. Though I couldn't find a way to stop the component being cached when navigating back to the page.

timarney:
@thiessenp-cds 

FYI - paired with Dave and did some additional work to speed up the loading etc...

PR #3409, opened 2024-03-22
fix: Can't save a Setting with blank Description fields

wmoussa-gc:
to fix https://app.zenhub.com/workspaces/gcforms-60cb8929764d71000e481cab/issues/gh/cds-snc/platform-forms-client/3330


dsamojlenko (approved):
Works great!

PR #3416, opened 2024-03-25
Updates header forms link

thiessenp-cds:
# Summary | Résumé

Updates Header to have less forms :) 
(right Forms link is removed and left Forms text becomes a link)

### DIFF
--- a/components/clientComponents/globals/Header/Header.tsx
+++ b/components/clientComponents/globals/Header/Header.tsx
@@ -18,11 +18,7 @@ type HeaderParams = {
 
 export const Header = ({ context = "default", className }: HeaderParams) => {
   const isFormBuilder = context === "formBuilder";
-  const isAdmin = context === "admin";
-  const isDefault = context === "default";
-
   const { status } = useSession();
-
   const {
     t,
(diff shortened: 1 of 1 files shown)


components/clientComponents/globals/Header/Header.tsx
```diff
@@ -45,34 +41,18 @@ export const Header = ({ context = "default", className }: HeaderParams) => {
...
-          {isAdmin && (
-            <div className="mt-3 box-border block h-[40px] px-2 py-1 text-xl font-semibold">
-              {t("adminNav.allForms", { ns: "admin-login" })}
-            </div>
-          )}
-          {isFormBuilder && status === "authenticated" && (
-            <div className="pt-3">
-              <Link href={`/${language}/forms`}>{t("adminNav.allForms", { ns: "common" })}</Link>
-              <span className="mx-2 inline-block"> / </span>
-            </div>
```
thiessenp-cds, line 63: I made the assumption we only want to show the Forms link if a user is authenticated and only show the input on the form-builder pages

timarney (approved):

Issue #3430, opened 2024-03-27
Sign-in Page - 


timarney:
## Description

Logging some variances vs current staging

Pressing submit without entering values shows multiple errors for the same input

App router

<img width="500" alt="Screenshot 2024-03-27 at 2 35 37 PM" src="https://github.com/cds-snc/platform-forms-client/assets/62242/21588faf-0607-4913-83b3-e4f17ef0d5a7">

Staging:
<img width="500" alt="Screenshot 2024-03-27 at 2 35 57 PM" src="https://github.com/cds-snc/platform-forms-client/assets/62242/cf38a5ea-ee30-4be1-8df3-a1c58cf71b1f">

<hr>

Input fields are indicated with a red border around the input and also indicate the error for the specific field.

App Router
<img width="500" alt="Screenshot 2024-03-27 at 2 36 21 PM" src="https://github.com/cds-snc/platform-forms-client/assets/62242/8f2a98ec-d69d-40cf-ad83-e01729940ae5">

Staging
<img width="500" alt="Screenshot 2024-03-27 at 2 36 12 PM" src="https://github.com/cds-snc/platform-forms-client/assets/62242/74c1db35-3855-4d70-987d-319d532b5ac3">

PR #3433, opened 2024-04-02
fix: Validations passing multiple issues per field and Error highlighting on input fields

bryan-robitaille:
# Summary | Résumé

fixes: #3430 

- Modified Valibot to abort validation pipeline early so only 1 validation error is generated per field.
- Ensured validation errors from form state were being passed down to input components.

Fixes pages: 
- Support
- Unlock Publishing
- Login
- MFA
- Register
- Reset Password
- Security Question Setup
- Profile

| image | image |


wmoussa-gc (approved):
lgtm

Issue #3422, opened 2024-03-27
Deleting a form provides no visual feedback


timarney:
## Description
A clear and concise description of what the problem is.

## Steps to reproduce
1. Go to Manage forms
2. Click on delete form 
3. Confirm the deletion 
4. Delete the form
5. No visual indication that a form has been deleted happens 
6. Form is deleted but the only way to know this is by refreshing the page

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Preview
- **Language**: En

## Expected behaviour

Deleting matches current staging deletion for visual feedback

PR #3442, opened 2024-04-03
fix: Manage Forms page refresh after delete

bryan-robitaille:
# Summary | Résumé

closes: #3422 

- Ensures the data is refreshed when a form is deleted or modified on the manage forms page
- Fixes permissions on the account page.
  -  fixed user view permission so it checks if a user has view all users
  - updated permission to manage all forms to pass to card list.


thiessenp-cds (approved):
Nice work. LGTM :)

PR #3470, opened 2024-04-11
feature: Enhance security on Next-Auth and remove HOST_URL variable

bryan-robitaille:
# Summary | Résumé

This PR does 2 main things:

1. It removes all API endpoints for Next-Auth that are not used within the application.  So only known Next-Auth API calls are passed through:

links in this section

# Pull Request Checklist

### DIFF
--- a/lib/auth/nextAuth.ts
+++ b/lib/auth/nextAuth.ts
@@ -106,7 +106,8 @@ export const {
     // Seconds - How long until an idle session expires and is no longer valid.
     maxAge: 2 * 60 * 60, // 2 hours
   },
-  trustHost: true,
+  // Only trust the host if we don't explicitly have a AUTH_URL set
+  trustHost: process.env.AUTH_URL ? false : true,
   debug: process.env.NODE_ENV !== "production",
   logger: {
     error(code, ...message) {
--- a/lib/origin.ts
+++ b/lib/origin.ts
@@ -1,0 +1,25 @@
+import { headers } from "next/headers";
+
+/**
+ * Extract the origin from either `AUTH_URL` or `NEXTAUTH_URL` environment variables,
+ * or the request's headers option.
+ */
+export function getOrigin(): string {
+  const envUrl = process.env.AUTH_URL ?? process.env.NEXTAUTH_URL;
+  const h = headers();
+
+  let url: URL;
+  if (envUrl) {
+    url = new URL(envUrl);
+  } else {
+    const detectedHost = h.get("x-forwarded-host") ?? h.get("host");
+    const detectedProtocol = h.get("x-forwarded-proto") ?? "https";
+
+    url = new URL(`${detectedProtocol}://${detectedHost}`);
+  }
+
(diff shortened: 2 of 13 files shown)


middleware.ts
```diff
@@ -9,9 +8,16 @@ import CredentialsProvider from "next-auth/providers/credentials";
 import NextAuth, { Session } from "next-auth";
 import { JWT } from "next-auth/jwt";
 
-acceptLanguage.languages(languages);
 const localPathRegEx = new RegExp("^(?!((?:[a-z+]+:)?//))", "i");
 
+const verboseDebug = false;
+
+const debugLogger = async (message: string) => {
+  if (verboseDebug) {
```
timarney, line 16: Is this for local debugging to flip it to true or something?

bryan-robitaille: Yes, I found the middleware debug logs to be really noisy when not trying to debug what's going on with the routing but also really helpful if ever someone is having issues.  I thought about using a env var flag to trigger more verbose logging but was worried it might just be forgotten about.  So I went with a local variable in the file.  This way you can also switch it on the fly without having to restart your dev environment.

craigzour (approved):
Reviewed and tested! Nice work :)

lib/auth/nextAuth.ts
```diff
@@ -100,13 +100,15 @@ export const {
       },
     }),
   ],
-  secret: process.env.TOKEN_SECRET,
+  // When building the app use a random UUID as the token secret
+  secret: process.env.TOKEN_SECRET ?? crypto.randomUUID(),
```
thiessenp-cds, line 104: Would we need to point out that we used this crypto method (WebCrypto https://w3c.github.io/webcrypto/#Crypto-method-randomUUID) for our Protected B status?
I'm assuming this is fine but thought I'd ask just encase.

bryan-robitaille: No because the key is a throw away that is only used at build time.  It's only there to satisfy Next-Auths constructor.  Once the application 'starts' the real secret key that is passed in as an environment variable is used.

lib/origin.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+export function getOrigin(): string {
+  const envUrl = process.env.AUTH_URL ?? process.env.NEXTAUTH_URL;
+  const h = headers();
+
+  let url: URL;
+  if (envUrl) {
+    url = new URL(envUrl);
+  } else {
+    const detectedHost = h.get("x-forwarded-host") ?? h.get("host");
```
timarney, line 15: Do you think this could be exploited?

I'm looking into this a bit on my end.

bryan-robitaille: There is a potential for it to be exploited which is why it's only used on our PR Review and local environments.   In our live environments the NEXTAUTH_URL or AUTH_URL will always be available.

timarney (approved):

thiessenp-cds (approved):
Nice work! LGTM :)

PR #3475, opened 2024-04-12
feature: Prerender caching with Redis

bryan-robitaille:
# Summary | Résumé

Completely experimental work in progress

links in this section

# Pull Request Checklist


package.json
```diff
@@ -82,7 +83,7 @@
     "lodash.set": "4.3.2",
     "lodash.unset": "4.5.2",
     "markdown-to-jsx": "7.3.2",
-    "next": "^14.2.0-canary.38",
```
timarney, line 85: Can bump [14.2 now](https://github.com/vercel/next.js/releases/tag/v14.2.2)

nextCacheHandler.mjs
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable no-console */
+import { CacheHandler } from "@neshca/cache-handler";
+import createLruHandler from "@neshca/cache-handler/local-lru";
+
+// This file is currently in js module format because it is imported into next.config.mjs
+// Once next js config supports type script this file can be convereted to ts module format
```
timarney, line 6:
```suggestion
// Once Next.js config supports TypeScript this file can be converted to `ts` module format
```

package.json
bryan-robitaille: The version isn't pinned so it should grab the latest patch version on build.

timarney: Ya I think it should be pinned.

timarney (approved):
Look good -- as you mentioned it's hard to test ... 

Will approve and we'll check the logs as we go to see if this corrects the issue.

package.json
bryan-robitaille: Pinned

timarney (approved):

Dockerfile.pr
```diff
@@ -39,6 +39,9 @@ ENV TAG_VERSION=$TAG_VERSION
 ENV AWS_LWA_ENABLE_COMPRESSION=true
 ENV HOSTNAME=localhost
 ENV PORT=3000
+# Use the REVIEW_ENV environment variable to determine if the app is running in a review environment
+# Enables memory caching for the prerendering of pages
+ENV REVIEW_ENV=1
```
wmoussa-gc, line 44: so if it is `1` what does it mean exactly? if it represents a boolean truth,  later on I see the following check which is confusing: `process.env.NODE_ENV === "production" &&
    process.env.REVIEW_ENV`

nextCacheHandler.mjs
```diff
@@ -0,0 +1,21 @@
...
+import createLruHandler from "@neshca/cache-handler/local-lru";
+
+// This file is currently in js module format because it is imported into next.config.mjs
+// Once Next.js config supports TypeScript this file can be converted to `ts` module format
+
+CacheHandler.onCreation(async () => {
+  // Create an in-memory Least Recently Used cache for use in PR review environments
+  const localHandler = createLruHandler({
+    maxItemsNumber: 1000, // 1000 items
+    maxItemSizeBytes: 1024 * 1024 * 50, // 50 MB
```
wmoussa-gc, line 12: do I read this like a possible memory space of 50,000 MB (50GB)?

next.config.mjs
```diff
@@ -77,7 +85,7 @@ const nextConfig = {
 
   experimental: {
     instrumentationHook: true,
-    ppr: true,
+    // ppr: true,
```
wmoussa-gc, line 88: whats the plan with this feature (partial pre-rendering)? do we need to keep the commented config in the code?

bryan-robitaille: Good catch, I'll have to review why I commented it out.

nextCacheHandler.mjs
bryan-robitaille: Yes, you're right that is the worse case scenario.  We can shrink them if need be but those limits will not be attained by caching nextjs pages.  They were the defaults provided by the package author.

Dockerfile.pr
bryan-robitaille: I can simplify it for now by removing the NODE_ENV check until we also include the external caching in non-PR-Review environments.
The "1" simply assigns an arbitrary value that results in a Boolean 'true' to check if the variable is defined.  I could also assign the value to 'TRUE' if it helps with readability.

bryan-robitaille: NODE_ENV = production also when you build and run the app locally.

bryan-robitaille: Will create a new PR with the changes

next.config.mjs
bryan-robitaille: The PPR feature is only enabled in the latest canary versions and since we've switched back to their normal release branch the feature is no longer available.
Screenshot 2024-04-22 at 1 13 05 PM

PR #3485, opened 2024-04-18
If form isn't saved, still allow downloading

dsamojlenko:
# Summary | Résumé

Previously Form Management would throw a not found error if you navigated to it without having saved the form first.
This PR removes the notFound response and displays the DownloadForm component.


timarney:
Code looks great --- much easier to follow.

Let's see if we can get this out for some user QA with the different scenarios.

timarney (approved):
Approving --- test cases will be added to the main test cases so we should get this rolled into the main update branch to ensure test tests are performed against these updates.

PR #3486, opened 2024-04-18
feat: implement temporary hack to handle Submission lambda name change

craigzour:
# Summary | Résumé

linked to https://github.com/cds-snc/forms-terraform/pull/626

- Adds a hack to handle Submission lambda function name which is going to be changed in an upcoming infra update.

# IMPORTANT:

This should be deployed to Staging and Production "as soon as possible" so that the Lambda containerization update on the infra side can also be released.

### DIFF
--- a/components/form-builder/app/edit/elements/lexical-editor/tests/RichTextEditor.cy.tsx
+++ b/components/form-builder/app/edit/elements/lexical-editor/tests/RichTextEditor.cy.tsx
@@ -14,7 +14,7 @@ describe("<RichTextEditor />", () => {
     );
 
     // Add some strings to get formatted
-    cy.get(`[id^="editor-"]`).type(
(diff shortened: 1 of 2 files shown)


bryan-robitaille (dismissed):
Nice, looks good to me.

wmoussa-gc (dismissed):

craigzour:
@bryan-robitaille @wmoussa-gc had to implement a fix for a test that was not passing anymore (no idea why it suddenly happened). The issue is that the modal for creating a link in a RichTextBox appears stay up even though we press `escape` in the Cypress code ( `.type("example.com{enter}{esc}");` ). The result is that Cypress is not able to interact with HTML element underneath that modal. Decided to move the test at the end of all the others until we find a proper solution.

bryan-robitaille (approved):

PR #3495, opened 2024-04-22
chore: add script to convert TTL data type from String to Number in the ReliabilityQueue DynamoDB table

craigzour:
# Summary | Résumé

Follow up for https://github.com/cds-snc/forms-terraform/pull/637

- Adds script to convert TTL data type from String to Number in the ReliabilityQueue DynamoDB table.

# IMPORTANT

This script should be executed in both environments once https://github.com/cds-snc/forms-terraform/pull/637 has been deployed. We can already do it in Staging.


wmoussa-gc (approved):
lgtm

PR #3496, opened 2024-04-22
fix: Reduce mem cache size and simplify config logic

bryan-robitaille:
# Summary | Résumé

Introduces changes suggested by Wissam in PR #3475


wmoussa-gc (approved):
ty

PR #3503, opened 2024-04-23
chore: Remove v1 of Response Retrieval API

bryan-robitaille:
# Summary | Résumé

Removes the previous experimental API for response retrieval.  This feature will be completely rewritten and implemented.

A prisma migration will need to be run once this code is in Staging to remove the 'bearerToken' field from the Templates table.


craigzour (dismissed):
Good riddance :)

craigzour (approved):

PR #3511, opened 2024-04-24
chore: Cypress E2E test refactor

bryan-robitaille:
# Summary | Résumé

This refactor aims at enabling our previous existing e2e tests with the new App Router authentication and available APIs.

This proof of concept introduces several changes:
 - Store and leverage the JWT of an authenticated user across multiple tests
 - Fixes typing into unhandled form input elements and WYSIWYG components
 - Fixes Cypress runner beginning test before page has been hydrated client side.

Biggest change for performance was removing all the interceptors and instead adding a global component with 0 height in the layout that will only render client side.  This way we can check that the div exists and then begin out cypress test.

Why?  The SSR was rendered and presented client side with the form inputs existing in the HTML. However the various event listeners were not yet wired up through hydration.  This was causing our test runner to try and interact with inputs and buttons that were not yet ready for interactivity.

Bugs found and fixed during test activation:
- Deactivated user is now routed to the deactivated users page
- Added missing required validation messages on Registration Form for password field.


app/(gcforms)/[locale]/(user authentication)/auth/register/action.ts
```diff
@@ -58,6 +58,7 @@ const validate = async (
         ),
       ]),
       password: v.string([
+        v.minLength(1, t("input-validation.required", { ns: "common" })),
```
Comment (wmoussa-gc) [inline app/(gcforms)/[locale]/(user authentication)/auth/register/action.ts:61]: Do we still need this one, even though we have another minimum length check below?

cypress/e2e/authenticatedFlows/acceptable_use.ts
```diff
@@ -0,0 +1,50 @@
+const AcceptableUsePage = () => {
+  describe("Test acceptable use Page", () => {
+    beforeEach(() => {
+      cy.userSession({ admin: false, acceptableUse: false });
+      cy.visitPage("/en/auth/policy");
+    });
+
+    // The calls to api/session are causing the Accept button to become in a detached state when trying to click on it.
```
wmoussa-gc, line 8: i think this comment is not relevant anymore?

timarney:
Didn't test locally as I know this is still a draft just pulling out some logs from the test run --- 

No action needed just making a note.

```
warn(prisma-client) This is the 10th instance of Prisma Client being started. Make sure this is intentional.
{"level":"info","time":1713988585759,"msg":"Seeding database"}


{"level":"error","time":1713988679984,"err":{"type":"s","message":"Could not load credentials from any providers","stack":"CredentialsProviderError: Could not load credentials from any providers\n    at 

```

app/(gcforms)/[locale]/(user authentication)/auth/register/action.ts
Comment (bryan-robitaille) [reply app/(gcforms)/[locale]/(user authentication)/auth/register/action.ts:61]: Yes, they drive different validation error messages.  The min 1 drives the 'field required' and the min 8 drives 'password must have at least 8 characters'

cypress/e2e/authenticatedFlows/acceptable_use.ts
bryan-robitaille: yup, no longer relevant

components/clientComponents/globals/HydrationCheck.tsx
```diff
@@ -0,0 +1,12 @@
+"use client";
+import { useEffect, useState } from "react";
+
+export const HydrationCheck = () => {
+  const [hydrationComplete, setHydrationComplete] = useState(false);
```
timarney, line 5: Wondering if we might want to rename this so it doesn't get mixed up with the template store hydration.

bryan-robitaille: Good idea.

cypress/e2e/form-builder/autocomplete.cy.ts
```diff
@@ -1,69 +0,0 @@
-describe("Test FormBuilder autocomplete props", () => {
```
timarney, line 1: What was the issue here - I know these tests have got some issues in the past.

Maybe move to component test?

bryan-robitaille: The test specs that were deleted had their test moved to other similar specs.  For example the tests in this autocomplete spec were moved to the same spec file as 'add an element'.
The reason for the colocating similar tests is that the cypress runner closes and reopens the chrome process between each test spec.  So by colocating similar tests we save a bit of run time.

timarney (approved):
Great work -- getting these running again.

🥇

PR #3516, opened 2024-04-25
Fixes Card Title Localization

thiessenp-cds:
# Summary | Résumé

Fixes card title localization string by adding it to the my-forms locale file.  The localization key was copy+paste from the form-builder file. The key is used in form-builder also so there is a chance that the key might be updated and then a bug created in my-forms.

Also, the locale from the Page params is used to be more consistent (vs. using 18n locale one place and page params locale another place)

**Before**
Screenshot 2024-04-25 at 9 18 38 AM

**After**
Screenshot 2024-04-25 at 9 33 08 AM

## Test

Go to the Forms (my-forms) page. Select French for the language. Any card without a name should have the French string printed out.


wmoussa-gc:
and the english label is also fixed, thank you
image

wmoussa-gc (approved):

Issue #3352, opened 2024-03-18
"Manage Ownership" search box


Abi-Nada:
## Description
The search box for "Manage Ownership" doesn't appear.

## Steps to reproduce
1. Open up a form (draft)
2. Go to Settings
3. Go to the "Form Management" tab
4. Scroll down to the "Manage Ownership" section and see error.

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: PR
- **Language**: English

## Expected behaviour
Ideally we (not other clients, to date) should be able to search folks to add to the form.

<img width="1652" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMTNzQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--412d70fcc3f8b0998277ce068e33190ba8134c6a/Screenshot%202024-03-18%20at%201.24.15%E2%80%AFPM.png" alt="Screenshot 2024-03-18 at 1.24.15 PM.png" />

## Screenshots or videos

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMkhzQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--68d916a42d17fa4e3478493a6753c5838049e920/Screen%20Recording%202024-03-18%20at%201.22.51%E2%80%AFPM.mov)

## Additional context
Any other information about the problem here.

timarney:
> 6958-0492abac00a8e2d0.js:1 Refused to apply inline style because it violates the following Content Security Policy directive: 

Believe this issue is limited to the preview server --- 

I would suggest we take a look once the PR is on the staging server.

timarney:
@Abi-Nada going to assign this to you --- **we'll need to re-check once this is on staging** -- the styles get blocked on the preview url.

Maybe we add a check on staging label or something?

timarney:
I took a further look at this --- looks like the CORS scripts we're updated to block how the styles are loaded for that component (in the app router branch --- better security) -- which means it's different vs current staging works. 

Will look into finding a solution for this.

timarney:
This is fixed --- mostly.

@Abi-Nada code has been merged that uses the `nonce` value to ensure the styles are as safe as can be for inline styles.


[...]

PR #3521, opened 2024-04-26
React select styling - unsafe-inline issue

timarney:
# Summary | Résumé

Fixes: https://github.com/cds-snc/platform-forms-client/issues/3352

Current on preview environment. 

screenshot_2024-04-25_at_2 25 08___pm

<hr>

## Notes

With our CSP script blocking `unsafe-inline` the react select library styles are getting blocked.

There is an option to set an `unstyled` property however some base styles still try to load and are currently needed.

> If you are trying to style everything from scratch you can use the unstyled prop. This removes all the presentational styles from React Select (leaving some important functional styles, like those for menu positioning and input width in multi select).

https://react-select.com/styles#the-unstyled-prop

There is a closed issue on the repo discussing the issue here:

https://github.com/JedWatson/react-select/issues/2706

## Fix 

Screenshot 2024-04-26 at 11 47 46 AM

This PR adds `unstyled` prop & uses Tailwind for most of the styling however as noted above inline styles are still required.

👉  To fix the inline styles issue the CSP has been updated for now to allow inline styles when a cookie session exists


timarney:
Might be possible to use 

https://emotion.sh/docs/@emotion/cache#nonce-string

timarney:
Updated to use Emotion Cache key to provide a nonce and avoid unsafe inline


https://github.com/cds-snc/platform-forms-client/pull/3521/files#diff-6d7e944d154e9748f30bffc299f0627af3c9d572a96b837b438d0701b6ac3515R46

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/FormOwnership.tsx
```diff
@@ -1,28 +1,25 @@
...
 import { logMessage } from "@lib/logger";
 import { Button } from "@clientComponents/globals";
-import Select from "react-select";
-import makeAnimated from "react-select/animated";
+import { FormOwnerSelect, usersToOptions } from "./FormOwnerSelect";
+
+import { CacheProvider, EmotionCache } from "@emotion/react";
+import createCache from "@emotion/cache";
+
+// import makeAnimated from "react-select/animated";
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/FormOwnership.tsx:12]: Assuming commenting out the import is left over debugging?

```suggestion
```

lib/cspScripts.ts
```diff
@@ -45,7 +46,7 @@ export const generateCSP = (): { csp: string; nonce: string } => {
   const cspHeader = `
     default-src 'self';
     script-src 'self' 'nonce-${nonce}' 'strict-dynamic';
-    style-src 'self' 'nonce-${nonce}';
+    style-src 'self' ${styleSrc};
```
bryan-robitaille, line 49: Curious why the change in value to a new declared variable?
```suggestion
    style-src 'self' 'nonce-${nonce}'
```

lib/cspScripts.ts
```diff
@@ -24,8 +24,9 @@ if (window.location.host === "forms-formulaires.alpha.canada.ca") {
 }
 `;
 
-export const generateCSP = (): { csp: string; nonce: string } => {
+export const generateCSP = () => {
   const nonce = Buffer.from(crypto.randomUUID()).toString("base64");
+  const styleSrc = `'nonce-${nonce}'`;
```
bryan-robitaille, line 29: Might be better to leave inline?
```suggestion
```

lib/cspScripts.ts
```diff
@@ -24,8 +24,9 @@ if (window.location.host === "forms-formulaires.alpha.canada.ca") {
 }
 `;
 
-export const generateCSP = (): { csp: string; nonce: string } => {
```
bryan-robitaille, line 27: Did we run into an issue further in the PR with the typing?

bryan-robitaille (changes requested):
A little left over cleanup after doing amazing work to get the Emotion lib to use nonce's.

bryan-robitaille (dismissed):
Nice 👍

dsamojlenko (approved):

Issue #3530, opened 2024-05-01
Form Collaborator does not get an email notification


Abi-Nada:
## Description
When you add a new collaborator to a published form, they do not receive an email notification alerting them of the new access.

## Steps to reproduce 
1. (You'll need Manage All Forms permissions + another person to validate, or two accounts)
2. Go to Settings -> Form Management
3. Add a new email under "Manage Ownership"
4. Error: the new collaborator does not receive an email notification

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: PR
- **Language**: English

## Expected behaviour
Added user should get an email notification, like what is currently sent out in prod:
<img width="1591" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBenFwQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a1a8c99d30867b6e034f58942777e08f7d69be9e/Screenshot%202024-05-01%20at%208.38.11%E2%80%AFAM.png" alt="Screenshot 2024-05-01 at 8.38.11 AM.png" />

timarney:
Tracked down some error logs 

In a couple of cases we're seeing 

{"status": "failed", "message": "Notify failed to add ownership email", "error":Request failed with status code 403}

![screenshot_2024-05-01_at_3 40 50___pm_720](https://github.com/cds-snc/platform-forms-client/assets/62242/ab46d831-f1f0-46bc-93ac-fb99ac666bc2)

![screenshot_2024-05-01_at_3 47 08___pm_720](https://github.com/cds-snc/platform-forms-client/assets/62242/b9a75013-5f49-44d9-bf91-09337fa8c878)

Looking at the Notify UK docs it's possible this is related to the system clock.

timarney:
@Abi-Nada 

@bryan-robitaille 's PR is merged if you want to retest on the app router preview URL

Abi-Nada:
Re-tested 20 times in Staging across numerous forms and accounts, issue no longer persists.

PR #3536, opened 2024-05-02
fix: Notify Error Handling Centralization

bryan-robitaille:
# Summary | Résumé

Centralizes the `sendEmail` function calls and error handling into the `@lib/integrations`.
- The error handling includes extracting the exact detailed error message from Notify to include it in the logs.
- Adds a check to ensure the env var `TEMPLATE_ID` is set before calling the `sendEmail` function.

Should provide more context for issue: #3530

### DIFF
--- a/lib/integration/notifyConnector.ts
+++ b/lib/integration/notifyConnector.ts
@@ -53,7 +53,7 @@ export const getNotifyInstance = () => {
 export const sendEmail = async (email: string, personalisation: Record<string, string>) => {
   try {
     const templateId = process.env.TEMPLATE_ID;
-    if (!templateId) {
+    if (!templateId && process.env.APP_ENV !== "test") {
       throw new Error("No Notify template ID configured.");
     }
 
--- a/lib/integration/notifyConnector.ts
+++ b/lib/integration/notifyConnector.ts
@@ -78,7 +78,7 @@ export const sendEmail = async (email: string, personalisation: Record<string, s
          * is an instance of XMLHttpRequest in the browser and an instance
          * of http.ClientRequest in Node.js
          */
-        errorMessage = `${error.request}.`;
+        errorMessage = `Error sending to Notify with request :${error.request}.`;
       }
     } else if (error instanceof Error) {
       errorMessage = `${(error as Error).message}.`;
--- a/lib/integration/notifyConnector.ts
+++ b/lib/integration/notifyConnector.ts
@@ -1,5 +1,6 @@
 import { NotifyClient } from "notifications-node-client";
 import { logMessage } from "@lib/logger";
+import { AxiosError } from "axios";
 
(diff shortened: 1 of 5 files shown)


lib/integration/notifyConnector.ts
```diff
@@ -48,3 +49,50 @@ export const getNotifyInstance = () => {
 
   return notifyInstance;
 };
+
+export const sendEmail = async (email: string, personalisation: Record<string, string>) => {
+  try {
+    const templateId = process.env.TEMPLATE_ID;
```
timarney, line 55: So we allow the options to use a different template ID and default to this one?

Fine with leaving as is just asking questions

lib/integration/notifyConnector.ts
```diff
@@ -48,3 +49,50 @@ export const getNotifyInstance = () => {
...
+  } catch (error) {
+    let errorMessage = "";
+
+    if (error instanceof AxiosError) {
+      if (error.response) {
+        /*
+         * The request was made and the server responded with a
+         * status code that falls out of the range of 2xx
+         */
+        const notifyErrors = Array.isArray(error.response.data.errors)
```
wmoussa-gc, line 71: do we need to check for presence of  data, errors and status?

lib/integration/notifyConnector.ts
```diff
@@ -48,3 +49,50 @@ export const getNotifyInstance = () => {
...
+         * is an instance of XMLHttpRequest in the browser and an instance
+         * of http.ClientRequest in Node.js
+         */
+        errorMessage = `${error.request}.`;
+      }
+    } else if (error instanceof Error) {
+      errorMessage = `${(error as Error).message}.`;
+    }
+
+    logMessage.error(
```
wmoussa-gc, line 87: added the future alarm to https://docs.google.com/spreadsheets/d/1FXChe3eJq6BJgMBbF8Y_SXCnoOXoO3IiXMRWt7GoW0w/edit#gid=0

lib/integration/notifyConnector.ts
wmoussa-gc: I don't think we should include this option unless it's needed at the moment.

lib/integration/notifyConnector.ts
bryan-robitaille: According to the AxiosError type:
```js
export interface AxiosResponse<T = any, D = any> {
  data: T;
  status: number;
  statusText: string;
  headers: RawAxiosResponseHeaders | AxiosResponseHeaders;
  config: InternalAxiosRequestConfig<D>;
  request?: any;
}
```
`status` and `data` will aways be defined on `response`.
If `errors` is not defined than that `undefined` value will filter through and show up as `"... returned the following detailed errors ."`

wmoussa-gc (approved):

PR #3538, opened 2024-05-02
chore: update github action plugins to latest

bryan-robitaille:
# Summary | Résumé

Updates github actions to run latest versions as Node 16 is now depreciated.


craigzour (approved):
LGTM :)

PR #3547, opened 2024-05-03
fix: Replace Notify Client with custom instance

bryan-robitaille:
# Summary | Résumé

Removes `notifications-node-client` and implements a custom instance that creates a custom axios config.

Testing now shows that the detailed error is included in the logs.

Screenshot 2024-05-03 at 1 47 58 PM

### DIFF
--- a/app/api/share/route.ts
+++ b/app/api/share/route.ts
@@ -1,6 +1,6 @@
 import { NextResponse } from "next/server";
 import { middleware, sessionExists } from "@lib/middleware";
-import { getNotifyInstance } from "@lib/integration/notifyConnector";
+import { sendEmail } from "@lib/integration/notifyConnector";
 import { logMessage } from "@lib/logger";
 import { MiddlewareProps, WithRequired } from "@lib/types";
 import { slugify } from "@lib/client/clientHelpers";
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -86,7 +86,6 @@
     "next": "14.2.3",
     "next-auth": "5.0.0-beta.16",
     "node-fetch": "^3.3.2",
-    "notifications-node-client": "7.0.6",
```
timarney, line 89: :)

app/api/share/route.ts
```diff
@@ -19,21 +19,17 @@ export const POST = middleware([sessionExists()], async (req, props) => {
...
-            },
-            subject: "Form shared | Formulaire partagé",
-            formResponse: `
+        return sendEmail(email, {
+          application_file: {
+            file: base64data,
+            filename: `${slugify(filename)}.json`,
+            sending_method: "attach",
+          },
+          subject: "Form shared | Formulaire partagé",
```
timarney, line 31: Can't remember does the "real subject" come from the template itself?

Seeing the current email subject as `GC Forms | Formulaires GC`

bryan-robitaille: Not for the Template used throughout the app.  We have a personalization holder setup in the Subject of the Template.

craigzour (dismissed):
LGTM! Nice work :)

timarney (dismissed):

craigzour (approved):

PR #3565, opened 2024-05-07
fix: create only one instance of DynamoDBDocumentClient

craigzour:
# Summary | Résumé

- Trying to fix the `CredentialsProviderError: Could not load credentials from any providers` error by initializing the `DynamoDBDocumentClient` only once.

### DIFF
--- a/lib/integration/dynamodbConnector.ts
+++ b/lib/integration/dynamodbConnector.ts
@@ -1,15 +1,9 @@
 import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
 import { DynamoDBDocumentClient } from "@aws-sdk/lib-dynamodb";
 
-/**
- * Helper function to instantiate DynamoDB and Document client.
- * https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/dynamodb-example-document-client.html
- */
-export function connectToDynamo(): DynamoDBDocumentClient {
-  return DynamoDBDocumentClient.from(
-    new DynamoDBClient({
-      region: process.env.AWS_REGION ?? "ca-central-1",
-      ...(process.env.LOCAL_AWS_ENDPOINT && { endpoint: process.env.LOCAL_AWS_ENDPOINT }),
-    })
-  );
-}
+export const dynamodbClient = DynamoDBDocumentClient.from(
+  new DynamoDBClient({
+    region: process.env.AWS_REGION ?? "ca-central-1",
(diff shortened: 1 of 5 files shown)


lib/integration/dynamodbConnector.ts
```diff
@@ -1,15 +1,21 @@
...
  * https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/dynamodb-example-document-client.html
  */
 export function connectToDynamo(): DynamoDBDocumentClient {
-  return DynamoDBDocumentClient.from(
-    new DynamoDBClient({
-      region: process.env.AWS_REGION ?? "ca-central-1",
-      ...(process.env.LOCAL_AWS_ENDPOINT && { endpoint: process.env.LOCAL_AWS_ENDPOINT }),
-    })
-  );
+  if (!dynamodbInstance) {
```
wmoussa-gc, line 11: It reminds me of the era of the Java singleton pattern, though I have no idea how this applies to Node

craigzour: Pretty much experimenting here. If that fixes the error then there are two things that I can think of:
- We did not experience it in the PR environment because it runs in a Lambda function so the connection between AWS services may be handled differently than with our EC2 instance (Staging).
- Since we pretty much had the same implementation before merging the App Router refactoring we could imagine that the code that loads the forms is executed way more times than with the Pages Router.

wmoussa-gc: could we adopt this v3 way? https://github.com/awsdocs/aws-doc-sdk-examples/blob/main/javascriptv3/example_code/cross-services/ddb-item-tracker/public/libs/dynamoClient.js

lib/integration/dynamodbConnector.ts
```diff
@@ -1,15 +1,21 @@
 import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
 import { DynamoDBDocumentClient } from "@aws-sdk/lib-dynamodb";
 
+let dynamodbInstance: DynamoDBDocumentClient | undefined = undefined;
+
 /**
  * Helper function to instantiate DynamoDB and Document client.
  * https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/dynamodb-example-document-client.html
```
wmoussa-gc, line 8: since we are now using v3, we could do the new way

craigzour: Sure I can try now. I was thinking about testing this fix and then refactor the code so that all AWS services are going through such type of singleton.

Let me rework this one first and then we can test before modifying all the others.

wmoussa-gc (approved):

PR #3571, opened 2024-05-07
chore: refactor AWS SDK initialization to make sure there is always one instance opened

craigzour:
# Summary | Résumé

- After experiencing an issue with the DynamoDB SDK being initialized multiple times in a row we decided to rework all SDK initialization inside the web application.


wmoussa-gc (approved):
lgtm

Issue #3574, opened 2024-05-08
User not automatically logged out after 2 hours of inactivity


Abi-Nada:
## Description
I logged into Staging the previous night and left the tab open - 12 hours later, I'm still logged into Staging/didn't need to input a security code.

## Steps to reproduce
1. Log into Staging
2. Keep the tab open for over 2 hours
3. Navigate around the application

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Staging
- **Language**: English

## Expected behaviour
It should have logged me out after 2 hours of non-activity.

## Screenshots or videos
Date of last security code sent was May 7 at 8:06 PM. On May 8 around 8:20AM, I'm still able to move around the application without getting logged out.

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOFBEQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b0386d81e11beb9d35857f69f75683ad2d3fa019/Screen%20Recording%202024-05-08%20at%208.27.31%E2%80%AFAM.mov)

bryan-robitaille:
Confirmed that this is not an issue in Production and is only showing up in Staging.

PR #3578, opened 2024-05-08
Fix: Session expiry set to 2 hours

bryan-robitaille:
# Summary | Résumé

fixes: #3574 

The NextAuth middleware was using the default value of 30 days for session length even though the main NextAuth configuration in the app was set to 2 hours.  This was creating a condition where the middleware would overwrite the JWT token expiry and the effectively remove the 2 hour limitaiton.

Fix:
- Explicitly set the session length for the JWT in the middleware configuration

- Verify next auth session cookie in Staging will have expiry set for 30 days after logging in.
- Ensure PR Review environment has same cookie set for 2 hours after logging in.

- Dig deeper into the code base of Auth JS to see if there is a way to reduce the possible duplication of setting cookies between the middleware running on Edge and the Node application.

### DIFF
--- a/lib/auth/nextAuth.ts
+++ b/lib/auth/nextAuth.ts
@@ -106,6 +106,7 @@ export const {
     strategy: "jwt",
     // Seconds - How long until an idle session expires and is no longer valid.
     maxAge: 2 * 60 * 60, // 2 hours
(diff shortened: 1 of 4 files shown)


timarney (approved):

wmoussa-gc:
good catch

PR #3584, opened 2024-05-10
Add util to sync layout array with group elements

timarney:
# Summary | Résumé

Adds utilities to sync group layout with group element order

Note: this is hardcoded to not be used yet.


thiessenp-cds (approved):
LGTM :)

PR #3635, opened 2024-05-14
Feature/fix adds groups history to the form builder

thiessenp-cds:
# Summary | Résumé

Adds a history array and util to keep track of the "path" a user navigates through a form. This is useful for showing what groups and related answers a user filled out. It may also be helpful for things like analytics in the future.


lib/hooks/useGCFormContext.tsx
```diff
@@ -56,6 +65,7 @@ export const GCFormsProvider = ({
 
       if (typeof nextAction === "string") {
         setCurrentGroup(nextAction);
+        addHistory(nextAction);
```
thiessenp-cds, line 68: Instead of coupling here, this could be moved to the Next button but then some of this logic would probably need to be duplicated.

timarney (approved):

PR #3648, opened 2024-05-15
Fix "forms" icon fill when active

timarney:
# Summary | Résumé

https://github.com/cds-snc/platform-forms-client/assets/62242/8cd390a1-3834-4b63-8600-3354ff85695c


thiessenp-cds (approved):
LGTM :)

PR #3650, opened 2024-05-15
Fix/sign out time

thiessenp-cds:
# Summary | Résumé

Updates the sign out page to use the browser's local timezone to set the current time. 
Previously this was printing UTC time because that is what the AWS node instances are set to.

## Test

Test by going to the sing out page and checking the time is right. Also for bonus points open the browser developer tools and make sure there are no errors.
Screenshot 2024-05-15 at 1 41 20 PM


bryan-robitaille (approved):
Nice, works as intended 👍

Issue #3623, opened 2024-05-14
Blank page during sign-in


anikbrazeau:
## Description
Blank screen after hitting cancel or back on Accepting the terms.

## Steps to reproduce
1. Go to GC Forms
2. Sign in
3. Don't accept terms
4. Hit back

## Details
- **Browser**: Edge
- **Operating system**: Mac
- **Environment**: Staging
- **Language**: French
- **Viewport**: Desktop

## Expected behaviour
Back to sign in maybe?

## Screenshots or videos
<img width="888" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd3JiQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8b6a4c7cd57a8e0559ee499aec4df0d80574fcbc/Screenshot%202024-05-14%20at%207.51.06%E2%80%AFAM.png" alt="Screenshot 2024-05-14 at 7.51.06 AM.png" />

srtalbot:
Retested on Production and could not replicate - likely an app router issue.

srtalbot:
Confirmed this is the app router

timarney:
Does this happen outside of edge?


- Sign in 
- Enter security code
- Land on acceptable use page
- Press browser back button

I'm seeing the session expired screen 

https://github.com/cds-snc/platform-forms-client/assets/62242/a8c50ed6-cabf-4aad-a562-9a6c670d5d3c

anikbrazeau:
I only saw it on Edge, haven't been able to replicate in Chrome. 

Though on the "Know your responsibilities" page you can click on "Forms" in the top nav or on the GC Forms logo and get a blank page also, in both Edge and Chrome.

PR #3662, opened 2024-05-16
Remove forms link if acceptable use hasn't been pressed 

timarney:
# Summary | Résumé

Removes the link to forms from the acceptable use and other pages.  Forms is an auth'd link that requires the user to accept the terms so it shouldn't be available as a link.

> Though on the "Know your responsibilities" page you can click on "Forms" in the top nav or on the GC Forms logo and get a blank page also, in both Edge and Chrome.

https://github.com/cds-snc/platform-forms-client/issues/3623#issuecomment-2115636549

Noting this just hides the link -- it's still possible to reach the blank page as it falls through the middleware function.


thiessenp-cds (approved):
Nice. LGTM :)

When I was testing the acceptable use page. I clicked the accept button and while the page was loading, I could see the Forms link flash in long enough for me to click it. When I clicked it I was taken to the Forms page. 
I'm guessing at this point the session was updated so clicking the Forms link works as expected.

timarney:
Right --- we'll consider this a partial fix and go from there.

PR #3666, opened 2024-05-17
fix: auth error handling noise

### LINKED ISSUE #3657: Next Auth creating unnecessary errors
The new version of NextAuth throws an error for any bad login attempt. This affects the MFA section

### Acceptance criteria:
- [ ] Review NextAuth errors
- [ ] Decide whether we want to pass through the errors to the log message function.

wmoussa-gc:
# Summary | Résumé

Fixes #3657 (The new version of NextAuth throws an error for any bad login attempt. This affects the MFA section)

This PR fine-tunes the AuthJs logger customization.  
It also fixes a flaky test.

### DIFF
--- a/lib/auth/nextAuth.ts
+++ b/lib/auth/nextAuth.ts
@@ -113,19 +113,8 @@ export const {
   debug: process.env.NODE_ENV !== "production",
   logger: {
     error(code, ...message) {
-      logMessage.error(
-        `NextAuth error - Code: ${code}. Error: ${JSON.stringify(message, (_, value) => {
-          if (value?.constructor.name === "Error") {
-            return {
-              name: value.name,
-              message: value.message,
-              stack: value.stack,
-              cause: value.cause,
-            };
-          }
-          return value;
-        })}`
-      );
+      // Log the auth error as warn to avoid triggering a Slack alert and leave it to the front end to handle
+      logMessage.warn(`NextAuth error - Code: ${code} - Message: ${message}`);
     },
     warn(code) {
       logMessage.warn(`NextAuth warning - Code: ${code}`);
--- a/lib/auth/nextAuth.ts
+++ b/lib/auth/nextAuth.ts
@@ -112,10 +112,7 @@ export const {
   trustHost: process.env.AUTH_URL ? false : true,
   debug: process.env.NODE_ENV !== "production",
   logger: {
-    error(code, ...message) {
-      // Log the auth error as warn to avoid triggering a Slack alert and leave it to the front end to handle
-      logMessage.warn(`NextAuth error - Code: ${code} - Message: ${message}`);
-    },
+    error() {},
     warn(code) {
       logMessage.warn(`NextAuth warning - Code: ${code}`);
     },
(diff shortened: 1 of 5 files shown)


lib/nagware.ts
```diff
@@ -30,7 +30,7 @@ export async function detectOldUnprocessedSubmissions(
           const diffMs = Math.abs(currentDate - current.createdAt);
 
           // 86400000 milliseconds = 1 Day
-          const diffDays = Math.ceil(diffMs / 86400000);
+          const diffDays = Math.floor(diffMs / 86400000);
```
wmoussa-gc, line 33: This is to solve for a flaky test: nagware.test.ts.

A quick test case simulation shows that using Date.now in the test and Date.now later in the app code creates a tiny delta that is amplified by Math.ceil.

image

lib/auth/nextAuth.ts
```diff
@@ -113,19 +113,8 @@ export const {
   debug: process.env.NODE_ENV !== "production",
   logger: {
     error(code, ...message) {
-      logMessage.error(
```
wmoussa-gc, line 116: This log statement has been more noisy than useful

bryan-robitaille:
Completely removing the logger configuration for `error` does have some consequences that we'll need to handle before merging.  Removing the configuration block results in any Errors that are thrown in Next-Auth, including in our custom code in Callbacks and Events, to not be logged.  If there are any issues with our backend services while creating the JWT or Session and logging the events of sign in or sign out, the errors will not be logged and we would not be advised of any possible issues.

wmoussa-gc:
> Removing the configuration block results in any Errors that are thrown in Next-Auth, including in our custom code in Callbacks and Events, to not be logged

But this comment assumes we are not catching auth errors or any unhandled errors anywhere in the application. Do you mean a different format of error logging that we might be missing?

bryan-robitaille:
> Keep in mind, AuthJs should not be swallowing the errors thrown by the framework. 

In testing I threw errors in the callbacks and events and they did not bubble up.  This is because the `error() {}` swallows the errors because the property is defined.  If the `error()` is not in the configuration I assume the credential Errors will also be thrown as an unhandled error?

lib/auth/nextAuth.ts
```diff
@@ -175,11 +174,14 @@ export const {
       // account is only available on the first call to the JWT function
       if (account?.provider) {
         if (!token.email) {
-          throw new Error(`JWT token does not have an email for user with name ${token.name}`);
+          logMessage.error(`JWT token does not have an email for user with name ${token.name}`);
```
wmoussa-gc, line 177: Logs example when in production:

```
{"level":"error","time":1716406719058,"msg":"JWT token does not have an email for user with name Wissam Moussa"}
{"level":"error","time":1716406719058,"msg":"NextAuth error: {\"name\":\"s\",\"type\":\"CallbackRouteError\",\"kind\":\"error\",\"message\":\"Read more at https://errors.authjs.dev#callbackrouteerror\"}."}

```

bryan-robitaille (approved):
LGTM

PR #3669, opened 2024-05-21
Fix: WIP JWT session optimization

bryan-robitaille:
# Summary | Résumé

This PR looks complex because it touches many files but in reality it's just replacing existing functions with new versions.

fixes: #3623 

## New Functions

Replaces the `auth()` function server side to instead use a cached version that returns the same session for a single server side request.  This reduces the complexity of the session being created on every function call and instead the JWT and Session are only generated once per server request.  The function comes with 2 different versions. 

*`authCheckAndThrow()`*
If there is no session the function throws an error which needs to be caught and handled from the calling function.

*`authCheckAndRedirect()`*
If there is no session the function does not throw and instead redirects to the login page

Screenshot 2024-05-27 at 11 36 54 AM

## Some light weight changes

1.  Added `client-only` package that can be imported into files to throw errors if ever the module is mistakenly imported into a server component
2. Added `server-only` package that can be imported into files to throw errors if ever the module is mistakenly imported into a client component.
3. Updated `next-auth` to latest beta version.  This includes a fix to the Session Provider where it was making 3x the calls necessary to the `/api/auth/session` endpoint.
4. Ensure that some Next Links do not prefetch.  For example header links that includes links to Forms or language switcher.
5. Adds a Content Security Policy to API calls to `default: none;` which protects in case the API is opened in a browser
6. Updated the page Content Security Policy to add `frame-ancestors: none;` as an added protection against i-frames
7. Updated `pino` and `pino-prettier` now that they are compatible with Next app router and removed workaround from package.json launch scripts
8. Remove the image property from the session as we do not have profile images for users.
9. Middleware Regex now filters out all files and only treats pages, prerenders, api, and action requests.
10. Removes the referer from the `/auth/policy` page and middleware.  It was trying to be cute but did not add any value to the user experience.

## Workarounds that require more work but are out of scope of current PR.
The Session Provider provided by next-auth is not longer built to be used with custom pages and app router patterns.  In order to ensure the Session Provider is updated accordingly with the auth actions in our app I had to create a function that broadcasts to the channel that next-auth is listening on in order to trigger a Session Provider update.  While this works for the current version there is no guarantee that it will not break on package updates.

Future work would be to remove the Session Provider completely and instead adopt a pattern where the Server Components fetch the session server side and pass it down to the client components that require session user data.

Start the app and navigate around and log in / log out.
- The console should be a lot less busy with JWT / Session updates
- No functionality should be broken.

# Pull Request Checklist

- [ ]

### DIFF
--- a/app/(gcforms)/[locale]/(form administration)/forms/layout.tsx
+++ b/app/(gcforms)/[locale]/(form administration)/forms/layout.tsx
@@ -2,7 +2,6 @@ import React from "react";
 import { ToastContainer } from "@formBuilder/components/shared/Toast";
 import { Header } from "@clientComponents/globals/Header/Header";
 import { authCheckAndRedirect } from "@lib/actions";
-import { redirect } from "next/navigation";
 import { SaveTemplateProvider } from "@lib/hooks/form-builder/useTemplateContext";
 import { TemplateStoreProvider } from "@lib/store/useTemplateStore";
 import { SkipLink } from "@serverComponents/globals/SkipLink";
@@ -15,9 +14,7 @@ export default async function Layout({
   children: React.ReactNode;
   params: { locale: string };
 }) {
-  await authCheckAndRedirect().catch(() => {
-    redirect(`/${locale}/auth/login`);
-  });
+  await authCheckAndRedirect();
 
   return (
     <TemplateStoreProvider {...{ locale }}>


timarney (approved):
Approved nice work

The one thing I debated while reviewing this is the `repetitive` catch redirect 

Wondering if we should think about a helper util for that 

My minor concern is typos for the path ... maybe missing the `/` 

and  also we don't do any checks on `language` and or `locale` being one of `en` or `fr`  --- which would end up with a 404 anyway but just thinking.

```javascript
 const { ability } = await authCheck().catch(() => {
    redirect(`/${language}/auth/login`);
  });
```

thiessenp-cds (approved):
Nice work! The code looks very clean and I tested the happy paths in Chrome and Edge. LGTM :)

PR #3683, opened 2024-05-22
Updates branding with new error pattern

thiessenp-cds:
# Summary | Résumé

Updates branding feature to use new error pattern. Also adds error code.

Screenshot 2024-05-22 at 2 25 23 PM

### DIFF
--- a/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/branding/components/Branding.tsx
+++ b/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/branding/components/Branding.tsx
@@ -11,7 +11,7 @@ import { Button } from "@clientComponents/globals";
 import Brand from "@clientComponents/globals/Brand";
 import { ExternalLinkIcon } from "@serverComponents/icons";
 import { updateTemplate } from "@formBuilder/actions";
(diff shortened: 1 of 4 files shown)


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/branding/components/Branding.tsx
```diff
@@ -54,20 +54,24 @@ export const Branding = ({ hasBrandingRequestForm }: { hasBrandingRequestForm: b
...
-        id,
-        formConfig: JSON.parse(getSchema()),
-        name: getName(),
-      });
-    } catch (e) {
-      logMessage.error(e);
-      toast.error(savedErrorMessage);
+    const result = await updateTemplate({
+      id,
+      formConfig: JSON.parse(getSchema()),
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/branding/components/Branding.tsx:59]: Thinking about this now. I wonder if the try-catch originally here should stay because of JSON.parse

Comment (timarney) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/branding/components/Branding.tsx:59]: ya we need the catch --- tested

timarney (approved):

Issue #3636, opened 2024-05-14
Sharing a form by email to an address containing a plus sign gets silently discarded


jprince-cds:
## Description
When the user wants to share their form via email, if they type an email address with a plus sign (+) as is allowed by Gmail and some other email providers, the operation appears to be successful to the user, but in the backend, the email is discarded without any notice and the email message is never sent. 

## Steps to reproduce
1. Go to any form.
2. Click on **Share**, then **Share by email**. 
3. Type an email address that is valid and which contains a +, for example assistance+forms-formulaires@cds-snc.ca
4. Click the Send button. 
5. You get a confirmation that "The message has been sent". 
6. The message was discarded and was not sent, which is misleading to the user. 

## Expected behaviour
An email address with a plus sign before the @ should be accepted as valid and be processed accordingly. More-so, the Freshdesk email address for support is assistance+forms-formulaires@cds-snc.ca so users wanting to share their form with us run into this issue, which creates confusion. 

## Screenshots or videos
[Bryan R. confirmed](https://gcdigital.slack.com/archives/C03T4FAE9FV/p1715697581176649?thread_ts=1715284330.260639&cid=C03T4FAE9FV) there is a regular expression that filters out email addresses with a + on the client side. 

<img width="720" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeFBlQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3bceaa33c5860c449262a2edb1df821ae0732fb3/image.png" alt="image.png" />

## Additional context
Issue was uncovered while troubleshooting [Freshdesk ticket 17383](https://cds-snc.freshdesk.com/a/tickets/17383).

timarney:
Related PR 

https://github.com/cds-snc/platform-forms-client/pull/3686

PR #3685, opened 2024-05-22
Update email validation + handle alias emails

timarney:
# Summary | Résumé

Fixes: https://github.com/cds-snc/platform-forms-client/issues/3636

To test try sharing a form to 

```
you+something@example.com
```


lib/tests/validation/validation.test.js
```diff
@@ -458,6 +458,8 @@ describe("Gov Email domain validator", () => {
     ["test.hi@cds-snc.ca", true],
```
thiessenp-cds, line 458: I wonder if we should be moving any unit tests we come across to vitests?

timarney: Will create a follow-up PR.

thiessenp-cds (approved):
LGTM :)

PR #3686, opened 2024-05-23
Share modal update error handling

timarney:
# Summary | Résumé

Fixes 2 issues with error handling while testing the share functionality

1) An empty array of emails could be sent in the payload
2) If an empty array was sent to the server it was getting past the validation as the value exists but was empty.

With fixes

Screenshot 2024-05-23 at 9 08 00 AM


thiessenp-cds (approved):
LGTM :)

Issue #3655, opened 2024-05-16
Review error handling for key form builder flows


srtalbot:
Linked to the [error handling investigation for server logs](https://app.zenhub.com/workspaces/gcforms-60cb8929764d71000e481cab/issues/gh/cds-snc/platform-forms-client/3580), review key form builder flows that make changes to a form to ensure errors are announced to the user.

In the client or server components, check if the error property exists. Decide on how we want to handle the error. 

### Acceptance criteria:
- [x] Support form --- email + freshdesk api error - Vivian https://github.com/cds-snc/platform-forms-client/pull/3688
- [x] Contact form --- email + freshdesk api error - Vivian https://github.com/cds-snc/platform-forms-client/pull/3689
- [x]  Unlock publishing  --- email + freshdesk api error - Vivian https://github.com/cds-snc/platform-forms-client/pull/3696
- [x] Sending an email test response - Tim
PR https://github.com/cds-snc/platform-forms-client/pull/3701
 Checked export async function submitForm --- returns id format and show submit error
- [x] Saving a form -- Dave PR https://github.com/cds-snc/platform-forms-client/pull/3698
- [x] Updating permissions -- Pete https://github.com/cds-snc/platform-forms-client/pull/3693
- [x] Downloading responses -- Dave
- [x] Confirming responses -- Dave (NOTE: not converted to server actions yet, will leave for now)
- [x] Decide on the approach for error handling
- [x] Publishing a form -- Tim
- [x] Updating any setting -- Pete (old style REST calls)
- [x] Updating branding (forms management) -- Pete https://github.com/cds-snc/platform-forms-client/pull/3683
- [x] Sharing a JSON through email.  -- Tim https://github.com/cds-snc/platform-forms-client/pull/3686 and https://github.com/cds-snc/platform-forms-client/pull/3685
- [x] Updating "more" modal -- Tim   https://github.com/cds-snc/platform-forms-client/pull/3678
- [x] Quick Settings --- settings panel -- Tim https://github.com/cds-snc/platform-forms-client/pull/3680
- [x] Settings page / tab -- Tim
- [x] Manage ownership -- Vivian
- [x] Deleting a form -- Pete
- [x] Opening a form -- Vivian

timarney:
Publish 

Server
```
} catch (error) {
    return { formRecord: null, error: (error as Error).message };
}
```

Client

```
} catch (e) {
      logMessage.error(e);
      setError(true);
      setErrorCode(500);
}
```

[...]

PR #3688, opened 2024-05-23
* Add missing error handler on client side for support form.

ShadeWyrm:
Part of #3655 

Adds a missing alert for the client side if a server side error occurs.

### DIFF
--- a/app/(gcforms)/[locale]/(support)/support/components/client/SupportForm.tsx
+++ b/app/(gcforms)/[locale]/(support)/support/components/client/SupportForm.tsx
@@ -69,6 +69,12 @@ export const SupportForm = () => {
         </p>
       </Alert.Warning>
       <form id="support" action={formAction} noValidate>
+        {state.error && (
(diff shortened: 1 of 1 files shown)


app/(gcforms)/[locale]/(support)/support/components/client/SupportForm.tsx
```diff
@@ -69,6 +69,12 @@ export const SupportForm = () => {
         </p>
       </Alert.Warning>
       <form id="support" action={formAction} noValidate>
+        {state.error && (
+          <Alert.Danger focussable={true} title={t("error")} className="mb-2 mt-2">
+            <p>{t(state.error)}</p>
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(support)/support/components/client/SupportForm.tsx:74]: We'll probably want to update the error content, same as https://github.com/cds-snc/platform-forms-client/pull/3689/.

This could be in another PR? In which case I can approve this PR?

thiessenp-cds (approved):
Error content can be updated in a future PR.
LGTM :)

PR #3689, opened 2024-05-23
* Add missing contact forms error alert on client side.

ShadeWyrm:
Part of https://github.com/cds-snc/platform-forms-client/issues/3655

Adds a missing alert for the client side if a server side error occurs.

### DIFF
--- a/app/(gcforms)/[locale]/(support)/contact/components/client/ContactForm.tsx
+++ b/app/(gcforms)/[locale]/(support)/contact/components/client/ContactForm.tsx
@@ -69,6 +69,11 @@ export const ContactForm = () => {
         </p>
       </Alert.Warning>
       <form id="contactus" action={formAction} noValidate>
+        {state.error && (
+          <Alert.Danger focussable={true} title={t("error")} className="mb-2 mt-2">
+            <p>{t(state.error)}</p>
+          </Alert.Danger>
+        )}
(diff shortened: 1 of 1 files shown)


thiessenp-cds (dismissed):
LGTM :)

app/(gcforms)/[locale]/(support)/contact/components/client/ContactForm.tsx
```diff
@@ -69,6 +69,11 @@ export const ContactForm = () => {
         </p>
       </Alert.Warning>
       <form id="contactus" action={formAction} noValidate>
+        {state.error && (
+          <Alert.Danger focussable={true} title={t("error")} className="mb-2 mt-2">
+            <p>{t(state.error)}</p>
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(support)/contact/components/client/ContactForm.tsx:74]: I didn't notice until I was looking at a similar PR. Currently this will print out the untranslated error from the server action because a translation key isn't returned.
Screenshot 2024-05-24 at 10 27 58 AM


We could update the server action to instead return a translation key? Or just hard code a translation key in the error since only one error exists. In either case, we'd probably want some new content for this since I don't think we have an error message for this.

I could approve this PR and the error content could be updated in another PR?

thiessenp-cds (approved):
Error content can be updated in a future PR.
LGTM :)

PR #3693, opened 2024-05-27
fix/Updates user permissions to handle errors

thiessenp-cds:
# Summary | Résumé

Updates user permissions to handle errors.

Example Error:

Screenshot 2024-05-27 at 9 40 40 AM

## Test

Probably the easiest way to test this is to locally checkout this branch and have the action return an error:
-action /app/(gcforms)/[locale]/(app administration)/admin/(with nav)/accounts/[id]/manage-permissions/actions.ts
-example url http://localhost:3000/en/admin/accounts/clwkrfijj000j8evpjmcp8xvs/manage-permissions

### DIFF
--- a/app/(gcforms)/[locale]/(app administration)/admin/(with nav)/accounts/[id]/manage-permissions/components/client/Privilege.tsx
+++ b/app/(gcforms)/[locale]/(app administration)/admin/(with nav)/accounts/[id]/manage-permissions/components/client/Privilege.tsx
@@ -2,6 +2,7 @@
 import { useTranslation } from "@i18n/client";
 import { PermissionToggle } from "./PermissionToggle";
 import { updatePrivileges } from "../../actions";
+import { toast } from "@formBuilder/components/shared";
 
 export const Privilege = ({
   canManageUsers,
@@ -26,7 +27,10 @@ export const Privilege = ({
           offLabel={t("off")}
           description={description || ""}
           handleToggle={async () => {
-            await updatePrivileges(userId, privilegeId, active ? "remove" : "add");
+            const result = await updatePrivileges(userId, privilegeId, active ? "remove" : "add");
+            if (result?.error) {
+              toast.error(t("error.permissions"), "wide");
+            }
           }}
         />
       ) : (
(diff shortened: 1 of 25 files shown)


app/(gcforms)/[locale]/(app administration)/admin/(with nav)/accounts/[id]/manage-permissions/components/client/Privilege.tsx
```diff
@@ -26,7 +27,10 @@ export const Privilege = ({
           offLabel={t("off")}
           description={description || ""}
           handleToggle={async () => {
-            await updatePrivileges(userId, privilegeId, active ? "remove" : "add");
+            const result = await updatePrivileges(userId, privilegeId, active ? "remove" : "add");
+            if (result?.error) {
+              toast.error(t("error.permissions"), "wide");
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(app administration)/admin/(with nav)/accounts/[id]/manage-permissions/components/client/Privilege.tsx:32]: In the future we should decide on a pattern for returning errors from an action. This is a hardcoded error and would break if the server action could return two or more errors.
One idea is to have the server action return the error key.
Future thought :)

i18n/translations/fr/admin-users.json
```diff
@@ -54,5 +54,8 @@
   },
   "systemAdministration": "Administration du système",
   "accountAdministration": "Administration des comptes",
-  "userAdministration": "Administration des l’utilisateurs"
+  "userAdministration": "Administration des l’utilisateurs",
+  "error": {
+    "permissions": "[FR]Failed to update permissions."
```
timarney, line 59: We should get a translation for this before merging.

thiessenp-cds: Sounds good to me :)
@anikbrazeau would it be possible to get a translation for this string "Failed to update permissions."?

timarney:
@thiessenp-cds --- pushed a tiny commit to update the loading animation colour to use something from the app colour palette  --- ping to to approve once we have the string.

Tested and it's working as expeected.

i18n/translations/fr/admin-users.json
anikbrazeau, line 59:
```suggestion
    "permissions": "Les autorisations n'ont pas pu être mises à jour."
```

timarney (approved):

PR #3708, opened 2024-05-28
Updates review page to use section new titles

thiessenp-cds:
# Summary | Résumé

Updates review page to use section new titles

### DIFF
--- a/components/clientComponents/forms/Review/Review.tsx
+++ b/components/clientComponents/forms/Review/Review.tsx
@@ -35,9 +35,12 @@ export const Review = (): React.ReactElement => {
   const questionsAndAnswers = getGroupHistory()
     .filter((key) => key !== "review") // Removed to avoid showing as a group
     .map((key) => {
+      const titleLanguageKey = getLocalizedProperty("title", lang) as "titleEn" | "titleFr";
+      const title = reviewGroups?.[key]?.[titleLanguageKey];
       return {
         id: key,
         name: reviewGroups[key].name,
+        title,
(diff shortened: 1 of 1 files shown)


timarney:
@thiessenp-cds 

Looks like your not using titleEn, titleFr for the review output?

Screenshot 2024-05-29 at 8 36 33 AM

Seeing


Screenshot 2024-05-29 at 8 34 26 AM


Screenshot 2024-05-29 at 8 36 15 AM

lib/hooks/useGCFormContext.tsx
```diff
@@ -107,11 +109,11 @@ export const GCFormsProvider = ({
...
 
-  // Note: once localized versions of the section name have been added, update the below to use
-  // something like `getLocalizedProperty("name", locale || "en")`
-  const getGroupTitle = (groupId: string | null) => {
+  const getGroupTitle = (groupId: string | null, language: Language) => {
     if (!groupId) return "";
-    return groups[groupId]?.name || "";
+    const titleLanguageKey = getLocalizedProperty("title", language) as "titleEn" | "titleFr";
+    const title = groups?.[groupId]?.[titleLanguageKey];
+    return title || groups[groupId]?.name || "";
```
thiessenp-cds, line 116: I was bit on the fence here. If there is no group title, then showing the group name seems like better UX. But then the function is kind of lying :) where it "may" return a group title :)

timarney: Better not to show as the name won't be a translated string.

timarney (approved):

PR #3712, opened 2024-05-30
Fix catch all --- after filtering rules

timarney:
# Summary | Résumé

Updates group select to handle filtering catch-all + disabling add button when rules are maxed out.


thiessenp-cds (approved):
LGTM :)

PR #3647, opened 2024-05-15
Pre-Publish Functionality / Salesforce Publish Integration

ShadeWyrm:
# Summary | Résumé

https://github.com/cds-snc/platform-forms-client/issues/2605
+
https://github.com/cds-snc/platform-forms-client/issues/3313

# Test instructions | Instructions pour tester la modification

Create form. Go to settings page, set the new setting.

# Unresolved questions / Out of scope | Questions non résolues ou hors sujet

-> Out of scope: Full [User] Lifecycle.
-> Department Identification

# Pull Request Checklist

Please complete the following items in the checklist before you request a review:

- [ ] Have you completely tested the functionality of change introduced in this PR? Is the PR solving the problem it's meant to solve within the scope of the related issue?
- [x] The PR does not introduce any new issues such as failed tests, console warnings or new bugs.
- [x] If this PR adds a package have you ensured its licensed correctly and does not add additional security issues?
- [x] Is the code clean, readable and maintainable? Is it easy to understand and comprehend.
- [x] Does your code have adequate comprehensible comments? Do new functions have [docstrings](https://tsdoc.org/)?
- [ ] Have you modified the change log and updated any relevant documentation?
- [ ] Is there adequate test coverage? Both unit tests and end-to-end tests where applicable?
- [ ] If your PR is touching any UI is it accessible? Have you tested it with a screen reader? Have you tested it with automated testing tools such as axe?

Issue #3313, opened 2024-03-11
Forms Publishing Page/Modal + (Non-)Administrative Form Setting


Abi-Nada:
### User Story

As a member of the GC Forms team,
I want to collect data on the forms being published,
So that we can make informed decisions and understand how the application is being used by users, without manually validating each individual published form.

### Acceptance Criteria:
- [x] Publishing Page modal designs are updated based off testing findings/feedback
- [x] We include a way to collect whether or not the form is being used for administrative purposes (consideration: we are not making that designation for them. We’re asking the client to make a selection somewhere in the product, rather than us seeing what kind of form it is and determining whether we think that’s administrative or not)
- [x] Implementation of designs

anikbrazeau:
Latest content: https://www.figma.com/file/yQmBXLZ07qFElbRE89kdal/GC-Forms---Preview-and-Share?type=design&node-id=3844-7394&mode=design&t=nCyWyPbB9vw5YpCV-4

anikbrazeau:
**Latest content**: https://www.figma.com/file/aFRMxTx2rGCJp2xTYdYc5Z/GC-Forms-SalesForce-Integration?type=design&node-id=3002-78&mode=design&t=ksJFzlI5b88Q7it5-0

Abi-Nada:
Figma file: https://www.figma.com/file/aFRMxTx2rGCJp2xTYdYc5Z/GC-Forms-SalesForce-Integration?type=design&node-id=3002-78&mode=design&t=KVCvJIL82zt9jJso-0

PR #3714, opened 2024-05-30
Admin vs Non-Admin Usage & Settings UX Update

ShadeWyrm:
This PR is a large chunk of the requirements from #3313

It is split off from #3647 to avoid issues with Salesforce Integration.

---

For Testing...

The UI/UX changes are on the Form Settings Page, and Publish page.

Publish Page:
* The User can no longer publish if the Admin vs Non-Admin use is set to Unset. (New Form)
* The User will get a happy checkmark that they can publish if it's set.

Settings Page:
* The user can pick between the two options for Administrative vs Non-Administrative Use of a form. This value is saved to the template store and by proxy the db.

The pages should work the same and have valid relevant text in both languages.


timarney:
@ShadeWyrm can you please add some test instructions to the PR description so folks without context can review without too much digging on what the PR does ?

timarney:
@ShadeWyrm this is looking good 

Thinking we can remove the getter / setter for the `useTemplateStore`  i.e.  https://github.com/cds-snc/platform-forms-client/commit/3af5a8e96726b3eaa0c18989a57300be6aeedcdb does that work?

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/components/ResponseDelivery.tsx
```diff
@@ -33,6 +33,12 @@ enum DeliveryOption {
   email = "email",
 }
 
+export enum PurposeOption {
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/components/ResponseDelivery.tsx:36]: Thinking we should start commenting these to make it easier to folks looking at these types

i.e. something like 


```
/*
  * PurposeOption is used to determine the purpose of the form
  * admin: The form is used to collect personal information
  * nonAdmin: The form is used to collect non-personal information
  */
```



I know these don't exist for Vault etc... but might make it easier to folks onboarding or who haven't touched this part of the code.

ShadeWyrm:
> @ShadeWyrm this is looking good
> 
> Thinking we can remove the getter / setter for the `useTemplateStore` i.e. [3af5a8e](https://github.com/cds-snc/platform-forms-client/commit/3af5a8e96726b3eaa0c18989a57300be6aeedcdb) does that work?
> 
> Get
> 
> ```js
>  formPurpose: s.formPurpose,
> ```
> 
> Update
> 
> ```js
>  updateField("formPurpose", purposeOption);
> ```

Yah that should work just fine, it was a hold over from my testing, and figuring out the template store. I had assumed it was necessary, but the Linter brought that into question on my end. :) So just the property should be enough.

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/components/ResponseDelivery.tsx
Comment (ShadeWyrm) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/components/ResponseDelivery.tsx:36]: Agreed.

lib/tests/templates.test.ts
```diff
@@ -65,12 +65,14 @@ const buildPrismaResponse = (
   jsonConfig: object,
   isPublished = false,
   deliveryOption?: DeliveryOption,
+  formPurpose?: string,
```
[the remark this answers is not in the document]
ShadeWyrm: Probably makes sense to do so, you're right.

timarney:
https://github.com/cds-snc/platform-forms-client/pull/3714#issuecomment-2142090035

Do you want to cherry pick that commit in or make the updates on you end?

timarney (approved):
Nice work :)

PR #3718, opened 2024-05-30
Use change key - from multi-next actions

timarney:
# Summary | Résumé

Adds change key update for next actions to ensure an array of actions triggers a render / update.


thiessenp-cds (approved):
Nice! LGTM :)

Issue #3716, opened 2024-05-30
Issue moving elements up/down/duplicating in tree view


anikbrazeau:
## Description
Unable to reorder/rearrange form elements within a section

## Steps to reproduce
1. Create new section
2. Add contact information element
3. Try to reorder component parts
4. Nothing happens

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Preview
- **Language**: English
- **Viewport**: Desktop

## Expected behaviour
Move when clicked

## Screenshots or videos

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeFVsQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--dba69d9082f7720d94cd8405e1562ad35b803f41/Screen%20Recording%202024-05-30%20at%201.58.08%E2%80%AFPM.mov)

timarney:
Moving elements up / down using "Move up" or "Move down" should be working now --- didn't look at duplicate

timarney:
Will look at duplicate today.

timarney:
PR for duplicate https://github.com/cds-snc/platform-forms-client/pull/3727

anikbrazeau:
Tested! Fixed! Thanks!

PR #3727, opened 2024-06-03
Handle duplicate for groups

timarney:
# Summary | Résumé

Update duplicate element code to work with groups

Fix for https://github.com/cds-snc/platform-forms-client/issues/3716


thiessenp-cds (approved):
Works as expected. LGTM :)

PR #3734, opened 2024-06-03
Logic view a11y initial updates

timarney:
# Summary | Résumé

Updates to ui

- Use button vs dv
- CSS updates for focus within + button focus.


thiessenp-cds (approved):
LGTM :)

Issue #3739, opened 2024-06-04
Missing translation keys for default section names and next button


anikbrazeau:
## Description
Missing translations

## Steps to reproduce
1. Go to conditional logic env
2. Go to "Edit"
3. Go to "Français"
4. Open default groups/sections

## Expected content

Previously: Start
EN: `Start`
FR: `Début`

Previously: Introduction
EN: `Form introduction`
FR: `Introduction au formulaire`

Previously: Policy
EN: `Privacy statement`
FR: `Avis de confidentialité`

Previously: Confirmation
EN: `Confirmation message`
FR: `Message de confirmation`

Previously: End
EN: `End`
FR: `Fin`

Previously: Next
EN: `Continue`
FR: `Continuer`

## Screenshots or videos

<img width="526" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeFUwQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b2c0a955ec4816b36dd3af0dd5847adca0f0d4f8/Screenshot%202024-06-04%20at%207.00.01%E2%80%AFAM.png" alt="Screenshot 2024-06-04 at 7.00.01 AM.png" />

<img width="1367" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeGMwQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--43881f93cf8ba543904d641cbf79fab3f45b4ee9/Screenshot%202024-06-04%20at%207.07.31%E2%80%AFAM.png" alt="Screenshot 2024-06-04 at 7.07.31 AM.png" />

## Additional context
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Staging / CL PR
- **Language**: French
- **Viewport**: Desktop

anikbrazeau:
#3740

anikbrazeau:
Completed with PR #3767

PR #3740, opened 2024-06-04
Conditional logic content translations and refinements

anikbrazeau:
Addresses part of #3739


i18n/translations/en/form-builder.json
```diff
@@ -865,6 +865,11 @@
       }
     },
     "sectionTitle": "Section title",
+<<<<<<< branching-content
```
timarney, line 868: @anikbrazeau looks like this was a bad merge conflict fix

timarney (approved):

PR #3743, opened 2024-06-04
Update init form to allow direct link to page for groups

timarney:
# Summary | Résumé

 Fixes issue where a direct link to `/en/form-builder/0000/edit`
 
 Was not initializing the group sections.


thiessenp-cds (approved):
Tested with and without groups enabled in both Chrome and Edge. Works great!
LGTM :)

Issue #3600, opened 2024-05-13
Translation spreadsheet output


Abi-Nada:
### Description:

Within the conditional logic work, we need to ensure that clients can still export a spreadsheet within the Translations tab, so they can send it over to the form to their translation team.

### Acceptance Criteria:
- [ ] Validate that the spreadsheet export appears correctly, with the questions in the right order.

Issue #3064, opened 2024-01-02
Panel: Translations


Abi-Nada:
### Description:
Currently in the Conditional Logic work, the Translation page is broken (we've pulled out question numbers, but the translation view might still have this). We need to ensure that all elements and questions are in the right order, and sections are reflected.

### Acceptance Criteria:
- [x] Ensure text boxes follow the logical flow/order of the form
- [x] Incorporate Sections into the page

PR #3746, opened 2024-06-05
Fix: Translate with Groups

ShadeWyrm:
Adds the Grouping logic to the Translate page and Download Translations function.

PR covers:
#3064 
&
#3600


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/translate/components/DownloadCSVWithGroups.tsx
```diff
@@ -0,0 +1,142 @@
...
+          formatText(group.titleFr),
+        ]);
+
+        groupElements.map((element) => {
+          elementIndex++;
+          parseElement(element, elementIndex);
+        });
+      });
+    } else {
+      // If the form doesn't have groups, use the old method...
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/translate/components/DownloadCSVWithGroups.tsx:96]: We can remove this 

All entry points for "group" wrappers already.

```
export default async function Page({ params: { id } }: { params: { id: string } }) {
  const conditionalLogic = await allowGrouping();

  return (
    <>
      {!conditionalLogic && <EditNavigation id={id} />}
      {conditionalLogic ? <TranslateWithGroups /> : <Translate />}
    </>
  );
}
```

Same sort of thing for the right panels

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/translate/components/TranslateWithGroups.tsx
```diff
@@ -243,13 +314,36 @@ export const TranslateWithGroups = () => {
...
+          {
+            // Loop through each group and render the elements in the group
+          }
+          {groups &&
+            Object.keys(groups).map((groupKey, index) => {
+              const thisGroup = groups[groupKey];
+              if (groupKey == "review" || groupKey == "end") return null;
+              return (
+                <div key={groupKey}>
+                  <SectionTitle>Section # {index}</SectionTitle>
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/translate/components/TranslateWithGroups.tsx:326]: If we keep this we need to use a translation sting here

lib/utils/form-builder/groupedFormHelpers.ts
```diff
@@ -0,0 +1,57 @@
...
+    return elementCollection[elId];
+  });
+
+  sortedElements.push(...groupElements);
+
+  return sortedElements;
+};
+
+//----
+//TODO: Move this into the template store, and use the template store to get the the collection of elements.
```
timarney, line 25: Can remove this comment --- we'll keep as a util vs moving to the template store

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/translate/components/TranslateWithGroups.tsx
Comment (timarney) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/translate/components/TranslateWithGroups.tsx:326]: Well use the section names to match the Tree View

timarney:
UI part is looking great -- 

The download still references Groups and Nums vs Section names

"Group 1","",""
"Group 2","",""

timarney (approved):
Nice work.

I'm sure we'll iterate on this further as we go but let's get this merged now and move things forward.

PR #3749, opened 2024-06-05
Fix TreeView rename

timarney:
# Summary | Résumé

- Update to add default text when there's no text filled in for form elements
- Pulled page title and or desc depending on type of form element
- Update group set title to use lang prop to ensure text goes back to the right language.

Also fixes read-only error for review and end groups


thiessenp-cds (approved):
LGTM :)

Issue #3750, opened 2024-06-05
UI fixes to Settings Page


Abi-Nada:
## Description

Figma File: https://www.figma.com/design/aFRMxTx2rGCJp2xTYdYc5Z/GC-Forms-SalesForce-Integration?node-id=3059-910&t=FLExykdEc8JWIkHF-0 

- [x] Purple help text is currently under "Data Classification" heading, when it should be under the "Settings" heading
- [x] Help text in incorrect it should read "These settings can only be changed before publishing." "Ces paramètres ne peuvent être modifiés qu’avant la publication."
- [x] Spacing is inconsistent and too cramped between each section: Data classification, Response delivery option, Purpose and use. Ensure there is a class of "mb-10" between each section.
- [x] text correction: Response delivery option should be "Delivery option" "Option de livraison"
- [x] text correction: "Retrieve responses within 45 days." should be "GC Forms holds data temporarily for 30 days." "Formulaires GC conserve les données temporairement pour 30 hours."
- [x] Bold the line "Help us understand how form responses will be used"
- [x] Italics "Administrative purposes," "Non-administrative purposes,"
- [x] Alignment of bullets "Personal information used to make a decision about an individual" and hint text is misaligned, try class "ml-12"
- [x] Help text for choosing a delivery option is missing
- [x] Help text for Administrative purposes reflects information for choosing a delivery option (the current info here should be for the missing help text in the point above^^)

## Steps to reproduce
1. Go to Staging -> the Settings page
2. Observe above flags

## Expected behaviour
Figma file: https://www.figma.com/design/aFRMxTx2rGCJp2xTYdYc5Z/GC-Forms-SalesForce-Integration?node-id=3059-910&t=FLExykdEc8JWIkHF-0

## Screenshots or videos
Currently in Staging:
<img width="1519" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL3c4QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--77f45da364247748addbb1971a6ccc7dcd3a9a75/Screenshot%202024-06-05%20at%202.32.06%E2%80%AFPM.png" alt="Screenshot 2024-06-05 at 2.32.06 PM.png" />

PR #3752, opened 2024-06-06
Minor UI Fixes for Settings Page

ShadeWyrm:
Requested changes are marked in #3750.

* Purple help text is currently under "Data Classification" heading, when it should be under the "Settings" heading
* Help text in incorrect it should read "These settings can only be changed before publishing." "Ces paramètres ne peuvent être modifiés qu’avant la publication."
* Spacing is inconsistent and too cramped between each section: Data classification, Response delivery option, Purpose and use. * * Ensure there is a class of "mb-10" between each section.
* text correction: Response delivery option should be "Delivery option" "Option de livraison"
* text correction: "Retrieve responses within 45 days." should be "GC Forms holds data temporarily for 30 days." "Formulaires GC conserve les données temporairement pour 30 hours."
* Bold the line "Help us understand how form responses will be used"
* Alignment of bullets "Personal information used to make a decision about an individual" and hint text is misaligned, try class "ml-12"


timarney (approved):
Approving based on feedback from design.

PR #3758, opened 2024-06-06
Update logic view nav and remove React flow default a11y

timarney:
# Summary | Résumé

- Updates styling for Logic view nav under the tab buttons
- Updates React flow to remove default keyboard a11y --- we'll just tab to each button.
- Fixes bug where when the active ID was set to "end" - end would be selected when you visit the logic tab which shouldn't be possible. 

https://github.com/cds-snc/platform-forms-client/assets/62242/7b51b12e-df60-4cbc-aa3a-ac95e5733b2a


thiessenp-cds (approved):
Nice! LGTM :)

PR #3763, opened 2024-06-07
Fix publish issue with form purpose

timarney:
# Summary | Résumé

Update useAllowPublish to pull from purpose from template state vs template.form

Noting --- skipping the publish test in this PR to unblock staging issue.

Will fix the test in a follow-up.


ShadeWyrm (approved):
Works as intended!~

ShadeWyrm (approved):

PR #3765, opened 2024-06-07
Fix allow publish test

timarney:
# Summary | Résumé

Fixes skipped test from https://github.com/cds-snc/platform-forms-client/pull/3763

Also adds a couple of notes and renames some variables to bring a little more clarity.


ShadeWyrm (approved):
Tested the publish being set/unset and the tests, and it looks to be good!~

PR #3766, opened 2024-06-07
Fix/3735 wcag 3

thiessenp-cds:
# Summary | Résumé

Adds a heading to a section wrapper for the Right Panel. This helps AT users by labeling the area that will show up in the page outline.
The H2 is set to "Right Panel". 
In the future, we could add a description and more if users are confused about where they are when they enter the panel.

Fixes and updates the logic view buttons to be a little more descriptive (hopefully).
Section button label is set to: "Edit section {SECTION_NAME}"
Section logic view label is set to: "Edit rules for section {SECTION_NAME}}"
I suspect the labels could be improved but hopefully this will be descriptive enough for now.

### DIFF
--- a/.github/workflows/release_generator.yml
+++ b/.github/workflows/release_generator.yml
@@ -1,7 +1,7 @@
 on:
   push:
     branches:
-      - main
+      - develop
 
 permissions:
   contents: write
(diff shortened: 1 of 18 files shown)


timarney:
@thiessenp-cds can you add a fix for the tabs so we can press enter to navigate to this PR?

timarney (approved):

timarney:
These are some nice updates @thiessenp-cds it's a solid improvement.

PR #3775, opened 2024-06-10
Fix saving from quick panel

timarney:
# Summary | Résumé

Fixes issue where template possibly doesn't exist when trying to save via the quick panel

Bug

https://github.com/cds-snc/platform-forms-client/assets/62242/ad251364-0f36-441a-84b6-86c04cad782c


thiessenp-cds (approved):
LGTM :)

PR #3778, opened 2024-06-10
fix: trusted host check and NEXTAUTH_URL

bryan-robitaille:
# Summary | Résumé
Removes any remaining reference to NEXTAUTH_URL in the codebase as we can implicitly trust that the Elastic Load Balancer safely sets the Host header and ignores any header provided in the original request.

Confirmed here: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#host-header-preservation

It would also appear that Auth.js guidance is now that the `trusted_host` should always be set to true: https://authjs.dev/reference/core#trusthost

### DIFF
--- a/lib/auth/nextAuth.ts
+++ b/lib/auth/nextAuth.ts
@@ -113,7 +113,7 @@ export const {
     maxAge: 2 * 60 * 60, // 2 hours
     updateAge: 30 * 60, // 30 minutes
   },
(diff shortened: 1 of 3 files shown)


thiessenp-cds (approved):
Works as expected. LGTM! :)
Screenshot 2024-06-10 at 3 17 12 PM
Screenshot 2024-06-10 at 3 19 16 PM

PR #3808, opened 2024-06-11
Fix: mfa resend when no session

bryan-robitaille:
# Summary | Résumé

take 2 of #3801 

Updates the handling of 2fa "resend code" path errors.

    Adds a log entry that identifies the email address trying to request a new token {"level":"warn","time":1718129549075,"msg":"Failed to send new verification code. Reason: Missing 2FA session for user bryan.robitaille@cds-snc.ca."}
    Fixes the error path to ensure the server redirects the user to /auth/login if a session does not exist instead of relying on the browser. In NODE_ENV Production the error was being sanitized and wasn't being properly handled client side.

Also modifies the "form submission" event log to identify that it is a form submission and not just listing the form ID.


thiessenp-cds (approved):
Works as expected. LGTM :)

PR #3824, opened 2024-06-12
fix: Add early return for responses page when a form hasn't been saved yet 

timarney:
# Summary | Résumé

Fixes issue where a user could land on the responses page with a form that has never been saved (doesn't have form responses).

Adds early return to avoid API lookup which throws

 ```
 Forms
 - Template 0000 must have associated owners to access responses
 ```

### DIFF
--- a/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/page.tsx
+++ b/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/page.tsx
@@ -38,6 +38,21 @@ export default async function Page({
     overdueAfter: Number(await getAppSetting("nagwarePhaseEncouraged")),
   };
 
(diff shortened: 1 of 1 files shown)


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/page.tsx
```diff
@@ -38,6 +38,21 @@ export default async function Page({
     overdueAfter: Number(await getAppSetting("nagwarePhaseEncouraged")),
   };
 
+  // Handle the case where the user is not authenticated and the form ID is 0000
+  // In this case we know no responses will be returned, so we can skip the API call
+  if (isAuthenticated && id === "0000") {
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/page.tsx:43]: This should check for a UUID vs just 0000

dsamojlenko (approved):

timarney:
Noting other string will still fall-through but will merge this.

PR #3834, opened 2024-06-14
chore: add first batch of healthcheck logs

craigzour:
# Summary | Résumé

- Adds first batch of HealthCheck logs to feed monitoring dashboard in CloudWatch

### DIFF
--- a/app/(gcforms)/[locale]/(user authentication)/auth/register/action.ts
+++ b/app/(gcforms)/[locale]/(user authentication)/auth/register/action.ts
@@ -15,6 +15,7 @@ import {
   CognitoIdentityProviderServiceException,
 } from "@aws-sdk/client-cognito-identity-provider";
 import { cognitoIdentityProviderClient } from "@lib/integration/awsServicesConnector";
+import { logMessage } from "@lib/logger";
 
 export interface ErrorStates {
   authError?: {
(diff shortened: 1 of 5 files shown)


app/(gcforms)/[locale]/(user authentication)/auth/register/action.ts
```diff
@@ -131,9 +133,11 @@ export const register = async (
 
   try {
     await cognitoIdentityProviderClient.send(signUpCommand);
+    logMessage.info("HealthCheck: cognito sign-up success");
   } catch (err) {
-    // if there is an error, forward the status code and the error message as the body
+    logMessage.info("HealthCheck: cognito sign-up failure");
```
Comment (patheard) [inline app/(gcforms)/[locale]/(user authentication)/auth/register/action.ts:138]: For all of the error state logs, what would you think of changing them to use `error` instead of `info`?  That would give us the specific healthcheck messages and a generic `ERROR` search term if we wanted to group all failure conditions.
```suggestion
    logMessage.error("HealthCheck: cognito sign-up failure");
```

Comment (craigzour) [reply app/(gcforms)/[locale]/(user authentication)/auth/register/action.ts:138]: The reason is that they would also appear in our Slack notifications..

patheard (approved):
Exactly what I was hoping for - thanks!

PR #3845, opened 2024-06-17
chore: reduce logging noise

wmoussa-gc:
Remove the logs that are created each time a user enters an invalid MFA code.

### DIFF
--- a/lib/auth/nextAuth.ts
+++ b/lib/auth/nextAuth.ts
@@ -118,9 +118,8 @@ export const {
   debug: process.env.NODE_ENV !== "production",
   logger: {
     error(error) {
-      if (error instanceof CredentialsSignin) {
-        logMessage.warn(`NextAuth - CredentialsSignin exception: ${JSON.stringify(error)}.`);
-      } else {
+      if (!(error instanceof CredentialsSignin)) {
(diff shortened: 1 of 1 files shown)


lib/auth/nextAuth.ts
```diff
@@ -118,9 +118,8 @@ export const {
   debug: process.env.NODE_ENV !== "production",
   logger: {
     error(error) {
-      if (error instanceof CredentialsSignin) {
-        logMessage.warn(`NextAuth - CredentialsSignin exception: ${JSON.stringify(error)}.`);
-      } else {
+      if (!(error instanceof CredentialsSignin)) {
+        // Not a CredentialsSignin error which is for invalid credentials (including 2FA)
```
bryan-robitaille, line 122:
```suggestion
        // Not a CredentialsSignin error which is for invalid 2FA credentials

bryan-robitaille (changes requested):
Just a quick change to ensure there is no confusion with the comment.  Only invalid 2fa codes throw the `CredentialsSignIn` error.

craigzour (approved):

bryan-robitaille (approved):

PR #3851, opened 2024-06-17
Fix: Sort questions in Add Rule question list

### LINKED ISSUE #3702: Moving a question down doesn't change it's place in the list when adding a rule
**OS and Browser**
Mac
Chrome 

**Instructions to recreate**
1. Create a number element
2. Create a radio button for a conditional question
3. Copy the number element
4. Move the number element to the last position in the form
5. Add a rule to the radio button - notice that the dropdown doesn't change the position

In the picture above "unique client ID:" is actually the second last element in the form.

dsamojlenko:
# Summary | Résumé

- Includes a minor update to the sortByLayout function to operate on a copy of the elements array rather than the elements array itself.
- Adds a sortByGroups util function
- Fixes #3702

### DIFF
--- a/lib/utils/form-builder/index.ts
+++ b/lib/utils/form-builder/index.ts
@@ -92,7 +92,7 @@ export const sortByLayout = ({
   layout: number[];
   elements: FormElement[];
 }) => {
-  return elements.sort((a, b) => {
+  return [...elements].sort((a, b) => {
     return layout.indexOf(a.id) - layout.indexOf(b.id);
   });
 };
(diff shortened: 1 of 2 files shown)


lib/utils/form-builder/index.ts
```diff
@@ -92,11 +92,23 @@ export const sortByLayout = ({
...
 
+export const sortByGroups = ({
+  form,
+  elements,
+}: {
+  form: FormProperties;
+  elements: FormElement[];
+}) => {
+  const groups = form.groups || {};
+  const layout = getLayoutFromGroups(form, groups);
```
timarney, line 108: If groups is empty here can we skip the  getLayoutFromGroups call?

timarney (approved):

PR #3859, opened 2024-06-18
feat: Conditional Input History - Form Builder

thiessenp-cds:
# Summary | Résumé

Adds the conditional input history feature to the **form-builder and Review page only**.  A followup PR will add this feature to Forms-forms submissions.

More Info:
Using the group history logic to allow a user to navigate the conditional logic in a form, update the user answers (form input) to only include answers that follow the users navigation. 
For example, a user may go down conditional logic option A but then later go back and choose conditional logic option B, then follow the option B branch. So the questions answered on option A's branch should not be included in the form submission, only the answers from the option B branch.

## Testing

Note:  show/hide value filtering is applied on the form-builder only. So submitting a form will still have the previous functionality of including both visible and hidden values from any show/hide elements.

Ideally test creating a form with conditional logic in both the Form-Builder Preview. Try going down a few conditional logic branches and check if the answers you inputted match the "conditional input history" you navigated within the form.

Here is an example manual test that focusses on making sure the correct "history" of answers is on the Review page. The user chose the A conditional "branch", so only those answers should be in the form submission. Not the branching answers the user also answered from going down the B conditional and C conditional branches  (though those answers remain in the form inputs for user convenience).

https://github.com/cds-snc/platform-forms-client/assets/107579368/34e109cb-032b-47ef-83f9-e855b3e21c3d


components/clientComponents/forms/Review/Review.tsx
```diff
@@ -22,9 +22,69 @@ type QuestionAnswer = {
...
+  group,
+  formRecord,
+  lang,
+}: {
+  group: ReviewGroup;
+  formRecord: TypeOmit<FormRecord, "name" | "deliveryOption">;
+  lang: string;
+}): React.ReactElement => {
+  const { t } = useTranslation("review");
+  function getElementNameById(id: string | number) {
```
timarney, line 35: We should be able to move this outside the component ya?

components/clientComponents/forms/Review/Review.tsx
```diff
@@ -22,9 +22,69 @@ type QuestionAnswer = {
...
+    const element = formRecord.form.elements.find((item) => String(item.id) === String(id));
+    return element ? element.properties?.[getLocalizedProperty("title", lang)] : t("unknown");
+  }
+  function getQuestionAndAnswer(element: QuestionAnswer) {
```
timarney, line 39: We should be able to move this outside the component ya?

components/clientComponents/forms/Review/Review.tsx
```diff
@@ -22,9 +22,69 @@ type QuestionAnswer = {
...
+    return { question, answer };
+  }
+  return (
+    <dl className="my-10">
+      {Array.isArray(group.elements) &&
+        group.elements.map((element) => {
+          const { question, answer } = getQuestionAndAnswer(element);
+          return (
+            <div key={Object.keys(element)[0]} className="mb-8">
+              <dt className="font-bold mb-2">{question}</dt>
```
timarney, line 51: Do you have the Tailwind ext installed ?

Will catch things like these classes being in the wrong order --- if not let's chat later about this.

Screenshot 2024-07-18 at 7 18 17 AM


Screenshot 2024-07-18 at 7 18 49 AM

components/clientComponents/forms/Review/Review.tsx
```diff
@@ -33,60 +93,67 @@ export const Review = ({ language }: { language: Language }): React.ReactElement
     function formatElementValue(elementName: string | null) {
```
timarney, line 93: Let's move this outside of the component

components/clientComponents/forms/Review/Review.tsx
```diff
@@ -33,60 +93,67 @@ export const Review = ({ language }: { language: Language }): React.ReactElement
...
       const value = formValues[elementName as keyof typeof formValues];
       if (Array.isArray(value)) {
-        return (value as Array<string>).join(", ");
+        return (value as Array<string>).join(", ") || "-";
       }
       return value || "-";
     }
+
+    function structureElements(elements: string[]) {
```
timarney, line 101: Let's move this outside of the component

lib/utils/form-builder/groupsHistory.ts
```diff
@@ -32,3 +33,51 @@ export const getPreviousIdFromCurrentId = (currentId: string, history: string[])
   }
   return null;
 };
+
+export const getRelevantValues = (
```
timarney, line 37: Let's add a doc block here to explain what this function is doing.

lib/utils/form-builder/groupsHistory.ts
```diff
@@ -32,3 +33,51 @@ export const getPreviousIdFromCurrentId = (currentId: string, history: string[])
...
+    groupValueIntersection.forEach((relevantElement: string) => {
+      const key = relevantElement as keyof typeof Response;
+      const value = relevantElement as keyof typeof allFormValues;
+      relevantValues[key] = allFormValues[value];
+    });
+  });
+  return relevantValues;
+};
+
+export const filterNonRelevantValues = (allValues: Responses, valuesInHistory: Responses) => {
```
timarney, line 56: Let's add a doc block here to explain what this function is doing.

lib/utils/form-builder/groupsHistory.ts
```diff
@@ -32,3 +33,51 @@ export const getPreviousIdFromCurrentId = (currentId: string, history: string[])
...
+  if (!allValues || !valuesInHistory) return {} as Responses;
+  const allValuesEmtpy = {} as Responses;
+  for (const key in allValues) {
+    allValuesEmtpy[key] = "";
+  }
+  return { ...allValuesEmtpy, ...valuesInHistory };
+};
+
+export const removeCustomFormValues = (values: Responses) => {
```
timarney, line 65: Let's look to come up with a better name for this 

Maybe removeFormContext values

lib/formContext.ts
```diff
@@ -389,6 +389,66 @@ export const checkRelatedRulesAsBoolean = (
...
+
+    return false;
+  });
+};
+
+/**
+ *
+ * @param elements Group type elements that are string[]
+ * @param elementsShown
+ * @returns
```
timarney, line 421: Can remove this doc block here and or fill in the returns

lib/formContext.ts
```diff
@@ -389,6 +389,66 @@ export const checkRelatedRulesAsBoolean = (
...
+};
+
+export const getElementIdsAsNumber = (elements: string[]) => {
+  if (!Array.isArray(elements)) {
+    return [];
+  }
+  return elements.map((element) => Number(element));
+};
+
+export const rebuildValuesFromShownElements = (values: Responses, elementsShown: FormElement[]) => {
```
timarney, line 437: Unclear as to what rebuilding means :)

lib/utils/form-builder/groupsHistory.ts
timarney: i.e. not sure what "Relevant" means

thiessenp-cds: What about filterValuesByVisitedGroup()

lib/utils/form-builder/groupsHistory.ts
thiessenp-cds: Yup will change

lib/tests/formBuilder.test.js
```diff
@@ -22,17 +23,16 @@ describe("getFormInitialValues() tests", () => {
         },
       ],
       8: "",
-      currentGroup: "",
     };
-    const initialValues = getFormInitialValues(
+    const initialValues = removeCustomFormValues(getFormInitialValues(
```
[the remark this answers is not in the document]
thiessenp-cds: Sounds good and is consistent with the other new name for a similar function

timarney (approved):

Issue #3843, opened 2024-06-17
Missing translation keys for objects in tree view and logic view


anikbrazeau:
## Description
Missing translation keys: some objects only show up in English even though French translation exists. This happens for sections "Start" and "End" in tree view and for  "Privacy Policy" "Review" in the logic view dropdown for sections and branching visualization.

## Steps to reproduce
1. Go to staging / conditional logic PR
2. Click on Edit
3. Toggle to French
4. See "Start" and "End" in FR
5. Go to "Logique"
6. See sections and blocks in English only even when editing in French

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Conditional logic env
- **Language**: French
- **Viewport**: Desktop

### Tree view (Default sections)

EN: `Beginning`
FR: `Début`

EN: `End`
FR: `Fin`

### Logic view (Default sections)

EN: `Introduction`
FR: `Introduction`

EN: `Privacy`
FR: `Confidentialité`

EN: `Review`
FR: `Révision`

EN: `Confirmation message`
FR: `Message de confirmation`

### Tree view

![Screenshot 2024-06-17 at 9 08 10 AM](https://github.com/cds-snc/platform-forms-client/assets/38330843/c76d9db8-98ef-44ad-b399-46659f938b75)

![Screenshot 2024-06-17 at 9 35 49 AM](https://github.com/cds-snc/platform-forms-client/assets/38330843/1d049a26-b98e-4ed9-b514-8484d5e4e4af)

### Logic view
![Screenshot 2024-06-17 at 9 08 33 AM](https://github.com/cds-snc/platform-forms-client/assets/38330843/761b2c44-a344-455f-a4d1-0174335da7d4)

![Screenshot 2024-06-17 at 9 49 08 AM](https://github.com/cds-snc/platform-forms-client/assets/38330843/405e2481-8dcb-40ef-a8c8-3b44bba04803)

PR #3861, opened 2024-06-19
chore: Conditional logic missing translations

timarney:
# Summary | Résumé

Fixes: https://github.com/cds-snc/platform-forms-client/issues/3843

### DIFF
--- a/app/(gcforms)/[locale]/(form administration)/form-builder/components/shared/right-panel/treeview/util/initializeGroups.ts
+++ b/app/(gcforms)/[locale]/(form administration)/form-builder/components/shared/right-panel/treeview/util/initializeGroups.ts
@@ -1,5 +1,10 @@
 import { FormProperties } from "@lib/types/form-types";
 import { LockedSections } from "../types";
+import {
+  getStartLabels,
+  getReviewLabels,
+  getEndLabels,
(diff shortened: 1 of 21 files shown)


app/(gcforms)/[locale]/(form administration)/form-builder/components/shared/right-panel/treeview/util/initializeGroups.ts
```diff
@@ -17,11 +18,12 @@ export const initializeGroups = (form: FormProperties, allowGroups: boolean): Fo
     // Default the forms groups with a start and an end group
     const elementIds = form.elements.map((element) => String(element.id));
 
+    // i18n
     const groups = {
       start: {
         name: "Start",
-        titleEn: "Start",
-        titleFr: "Début",
+        titleEn: getStartLabels().en,
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(form administration)/form-builder/components/shared/right-panel/treeview/util/initializeGroups.ts:25]: Nice!

thiessenp-cds (approved):
Nice work! LGTM :)

Side note, that an automated tool like Wave is complaining that the textareas on the translation page don't have labels. But I can see they do.. not sure what's up. I think there is already a bug for this so just an FYI (unrelated to this fix)

PR #3868, opened 2024-06-20
Fix the "More" Modal Pushing Data up and Down

### LINKED ISSUE #3867: More Modal Not Updating Page State w/ Groups
When using the "More" modal for Form Elements, the properties are updated in the template store, but the page state doesn't update.

This means the cypress test (in formbuilder.cy.ts):
` cy.get("#required-1-id").should("be.checked"); `
fails, because while the value was set in the template store, the UI never updates the required checkbox on the main page.

The issue was found when reviewing PR : https://github.com/cds-snc/platform-forms-client/pull/3853 (and is related, but not caused by it)

ShadeWyrm:
Fixes #3867 

Learning : 

> Fully uncontrolled — in this case your custom <Input> would take defaultChecked as a prop, and pass it to DOM <input>. The state would be inside the <Input> component itself. The prop value would only ever be used once, when the component is first rendered, and then ignored. The parent would have no way to "force" that state to become something else. If you ever need to reset the input, you'd have to mount it with a different key.

TL;DR : If you want a checkbox to change across components, change it's "key" value (eg: key="{the things id}+{checked or not}"

---

To test: 
1. Create a form.
2. Add a Question
3. Edit the Title, and the Required checkbox.
4. Note the values, and then click the "More" modal. The values should be the same. Edit them in the Modal.
5. Click save in the Modal.
6. Note the values, they should match what you changed. Edit the values again. Open the Modal, they should match again.

### DIFF
--- a/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/ElementPanel.tsx
+++ b/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/ElementPanel.tsx
@@ -58,13 +58,17 @@ export const ElementPanel = ({
   const { focusWithinProps, isWithin } = useIsWithin();
   const { refs } = useRefsContext();
 
+  const handleMoreClose = () => {
+    setChangeKey(String(new Date().getTime())); //Force a re-render
+  };
+
   const moreButton =
     item.type !== "richText"
       ? {
           moreButtonRenderer: (
             moreButton: JSX.Element | undefined
           ): React.ReactElement | string | undefined => (
-            <MoreModal item={item} moreButton={moreButton} />
+            <MoreModal item={item} moreButton={moreButton} onClose={handleMoreClose} />
           ),
         }
       : {};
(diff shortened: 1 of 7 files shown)


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/ElementPanel.tsx
```diff
@@ -103,7 +107,7 @@ export const ElementPanel = ({
         }
       }}
     >
-      <PanelBodyRoot item={item} />
+      <PanelBodyRoot item={item} key={"panel-body-" + item.id} onChangeMade={forceRefresh} />
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/ElementPanel.tsx:110]: I don't think we need the key here do we?

Comment (ShadeWyrm) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/ElementPanel.tsx]: Tested, and nope! It was a left over from me testing out the push of data across components. Removed it. :)

timarney (approved):

Issue #3788, opened 2024-06-11
Sections: Reordering within the tree-view did not update in the Preview/Test page 


samsadasivan:
"I created the section "Coordonnées" for names after I created "Nourriture" and "Activités" sections. I moved it to the top in the Question panel on the right, but when I went to test, it didn't put it first in the form flow.
Because you can reorder the sections, I thought it would also reorder the flow." 

[Participant 5 (French flow) - Form file in slack](https://gcdigital.slack.com/archives/C0676D34GPR/p1718118471278139)

### Steps to Recreate:
1. Add a couple sections
2. Reorder them in Tree View
3. Click Test and see Error: it still reflects the format the sections were originally added in

### Screenshot/Video:

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNTVyQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--874666edc903800cdfae08752a7ca9004d8a1922/Screen%20Recording%202024-06-17%20at%208.38.20%E2%80%AFAM.mov)

PR #3886, opened 2024-06-24
chore: Add temporary fix for AutoFlow issues

timarney:
# Summary | Résumé

Fixes: https://github.com/cds-snc/platform-forms-client/issues/3788

Adds a hook for AutoFlow all --- this is a temporary fix until we have a better solution.


thiessenp-cds (approved):
LGTM :)

Issue #3793, opened 2024-06-11
Dynamic fields: "Other" question field isn't showing up, so clicking the link "Other" takes you no where


samsadasivan:
"For the restrictions alimentaires, tried to add "Other" for Yes, but nothing came up automatically. Clicking "Other" on the question brings nothing. I thought it added an “Other” question automatically. "

[Participant 5 (French flow) - Form file in slack](https://gcdigital.slack.com/archives/C0676D34GPR/p1718118471278139)

anikbrazeau:
## Description
Rule seemingly gets added to first option, but input field does not show up.

## Steps to reproduce
1. Go to Staging > Edit
2. Add a radio button form element
3. Add three options
4. Add Other as a fourth option
5. Add an input field for the fourth option
6. See how it shows up for first option
7. Notice how there isn't an input question that shows up

## Screenshots or videos

<img width="831" alt="Screenshot 2024-06-24 at 10 18 18 AM" src="https://github.com/cds-snc/platform-forms-client/assets/38330843/2ff51f33-b599-4eb1-9b97-05b92e1c5d63">

PR #3887, opened 2024-06-24
fix: "Other" question field isn't showing up, so clicking the link "Other" takes you no where 

timarney:
# Summary | Résumé

Fixes: https://github.com/cds-snc/platform-forms-client/issues/3793

Fixes issue for Other field when form is "grouped".

When using "groups" groupID was not passed so the field was not being added to the group.


thiessenp-cds (approved):
LGTM :)

PR #3912, opened 2024-06-27
chore: Updates department list from gc-organizations

thiessenp-cds:
# Summary | Résumé

Updates department list from gc-organizations.

Test by making sure a new org is in the department dropdown, for example:
EN: Office of the Secretary to the Governor General
FR: Bureau du secrétaire du gouverneur général


timarney:
For reference do we have an issue and or info on where the updated info came from?

PR #3922, opened 2024-07-02
fix: Form Purpose not saving to DB

ShadeWyrm:
# Summary | Résumé

There was missing functionality for moving formPurpose from the template store to the DB.

This adds the missing stuff!~


lib/types/form-types.ts
```diff
@@ -142,6 +142,7 @@ export interface DeliveryOption {
 }
 
 export type SecurityAttribute = "Unclassified" | "Protected A" | "Protected B";
+export type FormPurpose = "" | "admin" | "nonAdmin";
```
timarney, line 145: Can we update or replace the current enum ?

```
export enum PurposeOption {
  none = "",
  admin = "admin",
  nonAdmin = "nonAdmin",
}
```

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/components/ResponseDelivery.tsx
```diff
@@ -245,6 +246,15 @@ export const ResponseDelivery = () => {
     }
 
     updateField("formPurpose", purposeOption);
+    result = (await updateTemplateFormPurpose({
+      id,
+      formPurpose: formPurpose,
+    })) as FormServerError;
+
+    if (result?.error) {
+      toast.error(<ErrorSaving errorCode={FormServerErrorCodes.DELIVERY_OPTION} />, "wide");
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/components/ResponseDelivery.tsx:255]: This error should match form purpose vs `DELIVERY_OPTION`

timarney:
Small lint error to fix up

Screenshot 2024-07-02 at 2 47 54 PM

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/components/ResponseDelivery.tsx
Comment (ShadeWyrm) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/components/ResponseDelivery.tsx:255]: For FormServerErrorCodes - do we have an established pattern for these? Should I just increment to 553?

Comment (timarney) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/components/ResponseDelivery.tsx]: yes for now just increment we haven't established patterns for these yet

lib/types/form-types.ts
timarney: This is okay as just minor concerns around having multiple types for the PurposeOption and needing to update in multiple places (possibly missing one or the other) if the types ever change.

 

We could use as const vs the enum:

Screenshot 2024-07-03 at 8 10 10 AM


https://youtu.be/jjMbPt_H3RQ?t=320

---

We can leave as is for now.

timarney (approved):
Approving -- to move this forward with getting the value to save.

As a future PR we can re-look at how the settings are saved across the multiple save buttons whether or not they should be independent of each other.

PR #3924, opened 2024-07-02
fix: nested check for response 

timarney:
# Summary | Résumé

Fixes small issues with a nested !response found while reviewing some other code.

Prior to PR:

```javascript
if (!response)
          if (!response)
```


bryan-robitaille (approved):
Thanks for catching

PR #3934, opened 2024-07-03
feat: off-board / exit sections

timarney:
# Summary | Résumé

Adds ability to mark section as an exit page

https://www.figma.com/design/bIcbQm61JbkCy9Nu8XE7u6/Conditional-logic-Alpha?node-id=4234-18134&m=dev

**Set exit action:**
Screenshot 2024-07-08 at 11 52 06 AM

**Right panel - Remove action panel:**
Screenshot 2024-07-08 at 11 52 35 AM

**Custom exit node for logic display:**
Screenshot 2024-07-08 at 11 55 31 AM

**Preview**
Screenshot 2024-07-08 at 11 55 47 AM

**Editor with custom link:**
Screenshot 2024-07-08 at 2 01 23 PM


app/(gcforms)/[locale]/(form administration)/form-builder/components/shared/right-panel/logic/GroupSelect.tsx
```diff
@@ -23,15 +23,18 @@ export const GroupSelect = ({
   return (
     <div className="mb-4">
       <label className="mb-2 block text-sm" id={labelId}>
+        {selected === "exit" && <span className="sr-only">{t("disabled")}</span>}{" "}
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(form administration)/form-builder/components/shared/right-panel/logic/GroupSelect.tsx:26]: Not an ideal solution but it should work. Why this solution?
-disabling a form field does not announce this to AT
-using our live region util would be awkward because the DOM change already announces the label
-adding the "disabled" text at the end of the label announces awkwardly in between the label update and the dropdown update
-this left putting the "disabled" text at the beginning of the label only "visible" to AT

thiessenp-cds (approved):
Works great! LGTM :)

PR #3935, opened 2024-07-03
chore: Move refs provider

timarney:
# Summary | Résumé

Moves refs provider outside of edit so we can make use of it from elements such as the TreeView.


thiessenp-cds (approved):
Works as expected. LGTM :)

PR #3936, opened 2024-07-03
chore: increase next server maximum memory

craigzour:
# Summary | Résumé

- Increases maximum allowed memory space for our Next server. Following recommendation from [Node documentation](https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes) I set it to 3.5 GB because our ECS is configured with 4 GB of memory.

After investigating on a recent Out of Memory issue in production I could not find any root cause that would explain it. Since it was the first time it happened I decided to just go with a simple change that was recommended on different online websites.

For reference, here is the Fatal error we got in production:
`FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`


patheard (approved):
Makes sense and we could always bump the ECS task's memory if we continue to see more of these.

bryan-robitaille (approved):

Issue #3933, opened 2024-07-03
Overlapping sections in Map view


connorscarolyns:
Below are my notes re: the glitched form. If you’d like additional details, let me know.
 
June 28, 2024
Notes for the Form that appears to be glitched. I will send my other notes from this session separately (i.e., not related to the glitched form).
Browser: Microsoft Edge
Windows 10 Enterprise
Json file attached

Issue summary:
I don’t remember exactly what I was doing when I noticed this, but I believe I was just working on adding in our own privacy statement so that I could see how the start of the form would look with the full statement.  
Sections "End" and "Beginning"  just appeared in the order as shown in screenshot below and I couldn't drag and drop to rearrange the sections back to how they were originally. I can drag and drop the sections I created but I can’t do anything with the locked sections. This prevented me from being able to set up a proper logic flows so I opted to just create a new form and start from scratch. Screenshot included.

<img width="457" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK2F1QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--00d748ae6bd490bfb67e0133f38643a83b0fbc47/image.png" alt="image.png" />

In the Logic tab, the logic flow tables are merged together and couldn't find a way to reassemble them because I can’t move individual tables within the logic section, just the entire workflow as a whole. I tried resetting the linear logic but that didn't help. Screenshot included.

<img width="974" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK2V1QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a7a05d02b07c46aa88b53c125456d00d52a76c4c/image.png" alt="image.png" />

connorscarolyns:
[psohp-sr-form-glitched-2024-07-02-json.txt](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK3l1QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--68cab2d682f538734b2553c6c605c9d2b9a4e655/psohp-sr-form-glitched-2024-07-02-json.txt)

timarney:
[...]

PR #3953, opened 2024-07-04
fix: Reset locked groups

timarney:
# Summary | Résumé

Adds reset function for when locked sections get out of order

see: https://github.com/cds-snc/platform-forms-client/issues/3933

https://github.com/cds-snc/platform-forms-client/assets/62242/32764fa9-8a46-4c03-9ff1-3564f299fa6b


thiessenp-cds (approved):
I tested in this environment to see the reset work and also in my dev environment to see the old behaviour (reset not work). Works great! Also the reference to the bug ticket in the new function is very helpful. Nice work! LGTM :)

PR #3956, opened 2024-07-04
fix: Add key for next action

timarney:
# Summary | Résumé

Fixes issue where the next action selector was ending up using a previously selected value from stale state.


thiessenp-cds (approved):
Works as expected. LGTM :)

PR #3972, opened 2024-07-08
refactor: More explcity privilege checks

bryan-robitaille:
# Summary | Résumé

Add more explicit privilege checks where `checkPrivilege` functions are used.

We often don't pass in the `object: {}` which ensures that only users that have global permissions can access that particular function.  The base function now treats all checks that don't include an object as a global permissions check.

App should work as normal.

This does not include going through each `lib` to ensure that all CRUD functions leverage proper ability checks.

# Pull Request Checklist

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:20@sha256:ab71b9da5ba19445dc5bb76bf99c218941db2c4d70ff4de4e0d9ec90920bfe3f
+FROM node:20@sha256:93d2e801dabc677ea0b30b47d3d729fab63ecb20be7ac0ab204cc3c65731297a
 
 ARG USERNAME=node
 ARG USER_UID=1000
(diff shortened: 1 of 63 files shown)


craigzour (dismissed):
LGTM!

timarney:
Couple of notes for the future 

- Passing {} called object is confusing :)
- It would be nice if we could default the {} so we don't have to pass it ... outside of the comment you can't just look at the code and understand it for "all"

bryan-robitaille:
> Couple of notes for the future
> 
>     * Passing {} called object is confusing :)
> 
>     * It would be nice if we could default the {} so we don't have to pass it ... outside of the comment you can't just look at the code and understand it's for "all"

That's a great suggestion. We'll have to rework the whole permission set when we start exploring teams later this year but for now I've modified the privilege check to check the global permission if no specific object is passed in.

timarney (approved):
🥇

Issue #3884, opened 2024-06-24
Question menu for moving/duplicating/deleting cuts off


anikbrazeau:
## Description
Panel cuts off question menu when window is intermediate size between full desktop and tablet view.

This was noticed in diary study session observation with client using a different tablet:

![cut](https://github.com/cds-snc/platform-forms-client/assets/38330843/144bceea-3c37-4daf-ad3d-0ae1bba01cee)

## Steps to reproduce
1. Go to staging > Edit
2. Toggle to French
3. Adjust the window size
4. See how the menu gets cut off 

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Staging
- **Language**: French
- **Viewport**: Desktop, partial width

## Expected behaviour
Responsive menu that does not cut off text?

## Screenshots or videos
<img width="1436" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeHlJQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4573e6f511ab2a9d8811c172c0cadc6682c61048/Screenshot%202024-06-24%20at%208.56.23%E2%80%AFAM.png" alt="Screenshot 2024-06-24 at 8.56.23 AM.png" />

Abi-Nada:
Action: Modify the z-index to a higher level than the right panel

PR #3981, opened 2024-07-10
fix:  action panel z-index

timarney:
# Summary | Résumé

Fixes: https://github.com/cds-snc/platform-forms-client/issues/3884

Fixes stacking order for panel actions and right menu

Screenshot 2024-07-10 at 11 54 01 AM


thiessenp-cds (approved):
LGTM :)

Issue #3918, opened 2024-06-28
Logic view: If the panel closed, when you click on an arrow, it should open the panel 


samsadasivan:
When you select one of the arrows and the panel is closed, it should open up to show what they clicked on.

timarney:
PR: 

https://github.com/cds-snc/platform-forms-client/pull/3984

PR #3984, opened 2024-07-10
fix: add toggle right panel from logic view nodes

timarney:
# Summary | Résumé

fixes: https://github.com/cds-snc/platform-forms-client/issues/3918

Fixes issue where the right panel should open when clicking to add a rule from a group node.

https://github.com/cds-snc/platform-forms-client/assets/62242/247827da-2468-4794-bc9f-08c74c2a8130


thiessenp-cds (approved):
Nice work. LGTM :)

For a follow up PR it looks like we could do some work on focus management. Like when closing the right panel, thinking where the focus should go. Also, when the right panel slides into view, this is not announced to a screen reader. (should be fixed at some point)

PR #3989, opened 2024-07-11
chore: use section name for logic view

timarney:
# Summary | Résumé

Update to use name vs section title 

Screenshot 2024-07-11 at 9 27 29 AM


thiessenp-cds (approved):
In the standup we decided on name (not bilingual). LGTM :)

PR #3995, opened 2024-07-12
chore: Safe JSON parsing utility

thiessenp-cds:
# Summary | Résumé

This PR simplifies the safeJSONParse() util and fixes a TS error.
What was simplified? Returning undefined instead of an object. The reason for this is that JSON.parse can only thrown one type of error. Returning undefined or an object with a string is pretty much the same.  So undefined is simpler and leaves the caller of safeJSONParse() to determine how to handle an error (vs. trying to send a generic error message or something like that).

## Testing

Test by trying some some common workflow patterns. Like signing on, creating a form, publishing a form..


app/(gcforms)/[locale]/(user authentication)/auth/mfa/resend/components/client/ReVerify.tsx
```diff
@@ -28,22 +28,28 @@ export const ReVerify = (): ReactElement => {
...
     async function handleError() {
       logMessage.error("Failed to parse authFlowToken JSON.");
       const errorText = await getErrorText(language, "InternalServiceException");
       setAuthErrorState(errorText);
     }
-    existingFlow.current = safeJSONParse(sessionStorage.getItem("authFlowToken") || "{}");
-    if (!existingFlow.current.email || !existingFlow.current.authenticationFlowToken) {
-      router.push(`/${language}/auth/login`);
-    }
+
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(user authentication)/auth/mfa/resend/components/client/ReVerify.tsx:42]: Assigning undefined to a ref will result in an error. So I added an optional error object to handle the error case. There may be a better way to do this.

app/(gcforms)/[locale]/(form filler)/id/[...props]/lib/rehydrateFormResponses.ts
```diff
@@ -74,9 +74,6 @@ function _rehydrateDynamicRowResponses(responses: [string, Response][]) {
 }
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(form filler)/id/[...props]/lib/rehydrateFormResponses.ts:75]: Unrelated but it looks like we call JSON.parse() when checking every checkbox. Looks like some performance optimizations could be done here. (Future PR)

app/(gcforms)/[locale]/(form filler)/id/[...props]/lib/rehydrateFormResponses.ts
```diff
@@ -74,9 +74,6 @@ function _rehydrateDynamicRowResponses(responses: [string, Response][]) {
...
 
 function _rehydrateCheckBoxResponse(response: Response) {
-  const responseParsed = safeJSONParse(response as string);
-  if (responseParsed?.error || !responseParsed?.value) {
-    return [];
-  }
-  return responseParsed.value;
+  const responseParsed = safeJSONParse(response as string) as { value: string[] } | undefined;
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(form filler)/id/[...props]/lib/rehydrateFormResponses.ts:77]: Instead of typing the return manually and adding the `undefined` as a possible return value we can also just pass in the expected return type to the function itself.  It will infer, based on the generics that were used in the `safeJSONParse` delcaration the return type and include `undefined` automatically.
```suggestion
const responseParsed = safeJSONParse<{ value: string[] }>(response as string);
```

app/(gcforms)/[locale]/(form administration)/form-builder/Start.tsx
```diff
@@ -42,21 +43,21 @@ export const Start = () => {
       fileReader.onload = (e) => {
         if (!e.target || !e.target.result || typeof e.target.result !== "string") return;
         const data = safeJSONParse(e.target.result, cleaner);
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(form administration)/form-builder/Start.tsx:45]: We should type `data` at the source.


```suggestion
        const data = safeJSONParse<FormRecord> (e.target.result, cleaner);
```

app/(gcforms)/[locale]/(form administration)/form-builder/Start.tsx
```diff
@@ -42,21 +43,21 @@ export const Start = () => {
...
-        if (data?.error) {
+        if (!data) {
           setErrors([{ message: t("startErrorParse") }]);
           target.value = "";
           return;
         }
 
-        const validationResult = validateTemplate(data);
+        const validationResult = validateTemplate(data as FormRecord);
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(form administration)/form-builder/Start.tsx:52]:
```suggestion
        const validationResult = validateTemplate(data);
```

app/(gcforms)/[locale]/(form administration)/form-builder/Start.tsx
```diff
@@ -42,21 +43,21 @@ export const Start = () => {
...
 
         if (!validationResult.valid) {
           setErrors(validationResult.errors);
           target.value = "";
           return;
         }
 
-        importTemplate(data);
+        importTemplate(data as FormProperties);
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(form administration)/form-builder/Start.tsx:60]: This seems to be wrongly typed and should be: 
```suggestion
        importTemplate(data.form);
```

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/Preview.tsx
```diff
@@ -44,13 +44,13 @@ export const Preview = ({
 
   // TODO probably redirecting to the error page makes more sense since the error is not recoverable
   const formParsed = safeJSONParse(getSchema());
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/Preview.tsx:46]:
```suggestion
  const formParsed = safeJSONParse<FormProperties>(getSchema());
```

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/Preview.tsx
```diff
@@ -44,13 +44,13 @@ export const Preview = ({
...
-  if (formParsed?.error) {
+  if (!formParsed) {
     toast.error(<ErrorSaving errorCode={FormServerErrorCodes.JSON_PARSE} />, "wide");
   }
 
   const formRecord: PublicFormRecord = {
     id: id || "test0form00000id000asdf11",
-    form: formParsed,
+    form: formParsed as FormProperties,
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/Preview.tsx:53]:
```suggestion
    form: formParsed
```

bryan-robitaille (changes requested):
A few quick changes to align on a common pattern and I think we're good to go!

lib/utils/form-builder/validate.ts
```diff
@@ -22,7 +22,7 @@ const getErrorMessageTranslationString = (error: ValidationError) => {
   };
 };
 
-export const validateTemplate = (data: FormRecord) => {
```
thiessenp-cds, line 25: In Preview the type used for an uploaded JSON file is FormProperties. I'm pretty sure that's what was intended here.

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/Preview.tsx
```diff
@@ -42,15 +43,14 @@ export const Preview = ({
     getSecurityAttribute: s.getSecurityAttribute,
   }));
 
-  // TODO probably redirecting to the error page makes more sense since the error is not recoverable
-  const formParsed = safeJSONParse(getSchema());
-  if (formParsed?.error) {
+  const formParsed = safeJSONParse<FormProperties>(getSchema());
+  if (!formParsed) {
     toast.error(<ErrorSaving errorCode={FormServerErrorCodes.JSON_PARSE} />, "wide");
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/Preview.tsx:48]: How to handle this unlikely error case needs more thought.
The problem is the error message shown will mention a JSON error and give the option to download the file. But the file will be the default created form and be valid JSON. (hiding the issue of invalid JSON)

To resolve this the component code probably needs to be refactored to handle an error case like this better.

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/Preview.tsx
```diff
@@ -42,15 +43,14 @@ export const Preview = ({
...
   }
 
   const formRecord: PublicFormRecord = {
     id: id || "test0form00000id000asdf11",
-    form: formParsed,
+    form: formParsed || defaultForm, // defaultForm to stop a redirect to the app error page
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/Preview.tsx:53]: Or perhaps it would be better to remove `defaultForm` and let an error be thrown and the app redirect to the error page. This way the user sees a correct error, not very helpful but correct. Versus handling the error but showing an unhelpful error.

bryan-robitaille (approved):

PR #4005, opened 2024-07-16
chore: use add button for add rule

timarney:
# Summary | Résumé

Updates code to use a button vs a span for the add rule "link"


thiessenp-cds (approved):
Nice, works as expected. LGTM :)

Issue #4007, opened 2024-07-16
Importing JSON form files doesn't parse body of form


Abi-Nada:
## Description
When downloading JSON form files from Production and opening them in Staging, the body of the form doesn't appear as expected.

**What appears:** Form title, Form description, and Confirmation Page, Privacy Notice
**What's missing:** Form questions

## Steps to reproduce
1. Go to https://articles.alpha.canada.ca/forms-formulaires/templates/ and download a JSON form template
2. Go to Staging and click "Open a form file"
3. Open the JSON
4. **See error**: Title, Description, Privacy Notice and Confirmation Page appear, but the form questions do not.

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Staging

## Expected behaviour
The entire form (aside from the confirmation page?) should appear under the "Start Page".

## Screenshots or videos
Showing what a downloaded JSON looks like in Production vs Staging:

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK3JsQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f523e909b7765f67a68bd7a8bbb3be72d677f34a/Screen%20Recording%202024-07-16%20at%206.56.03%E2%80%AFPM.mov)

Abi-Nada:
@Abi-Nada Test JSON in Staging for:
- Repeating Set form
- File upload form

PR #4008, opened 2024-07-17
fix: Import with groups

timarney:
# Summary | Résumé

See: https://github.com/cds-snc/platform-forms-client/issues/4007

Adjusts initialization code for groups when a form is imported without a groups layout property. 

### Testing 
Import the attached form which has no groups or group layout with the grouping feature on.
[my-form.json](https://github.com/user-attachments/files/16264914/my-form.json)

Note that the elements should end up under the Start page.


thiessenp-cds (approved):
LGTM :)

PR #4009, opened 2024-07-17
chore: set message property on custom errors

craigzour:
# Summary | Résumé

- Applied Typescript recommendation on how to write custom error objects (https://github.com/jestjs/jest/issues/8279#issuecomment-539775425) 
- Fixed broken unit test which got revealed after applying Typescript recommendation on custom error objects. The `expect` operator was not able to properly compare received vs expected error type
- Replaced some of the Jest matchers functions that were deprecated
- Set message property (with error name) when creating custom error objects

Reason:
When our logger (Pino) logs an error object it uses the name of the error and put it in the `err.type` property. The issue is that NextJS will rename error objects during compilation time. It makes it difficult for us to know which error was thrown in production. However the `err.message` property is not altered so we can use it to convey the error name.

Before
```
{
    "level": "warn",
    "time": 1721145061802,
    "err": {
        "type": "a",
        "message": "",
        "stack": "Error\n at f (/src/.next/server/chunks/3444.js:13:933)\n at async..."
    },
    "msg": ""
}
```

After
```
{
    "level": "warn",
    "time": 1721145061802,
    "err": {
        "type": "a",
        "message": "MyErrorName",
        "stack": "Error\n at f (/src/.next/server/chunks/3444.js:13:933)\n at async..."
    },
    "msg": ""
}
```


timarney (approved):

wmoussa-gc (commented):
nice workaround, I would not have thought about it

Issue #4018, opened 2024-07-17
Missing translation: "Start" in English in mapping view of logic in French


anikbrazeau:
## Description
Translate the name of the default page "Start" in French --> "Départ"

## Steps to reproduce
1. Go to Staging
2. Toggle to French
3. Open panel "Configuration du formulaire"
4. See the name of the start page in French which shows up in English

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Staging
- **Language**: French
- **Viewport**: Desktop

## Expected behaviour

Shows up as expected in Tree view and in dropdown but not in the visualization on the left. 

EN = Start
FR = Départ

## Screenshots or videos

<img width="1539" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMW5yQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--baec713ace6ef22b5f9f4c80a3e9b3d7073fddb2/Screenshot%202024-07-17%20at%204.06.21%E2%80%AFPM.png" alt="Screenshot 2024-07-17 at 4.06.21 PM.png" />

PR #4029, opened 2024-07-18
fix: update start page name for logic view

timarney:
# Summary | Résumé

Fixes issue where start page wasn't receiving the updated translation name.

See: https://github.com/cds-snc/platform-forms-client/issues/4018

Noting the start group is unique as it's name gets setup when the template is created given we need a place for store existing form elements.

In this case we can replace the name when we're creating nodes for the logic view.

Screenshot 2024-07-18 at 1 09 22 PM

Screenshot 2024-07-18 at 1 09 08 PM


anikbrazeau (approved):
LGTM! Thanks Tim!

PR #4034, opened 2024-07-19
refactor: use server side check privilege check for page level

bryan-robitaille:
# Summary | Résumé

Just a quick change to align all server side privilege checks to use the `checkPrivileges` functions instead of using `ability` directly.


craigzour (approved):
Perfect! This was something I added to my TODO list yesterday ^^

craigzour (approved):

app/(gcforms)/[locale]/(app administration)/admin/(with nav)/accounts/[id]/manage-permissions/actions.ts
```diff
@@ -10,7 +10,7 @@ export const updatePrivileges = async (
   action: "add" | "remove"
 ) => {
   const { ability } = await authCheckAndThrow();
```
Comment (wmoussa-gc) [inline app/(gcforms)/[locale]/(app administration)/admin/(with nav)/accounts/[id]/manage-permissions/actions.ts:12]: do we still need the `ability` object?

Issue #4027, opened 2024-07-18
Start Page doesn't open on-click after navigating back from another page


Abi-Nada:
## Description
Start Page doesn't open on-click after navigating back from another page

## Steps to reproduce
1. Go to a form with multiple pages
2. Click on a random page in the tree view (**not** Start page)
3. Click on Test
4. Navigate back to Edit
5. Click on Start page

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Staging
- **Language**: English

## Expected behaviour
Start Page should open on-click after navigating back from another page

## Screenshots or videos

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOEx2QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e6560a3ab2cf26daf36fc5c4c19301a80dde7d4f/Screen%20Recording%202024-07-18%20at%2011.18.45%E2%80%AFAM.mov)

Abi-Nada:
Waiting for PR to be approved.

Tested using the PR environment and issue no longer persists.

PR #4037, opened 2024-07-19
fix: adds additional check / set for active group 

timarney:
# Summary | Résumé

This fixes and issue where the ` onFocusItem` event does not trigger.

This adds an additional check under `onExpandItem` to ensure the correct active group is set. 

Fix: https://github.com/cds-snc/platform-forms-client/issues/4027


thiessenp-cds (approved):
LGTM :)

PR #3619, opened 2024-05-14
chore: DynamoDB retry with exponential backoff

wmoussa-gc:
# Summary | Résumé

We've experienced DynamoDB beginning to throttle the response confirmation requests. It returned a ThrottlingException to the caller.

```
{"level":"error","time":1713791577367,"err":{"type":"DynamoDBServiceException","message":"Throughput exceeds the current capacity of your table or index. DynamoDB is automatically scaling your table or index so please try again shortly. If exceptions persist, check if you have a hot key: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/bp-partition-key-design.html","stack":"ThrottlingException: Throughput exceeds the current capacity of your table or index. DynamoDB is automatically scaling your table or index so please try again shortly. If exceptions persist, check if you have a hot key: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/bp-partition-key-design.html
```

This PR suggests a simple exponential backoff method. Exponential backoff is a retry strategy that gradually increases the time between retries, typically doubling the waiting time after each failed attempt, to mitigate congestion and reduce the likelihood of overwhelming a system with retry requests.

# Test instructions | Instructions pour tester la modification

- It's not straightforward to test. One option is to load test the confirm API by overwhelming it, or configure LocalStack DynamoDB to randomly throw exceptions (see https://docs.localstack.cloud/references/configuration/#dynamodb).

# Pull Request Checklist

Please complete the following items in the checklist before you request a review:

- [ ] Have you completely tested the functionality of change introduced in this PR? Is the PR solving the problem it's meant to solve within the scope of the related issue?
- [x] The PR does not introduce any new issues such as failed tests, console warnings or new bugs.
- [x] If this PR adds a package have you ensured its licensed correctly and does not add additional security issues?
- [ ] Is the code clean, readable and maintainable? Is it easy to understand and comprehend.
- [ ] Does your code have adequate comprehensible comments? Do new functions have [docstrings](https://tsdoc.org/)?
- [ ] Have you modified the change log and updated any relevant documentation?
- [ ] Is there adequate test coverage? Both unit tests and end-to-end tests where applicable?
- [x] If your PR is touching any UI is it accessible? Have you tested it with a screen reader? Have you tested it with automated testing tools such as axe?

PR #4042, opened 2024-07-19
refactor: update confirm api to a server action

bryan-robitaille:
# Summary | Résumé

This PR refactors the confirm API to a server action as well implements database throttling to handle larger read / write loads.

Instead of trying to implement exponential back off and retries manually this PR instead uses the built in retry services that are included by default in the AWS SDK.  The number of retries for the dynamodb service has been increased from 3 to 15.  In LocalStack with the dynamodb set to return throttle capacity exceptions 50% of the time we can see the AWS retry logic at work.

As an additional throttling mechanism a 200ms delay is introduced between batched calls to more evenly spread out the load over the partition key (form ID)

| BatchGetCommand | TransactWriteItems Command | Query Command | 
| ---------- | ------------- | ------------- |
| Screenshot 2024-07-19 at 8 06 45 AM | Screenshot 2024-07-19 at 7 55 50 AM | Screenshot 2024-07-18 at 12 32 50 PM |

Setup:
- Publish a Form
- Generate a ton, and I mean of ton of responses for it
- Set the response download limit to something ridiculous like 1000

Test
- Download 1000 responses at once
- Confirm 1000 responses at once

The actions should not throw errors and should complete, albeit a little more slowly, for a user.
If there are dynamodb warnings thrown in Slack it should have no impact on the user.

Still need to refactor "report a problem" and "unprocessed submissions" APIs to server actions.

# Pull Request Checklist


thiessenp-cds:
I tested this and from what I can see it works, mostly.
I tested by 
1. creating a new form and adding 1000 submissions
2. Then trying to download 1000 submissions on staging develop. That threw a bunch of errors and seemed to fail.
3. Using this PRs branch, download 1000 submissions. Did not throw any errors. The download was fast and successful, as well marking the submissions as downloaded.
BUT the form response page seemed to have inconsistent state. Downloads appeared to have their status reverted and need to be downloaded again.

lib/utils/retryability.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Returns the number of milliseconds to be used in an exponential backoff
+ * @param  base The base number of milliseconds to wait
+ * @param  attempt The number of times the operation has been attempted
+ * @param  max The maximum amount of milliseconds to wait
+ * @param  withJitter Set to true to add jitter (randomness) to the exponential backoff time that is returned
+ * @returns The number of milliseconds to be used in an exponential backoff
+ */
+export const getExponentialBackoffTimeInMS = (
```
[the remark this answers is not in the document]
craigzour: This is something I would also like to discuss with Bryan when he gets back because it could just be left over code from a first iteration that was not removed when he decided to use the default retry mechanism that the DynamoDB SDK offers.

wmoussa-gc:
I feel bad for not knowing why my original fix didn't work. https://github.com/cds-snc/platform-forms-client/pull/3619

lib/utils/retryability.ts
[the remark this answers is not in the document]
wmoussa-gc: Last time I worked in that area, the `ThrottlingException` was not a retryable exception.

lib/vault.ts
```diff
@@ -408,11 +422,15 @@ export async function updateLastDownloadedBy(
...
+    if (index > 0) {
+      // eslint-disable-next-line no-await-in-loop
+      await delay(200);
+    }
 
-  // @todo - handle errors that can result from failing TransactWriteCommand operations
-  return runPromisesSynchronously(asyncUpdateRequests);
+    index++;
+
+    return dynamoDBDocumentClient.send(request);
```
craigzour, line 432: I will modify this as it should be an `await` instead of a `return`

craigzour:
Works pretty well! Good job @bryan-robitaille :)

If we decide to move forward with the increased response limit of 1000 I wonder if we should not also be adding delay in the loops for the following functions: `listAllSubmissions`, `retrieveSubmissions`.

Also just want to know whether the implementation of `getExponentialBackoffTimeInMS` in the `retrieveSubmissions` function is intentional or not because from my understanding the DynamoDB SDK would still do it for us in case we get `UnprocessedKeys` to reprocess.

lib/vault.ts
craigzour: Pushed a fix for that

lib/utils/retryability.ts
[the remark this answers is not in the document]
bryan-robitaille: When using the AWS SDK `Throttling Exception` is retried. https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html#Programming.Errors.RetryAndBackoff

bryan-robitaille:
> Works pretty well! Good job @bryan-robitaille :)
> 
> If we decide to move forward with the increased response limit of 1000 I wonder if we should not also be adding delay in the loops for the following functions: `listAllSubmissions`, `retrieveSubmissions`. However this could also have an impact on the user experience if form owners downloading forms start to refresh the page thinking that the process takes more time than usual (I did experiment it while testing this PR). Since we wait for the whole process to be completed to return the responses then if you refresh just before the end you will end up having responses that have been moved to the downloaded tab even though you did not even get them once.
> 
> Also just want to know whether the implementation of `getExponentialBackoffTimeInMS` in the `retrieveSubmissions` function is intentional or not because from my understanding the DynamoDB SDK would still do it for us in case we get `UnprocessedKeys` to reprocess.

@craigzour the 1000 response limit was only for testing that this does increase the performance based on our current baseline.  There is still much work to do before we can get to the point of setting 1000 as the default limit but I considered it out of scope of the current PR.

The use of exponential backoff in `retrieveSubmissions` is intentional because 'UnprocessedKeys' are not retried automatically because the 'Batch Commands' do not return a 4xx or 5xx error.  https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html#Programming.Errors.BatchOperations

bryan-robitaille:
> I feel bad for not knowing why my original fix didn't work. #3619

Don't fee bad!  It did work but only covered off confirming responses.  It was also redundant of the logic that TransactWriteItems was already doing behind the scenes to retry.  So it was kinda like retry logic on top of retry logic.

lib/vault.ts
craigzour: True! I will revert that specific change

craigzour: Oh I guess you already did ;)

bryan-robitaille:
> Also just want to know whether the implementation of `getExponentialBackoffTimeInMS` in the `retrieveSubmissions` function is intentional or not because from my understanding the DynamoDB SDK would still do it for us in case we get `UnprocessedKeys` to reprocess.

@craigzour sorry I misunderstood your comment.  I think I understand now what you were referencing.  Basically, why do we implement a delay if there are 'UnprocessedKeys' because in the next iteration of the `BatchGetCommand` in the while loop the AWS SDK will already implement retries and throttling for the unretrieved keys.

So my thought process with introducing the delay is as follows:
_Context_ : We pass in an array of 50 items to the BatchGetCommand

**Without the delay**
1. The BatchGetCommand fails and retries after 200ms, the second retry succeeds and returns 10 items with 40 in 'Unprocessed Keys.
2. 40 items are passed back in and are immediately retried and fails.  Internally the second retry also fails and delays, the 3rd retry succeeds and returns 10 items with 30 in 'UnprocessedKeys'

My conclusion was that it was kind of a rapid fire approach where we knew the dynamoDB was busy but we immediately retried anyway.  The subsequent invocation of BatchGetCommand runs immediately because the SDK retry mechanism starts fresh every time.

**With the delay**
1. The BatchGetCommand fails and retries after 200ms, the second retry succeeds and returns 10 items with 40 in 'Unprocessed Keys.
2. We have an indicator that the DB is busy so we delay 200ms externally before calling the BatchGetCommand again for the same data.
3. 40 items are passed back in and are retried and fail.  Internally the second retry also fails and delays, the 3rd retry succeeds and returns 20 items with 20 in 'UnprocessedKeys'
4. We have an indicatore that the DB is really busy so we delay 400ms externally before calling the BatchGetCommand again.

With this version I thought we were being proactive by giving the dynamoDB a chance to catch up by reducing the load on it and giving it a better probability of completing the request with all items.

craigzour:
> > Also just want to know whether the implementation of `getExponentialBackoffTimeInMS` in the `retrieveSubmissions` function is intentional or not because from my understanding the DynamoDB SDK would still do it for us in case we get `UnprocessedKeys` to reprocess.
> 
> @craigzour sorry I misunderstood your comment. I think I understand now what you were referencing. Basically, why do we implement a delay if there are 'UnprocessedKeys' because in the next iteration of the `BatchGetCommand` in the while loop the AWS SDK will already implement retries and throttling for the unretrieved keys.
> 
> So my thought process with introducing the delay is as follows: _Context_ : We pass in an array of 50 items to the BatchGetCommand
> 
> **Without the delay**
> 
> 1. The BatchGetCommand fails and retries after 200ms, the second retry succeeds and returns 10 items with 40 in 'Unprocessed Keys.
> 2. 40 items are passed back in and are immediately retried and fails.  Internally the second retry also fails and delays, the 3rd retry succeeds and returns 10 items with 30 in 'UnprocessedKeys'
> 
> My conclusion was that it was kind of a rapid fire approach where we knew the dynamoDB was busy but we immediately retried anyway. The subsequent invocation of BatchGetCommand runs immediately because the SDK retry mechanism starts fresh every time.
> 
> **With the delay**
> 
> 1. The BatchGetCommand fails and retries after 200ms, the second retry succeeds and returns 10 items with 40 in 'Unprocessed Keys.
> 2. We have an indicator that the DB is busy so we delay 200ms externally before calling the BatchGetCommand again for the same data.
> 3. 40 items are passed back in and are retried and fail.  Internally the second retry also fails and delays, the 3rd retry succeeds and returns 20 items with 20 in 'UnprocessedKeys'
> 4. We have an indicator that the DB is really busy so we delay 400ms externally before calling the BatchGetCommand again.
> 
> With this version I thought we were being proactive by giving the dynamoDB a chance to catch up by reducing the load on it and giving it a better probability of completing the request with all items.

Got it! Then it looks good to me ;)

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts
```diff
@@ -310,3 +315,13 @@ export const getSubmissionsByFormat = async ({
...
   }
 };
+
+export const confirmSubmissionCodes = async (confirmationCodes: string[], formId: string) => {
+  try {
+    const { ability } = await authCheckAndRedirect();
+
+    return confirmResponses(ability, confirmationCodes, formId);
+  } catch (e) {
+    throw new Error("There was an error. Please try again later.");
```
Comment (wmoussa-gc) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts:325]: should we pass in the e object?

craigzour (dismissed):

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts
Comment (bryan-robitaille) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts:325]: I purposefully left it out since this traverses the server / client boundary to avoid leaking any sensitive information back to the client.

Comment (wmoussa-gc) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts:325]: make sense, was hoping we could log the actual error somewhere to help troubleshoot, perhaps Im missing a layer

Comment (bryan-robitaille) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts:325]: I get what you're saying.  I'll go back through the layers to see if there is a chance that any unhandled errors could bubble up. The try / catch at that level what just a pattern we started adopting for the boundary actions that didn't have any precise error return objects.

Comment (bryan-robitaille) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts:329]: @wmoussa-gc I've added a catch all error that is logged server side as a Warning.

craigzour (approved):

PR #4046, opened 2024-07-22
feat: Add Link to Logic View in Options Modal

### LINKED ISSUE #3976: Moving people over to the logic view
### Description:
Iterated designs that were informed by research and testing :)

Figma file: https://www.figma.com/design/bIcbQm61JbkCy9Nu8XE7u6/Conditional-logic-Alpha?node-id=1-35&m=dev&focus-id=4396-2808

**Updated content in Comments/Figma**

### Acceptance Criteria:
- [ ] there will be an accordion just before the "Save" button to detail the branching logic feature we have
- [ ] there will be a button to directly take them to the logic view

ShadeWyrm:
Closes #3976 

Passes the formId across the application to be able to route to the logic view - so just a quick test to make sure options don't return "undefined" in the URL at the top.

To test:
-> Open or Create a form.
-> Add a question that can have options.
-> Click an option to show "add rules"
-> Click the "add rules", and then click the "Have a complex form? Try branching instead +"

Double check the form has all the stuff you had, and the URL matches. eg: "en/form-builder/0000/edit" vs "en/form-builder/undefined/edit"

### DIFF
--- a/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/Edit.tsx
+++ b/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/Edit.tsx
@@ -147,7 +147,9 @@ export const Edit = ({ formId }: { formId: string }) => {
             if (element) {
               const questionNumber = getQuestionNumber(element, elementTypes);
               const item = { ...element, index, questionNumber };
-              return <ElementPanel elements={sortedElements} item={item} key={item.id} />;
+              return (
+                <ElementPanel elements={sortedElements} item={item} key={item.id} formId={formId} />
+              );
             }
           })}
       </div>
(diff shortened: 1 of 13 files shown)


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/ModalFormRules.tsx
```diff
@@ -111,6 +113,11 @@ export const ModalFormRules = ({
           </Button>
           <AddOther item={item} onComplete={updateModalFromBase} />
         </div>
+        <div>
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/ModalFormRules.tsx:116]: Minor nit pick here. What about using a <Link> over a button here? This way an AT user would now that navigation will happen when they click/activate it.

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/Edit.tsx
```diff
@@ -147,7 +147,9 @@ export const Edit = ({ formId }: { formId: string }) => {
             if (element) {
               const questionNumber = getQuestionNumber(element, elementTypes);
               const item = { ...element, index, questionNumber };
-              return <ElementPanel elements={sortedElements} item={item} key={item.id} />;
+              return (
+                <ElementPanel elements={sortedElements} item={item} key={item.id} formId={formId} />
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/Edit.tsx:151]: Hopefully in the future we can use a react context for situations like this (lots in the app). Just a future refactoring note when/if we get around to refactoring the app state management and definitely something for outside of this PR.

thiessenp-cds (approved):
Nice work! LGTM :)

PR #4048, opened 2024-07-22
chore: vitest migration

wmoussa-gc:


craigzour:
Was not this a pull request you had already worked on previously? Do we want to merge this now because we are ready to migrate to Vitest?

wmoussa-gc:
the previous PR was against the nextjs-app-router..so had to create this new one to salvage some of the work

wmoussa-gc:
not worth migrating right now

Issue #4040, opened 2024-07-19
Unable to add page title in French


anikbrazeau:
## Description
The field for page title cannot be filled when editing in French.
Might be related to previous issue #3786

## Steps to reproduce
1. Go to GC Forms > Edit
2. Add a new page
3. Try adding a title
4. Toggle to translation, see that title isn't available

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Staging
- **Language**: French (editing), French and English interface
- **Viewport**: Desktop

## Expected behaviour
Ability to edit immediately in EN/FR in Edit/Translations modes when new page is added.

## Screencapture
[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMi8xQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7cb89c57dfe472e37636b4db5ffcabf40d71fdcc/Screen%20Recording%202024-07-19%20at%2011.38.51%E2%80%AFAM.mov)

timarney:
Is this referring to the translation tab / page? 


Tested and seem I can edit from the "questions" editor
 
https://github.com/user-attachments/assets/cd3a2101-01d5-4dee-bea2-3458d118139a

anikbrazeau:
Can't seem to add title when on English interface, editing in French?

Abi-Nada:
@anikbrazeau can you try re-testing? I was able to replicate the issue on Friday but now I can't seem to be able to 🤔

anikbrazeau:
@Abi-Nada was able to replicate just now under these conditions:

- Starting a new form and creating a new page, in French interface, editing in French from get go

thiessenp-cds:
I was also able to duplicate the bug locally
https://github.com/user-attachments/assets/24714c73-412f-431c-be46-e0cd74ff25fe

thiessenp-cds:
The bug is related to the language not being set when the title is update. The code defaults to english so only the english title is ever updated. Slowly digging into this bug.

thiessenp-cds:
@anikbrazeau when you have a minute, could you try testing a fix for this on environment:
https://ucjrj3npfslcehmz5iyzdepps40ftwlz.lambda-url.ca-central-1.on.aws/

PR #4050, opened 2024-07-23
fix: Updates sectionTitle to use form lang as default vs English

thiessenp-cds:
# Summary | Résumé

The issue was when both the page language and the form language match, the language would default to English. So a user with a French page language and a French form would receive English as the language and the app would set the English field even though the French field was active.
To solve this I set the default (if page language equals form language) to the form language.

# Test

Follow the steps in #4040


anikbrazeau (approved):
Works! Thanks Pete!

ShadeWyrm (approved):
Nice clean pass through!~ :)

PR #4088, opened 2024-07-29
fix: Use language property for review page label 

timarney:
# Summary | Résumé

Updates the next action selector to use the current language for the review page label.

Screenshot 2024-07-29 at 2 32 46 PM


thiessenp-cds (approved):
LGTM :)

Issue #4143, opened 2024-07-31
Number form element does not show up in Test/Preview


anikbrazeau:
## Description
A clear and concise description of what the problem is.

## Steps to reproduce
1. Go to GC Forms
2. Create a multi-page form
3. Add a number form element
4. Preview/test form
5. Notice form element missing

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Staging
- **Language**: French
- **Viewport**: Desktop
- **Output**: clzaa9wwk0001pyw2g98fwcww

## Expected behaviour
All pages and elements added to show up in preview

## Screenshots or videos
<img width="737" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMjRqQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b56c69aa282d1898ccd677e3274027807def47e7/Screenshot%202024-07-31%20at%204.20.37%E2%80%AFPM.png" alt="Screenshot 2024-07-31 at 4.20.37 PM.png" />

<img width="852" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMjhqQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3e6745d2134a91f920f679686aaa0239730177be/Screenshot%202024-07-31%20at%204.20.45%E2%80%AFPM.png" alt="Screenshot 2024-07-31 at 4.20.45 PM.png" />

<img width="1372" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM1FqQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e1ed4dcc371f555b2f3de0d841ec45ffdc124022/Screenshot%202024-07-31%20at%204.26.41%E2%80%AFPM.png" alt="Screenshot 2024-07-31 at 4.26.41 PM.png" />

anikbrazeau:
Form file here: https://gcdigital.slack.com/archives/C0676D34GPR/p1722459302012809?thread_ts=1722423776.402899&cid=C0676D34GPR

timarney:
[...]

Issue #4113, opened 2024-07-30
Issue with display rules when removed


anikbrazeau:
## Description
Adding display logic to an option on a question that shows/hides another question, does not get removed when the parent option is removed.

## Related issues

#3969 
#3627 
#4143

## Steps to reproduce
1. Go to GC Forms > Edit
2. Add a radio button or other selectable option question
3. Add options
4. Add rules to options
5. Remove option
6. Add rule
7. Rule gets reattached to new option
8. Deleting an option in the middle makes all other rules cascade upward breaking the intended logic

## Details
- **Browser**: Edge
- **Operating system**: Windows
- **Environment**: Staging
- **Language**: French
- **Viewport**: Tablet

## Expected behaviour
Deleting an option deletes associated rules
Deleting a question deletes associated options
Deleting a page deletes associated form elements

## Screenshots or videos

https://github.com/user-attachments/assets/a06a597d-784a-4b0f-8cb7-24dd6eee8433

![other](https://github.com/user-attachments/assets/ef127ae1-7d2c-4146-a63d-61d10f6d8531)

Abi-Nada:
^ Issue fixed in 4153

Issue #3969, opened 2024-07-08
Rules for deleted options persisted after deletion


connorscarolyns:
User feedback from the Diary study suggests that a rule applied to an option existed after the option was deleted, resulting in the user having to delete and recreate rules.

> I created a yes/no radio button question and experimented with ways to make it work as intended. Once it was working I realized I only needed one radio button, one that a user would use to trigger a new picklist to show. I deleted the unnecessary radio button (could have kept it but ultimately an unnecessary click for the user). The radio button I kept was the one with the rules applied to it. Removing the other radio button broke the rules from functioning but the "edit rules" option was still visible and the correct rules showed when I clicked on "edit rules". Had to remove and re-enter rules again.

Abi-Nada:
Closing as a duplicate of #4113

Issue #3627, opened 2024-05-14
Dynamic fields "Other" and rules behaving weirdly when associated option is removed


anikbrazeau:
## Description
When adding a rule such as "Go to Other" or "Go to X question", and changing the configuration or removing it, the rule seems to stick around even after the option is removed and a new option is added.

See related ticket here for more information: https://app.zenhub.com/workspaces/gcforms-60cb8929764d71000e481cab/issues/gh/cds-snc/platform-forms-client/3606

## Steps to reproduce
1. Go to GC Forms
2. Add a radio button or other select type question
3. Add options
4. Add "Other" rule
5. Remove option
6. Add option
7. See how rule remains associated with the latest option

## Details
- **Browser**: Edge
- **Operating system**: Mac
- **Environment**: Staging
- **Language**: French
- **Viewport**: Desktop

## Expected behaviour
Removing the option should eliminate any rules that were associated to that option

## Screenshots or videos

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMVBiQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d6d72cbe0207353de1750f1c611b3058f9f14e88/Screen%20Recording%202024-05-14%20at%207.57.20%E2%80%AFAM.mov)

samsadasivan:
[Hide/show: Only show the questions within that section in the dropdown#3873](https://app.zenhub.com/workspaces/gcforms-60cb8929764d71000e481cab/issues/gh/cds-snc/platform-forms-client/3873)

Abi-Nada:
^ Issue fixed in 4153

PR #4153, opened 2024-08-01
fix: Adjust ChoiceId indexes when an option is removed.

timarney:
# Summary | Résumé

Fix for related: 
https://github.com/cds-snc/platform-forms-client/issues/4113
https://github.com/cds-snc/platform-forms-client/issues/3969
https://github.com/cds-snc/platform-forms-client/issues/3627
https://github.com/cds-snc/platform-forms-client/issues/4143

Fixes code to use id for `cleanUpRules(String(id), index);` vs passing the index

Adds code to adjust choiceId indexes when an option is removed

Given an array of rules

```json
conditionalRules: [{ choiceId: "1.0" }, { choiceId: "1.1" }, { choiceId: "1.3" }, { choiceId: "1.4" }, { choiceId: "2.0" }],
```

If the `1.3` option is removed we need to re-index the existing choiceRules to account for an option being removed

In this case remove `1.3`

Look for other rules that match the 1.x pattern

If the rules are `below` the option being removed `leave` them alone 

If they are `above` we need to remove `-1` from the current value

i.e. remove `1.3` ... 1.4  is above ... and now becomes 1.3

resulting in

```json
conditionalRules: [{ choiceId: "1.0" }, { choiceId: "1.1" }, { choiceId: "1.3" }, { choiceId: "2.0" }],
```

https://github.com/user-attachments/assets/2d568303-e45e-4e60-b135-038958d5c212

**Test file:**
[show-hide-rules-form-2024-08-01.json](https://github.com/user-attachments/files/16471364/show-hide-rules-form-2024-08-01.json)


lib/formContext.ts
```diff
@@ -325,6 +325,255 @@ export const removeChoiceFromRules = (elements: FormElement[], choiceId: string)
...
+/**
+ * Find the difference between two arrays
+ * @param insideRange - The array to compare
+ * @param aboveRange - The array to compare
+ * @returns {Array} - Returns an array of the difference between the two arrays
+ */
+const diffAboveRange = (insideRange: ConditionalRule[], aboveRange: ConditionalRule[]) => {
+  const insideRangeSet = new Set(toChoiceValues(insideRange));
+  const aboveRangeSet = new Set(toChoiceValues(aboveRange));
+  const rangeDifference = new Set([...insideRangeSet].filter((x) => !aboveRangeSet.has(x)));
```
thiessenp-cds, line 484: Nice

timarney: Was hoping to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/difference

But can't yet

thiessenp-cds:
Nice work!

Testing, removing a rule correctly "cascades" updates to the related rules.

Unrelated I think: (future followup PR)
I ran into what may be an issue that I think is outside the scope of this PR. If I setup a rule to go to a page, delete the page, the logic view updates correctly. But the rule with the deleted page, it still tries to load the deleted page and shows a page in a broken state.

https://github.com/user-attachments/assets/a5b734a1-35ec-434b-b0d7-4b2a18584759

thiessenp-cds (approved):
Nice work! LGTM :)

timarney:
> Unrelated I think: (future followup PR)

Ya --- unrelated for this PR but group rules also likely need the same sort of logic applied as they also use the same indexes.

timarney:
Holding off on merging this until after the Conditional Logic updates go live

timarney:
Will wait for QA testing before merging.

timarney:
Confirmed the same issue exists for groups / next actions https://github.com/cds-snc/platform-forms-client/issues/4169

Abi-Nada:
QA Tested and LGTM, @timarney ! ✅

PR #4158, opened 2024-08-02
feat: add new file type to support SVG logo

wmoussa-gc:
# Summary | Résumé

We've been seeing issues with form owners trying to upload SVG logo files, but our system is blocking them because they are XML-based. SVG is a language for describing two-dimensional graphics in XML and is used to define vector-based graphics for the web.

This PR will allow XML-based files to be uploaded. Since we already allow files like MS Word, XLSX, and MS Excel, the addition of XML should not increase our risk posture.

## Before
image

## After
Response submitted


craigzour (changes requested):
LGTM! 
I think there is test file associated to this code. You could add the XML to the file extension test.

timarney:
Looks good -- do we have a file that was failing to test / before & after ?

wmoussa-gc:
vac-sig-fr
@timarney

craigzour (approved):

Issue #4128, opened 2024-07-30
V1 - Add API repo test and deploy workflows


patheard:
The `forms-api` repo should have GitHub workflows added to do the following for PRs:

- [x] Run API unit tests, linting and formatting.
- [x] Run API Docker image builds.

Once a PR has been merged to main, the following workflows should be added:

- [x] Build and push the new Docker image to the ECR.
- [x] Update the Staging `form-api` ECS service to use the new Docker image.
- [x] Run a Docker vulnerability scan and publish results to the repo.

patheard:
Hit a minor blocker on this in that `forms-api` repo uses `main` as its default branch, while the other two repos use `develop`.  As such, the OIDC `-apply` roles are being created to only trust `develop`, which then fails for the `forms-api` pushes to `main`:
https://github.com/cds-snc/aft-account-request/blob/2d311a7feb1e5fbf01dba255fa693ddf1d476edb/terraform/staging.tf#L113

If all our repos use the same default branch this would become a non-issue, but it can also be worked around.

patheard:
To workaround the OIDC problem, the trust relationship of the Staging `forms-api-apply` IAM role has been manually updated in the console to trust pushes to `main` rather than `develop`
```
arn:aws:iam::${{ env.AWS_ACCOUNT_ID }}:role/forms-api-apply
```

# ⚠ Note
This will get automatically overwritten the next time the AFT account customization pipeline is run against the Staging account so a more permanent fix is needed.

PR #4168, opened 2024-08-06
fix: only delete form viewer task definitions

patheard:
# Summary
Update the task definition cleanup workflow so that it only deletes `form-viewer` task definitions.

# Related
- https://github.com/cds-snc/platform-forms-client/issues/4128


craigzour (approved):
LGTM!

PR #4173, opened 2024-08-07
fix: FileInput max upload size for all files

dsamojlenko:
# Summary | Résumé

Add fileInput validation that validates the total payload of *all* uploaded files (ie, multiple fileInputs on a form, or in a repeating set)

Since all files are serialized and base64 encoded before transfer, we must account for a ~35% file size overhead, so `bodySizeLimit` in nextjs config had to be increased to 8.5MB to allow 6MB of file uploads.

This setting affects *all* serverActions, so this change should be revisited in future. Likely that the longterm solution should be to revert form processing from a serverAction to a regular API endpoint.

Note that there are a few known bugs when using fileInputs across multiple groups/pages. Not addressing that here as the immediate need is to enable a single-page form with multiple fileInputs.


anikbrazeau:
## Proposed default hint text for file upload: 

- **EN**: `Files must not exceed 6 MB in total.`
- **FR**: `Les fichiers ne doivent pas dépasser 6 Mo au total.`

## Proposed default error message text for file upload:

- **EN**: `Files added to the form exceed the total size limit. Upload smaller files for a maximum of 6 MB.`
- **FR**: `Les fichiers ajoutés au formulaire dépassent la limite de taille totale. Téléversez des fichiers plus petit pour un maximum de 6 Mo.`

Proposed link for error message: 
- Go to first file uploaded, assuming we're solving for a one-page context first.
- Revisit later when we make file uploads an available feature for all clients.

i18n/translations/fr/common.json
```diff
@@ -41,6 +41,7 @@
     "phone": "Entrer un numéro de téléphone valide.",
     "regex": "Veuillez utiliser le bon format.",
     "file-size-too-large": "Sélectionner un fichier plus petit.",
+    "file-size-too-large-all-files": "Les fichiers ajoutés au formulaire dépassent la limite de taille totale. Téléversez des fichiers plus petit pour un maximum de 6 Mo.",
```
anikbrazeau, line 44:
```suggestion
    "file-size-too-large-all-files": "Les fichiers dépassent le maximum de 6 Mo. Téléversez des fichiers plus petits.", 
```

timarney (dismissed):
Tested locally seems to work well - nice job.

timarney (approved):

Issue #4083, opened 2024-07-29
Investigate memory leak issue


srtalbot:
Details about investigation: https://docs.google.com/document/d/11jijvStR5T0tSUDBOF6fqXhh-GpKiIz6_M62xoXNF2U

Abi-Nada:
Current action items completed -
Moving this card to Review, as @Abi-Nada will break down remaining items into cards that we can prioritize accordingly.

PR #4179, opened 2024-08-08
* Move redirects from server to client on Contact and Support

ShadeWyrm:
Potential fix for part of #4083


craigzour (approved):
Great! It does fix the mem leak.

PR #4181, opened 2024-08-08
fix: * Add a check for legacy forms to use layout object instead.

ShadeWyrm:
Adds a check for if a form is legacy and has no groups to resort to the old layout for downloading responses.


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts
```diff
@@ -219,7 +219,8 @@ export const getSubmissionsByFormat = async ({
         }
       );
       let sorted: Answer[];
-      if (allowGroupsFlag) {
+      const formGroups = fullFormTemplate.form.groups ?? [];
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts:222]: Can use 

```
formHasGroups
```

Believe

Since conditional rules the groups object can exist but will be {} (object with no keys)

timarney (approved):

PR #4188, opened 2024-08-12
fix: Multiple fileInputs in a repeatingSet

dsamojlenko:
# Summary | Résumé

Fixes multiple fileInputs in a repeatingSet.

Previously, if you attempted to attach more than one file in a repeatingSet you would see the following error in the console and the form submit would fail:

image

Additionally, through testing it appears we'll need to set the maximum total file size lower to accommodate payload limits when sending to Notify. Dropped the overall bodySizeLimit for serverActions to 5MB, meaning the file size limit is actually about 4MB.

## Test instructions

- Create a form with a fileInput in a repeatingSet. ie: [multiple-file-uploads-2024-08-12.json](https://github.com/user-attachments/files/16600563/multiple-file-uploads-2024-08-12.json)
- Configure the form to send by email.
- Test the form in Preview and attach more than one file in the repeatingSet.


thiessenp-cds (commented):
Unrelated to this PR, I think this code could really help readability if it were refactored. Especially the deep nesting, including nested try-catches.

i18n/translations/fr/common.json
```diff
@@ -41,7 +41,7 @@
     "phone": "Entrer un numéro de téléphone valide.",
     "regex": "Veuillez utiliser le bon format.",
     "file-size-too-large": "Sélectionner un fichier plus petit.",
-    "file-size-too-large-all-files": "Les fichiers dépassent le maximum de 6 Mo. Téléversez des fichiers plus petits.",
+    "file-size-too-large-all-files": "Les fichiers dépassent le maximum de 3.5 Mo. Téléversez des fichiers plus petits.",
```
thiessenp-cds, line 44: Would be nice to pass in the variable for max vs continuing to update this string

i18n/translations/en/common.json
```diff
@@ -41,7 +41,7 @@
     "phone": "Enter a valid telephone number.",
     "regex": "Enter the correct format.",
     "file-size-too-large": "Select a smaller file.",
-    "file-size-too-large-all-files": "Files exceed the 6 MB maximum. Upload smaller files.",
+    "file-size-too-large-all-files": "Files exceed the 3.5 MB maximum. Upload smaller files.",
```
thiessenp-cds, line 44: Would be nice to pass in the variable for max vs continuing to update this string

thiessenp-cds: Actually let's do that in a separate PR

i18n/translations/fr/common.json
thiessenp-cds: Actually let's do that in a separate PR

dsamojlenko: will do later

i18n/translations/en/common.json
dsamojlenko: will do later

thiessenp-cds (approved):
I tested in the Preview mode and forms-form mode with files under and over the combined 5MB limit, works as expected. This code also looks pretty isolated so a full app test probably isn't necessary (beyond the happy paths)
Nice work, LGTM :)

PR #4193, opened 2024-08-13
fix: Handle repeatingSets that include more than just fileInputs

dsamojlenko:
# Summary | Résumé

If there are multiple inputs in a repeating set, and not all of them are fileInputs, loop over the answers and pull only the fileInputs.


thiessenp-cds (approved):
Works as expected. LGTM :)

Issue #3564, opened 2024-05-07
AddressComplete component


Abi-Nada:
# User story

As a **form builder / program administrator**,
I want to get complete and accurate addresses,
So that what I collect is legitimate and what I send as correspondence gets to the right destination.

As a **form filler / member of the public**,
I want my address to be filled in the right format,
So that I can fill out the form and get access to programs and services more easily, quickly, predictably, and efficiently.

# Context

[_Address_**Complete**](https://www.canadapost-postescanada.ca/ac/) is an address finder component developed by Canada Post, that includes real-time verification of addresses. This helps ensure the accuracy of data submitted in a form and helps people filling out a form save time.

AddressComplete takes a master international data set and supplements it with address data from some of the world’s major postal service providers to fully validate data.

This component enables a rapid searching functionality that works as you type. It includes validation for 240 countries using the official federal registry. Custom addresses can still be added if the address is not found while filling out the form.

There is a cost per API call based on a set limit of credits. As such, we will group all the fields as one block or entity to make one call per submission rather than multiple for each field. 

As part of the integration and implementation we will re-use the combo-box to have the component function similarly to our type-ahead searchable list.

There is a ["find" API and a "retrieve" API](https://www.canadapost-postescanada.ca/ac/support/api/). Need to sort out what the costing model is and if billing for lookup calls (10c?) is only for calls to the retrieve or also the find API? How will it work with credit balances and top ups?

# Links

- [GC Forms Discovery work](https://docs.google.com/document/d/1ugZCOjXXLwH6Jt2zP2m8_9fzrohdvnB1ATscGznZ6C0/edit#heading=h.8igshr6poky9)
- [GC Forms AddressComplete Figma](https://www.figma.com/design/5PqqBQR0s0Q8wkvB75pq2C/Address-complete?node-id=4444-1575&t=JZwaQ3lF3U2wnx2p-4)
- [COVID Alert Address Autocomplete](https://github.com/cds-snc/covid-alert-portal/blob/640379235416594970d5853f5f5c2c9b1790a15c/register/static/register/js/address-autocomplete.js#L36https://github.com/cds-snc/covid-alert-portal/blob/640379235416594970d5853f5f5c2c9b1790a15c/register/static/register/js/address-autocomplete.js%23L36)
- [MOU outline](https://docs.google.com/document/d/1WyOAXHkWl9Ds5dg2go-anjBaNOSv0C4I/edit)
[...]

PR #4197, opened 2024-08-15
feat: AddressComplete

ShadeWyrm:
Closes : #3564

Figma: https://www.figma.com/design/5PqqBQR0s0Q8wkvB75pq2C/Address-complete?node-id=4458-3095&node-type=frame&t=YDsbCF7ik9srXNnW-0

How to test/use:
-> Open a Form, add an Address (Auto-Fill) component. 
--> The preview window will be able to select an address, and fill the other fields with any relevant information.
-> Customize the field.
--> Test form will grab the fields as pulled by customize.
--> Responses will download as above.


timarney:
Just playing around if these are known issues no worries just pointing out a few things I've run into

I can't seem to toggle between full and split address until after I select one of the checkboxes

https://github.com/user-attachments/assets/72294fec-c51c-4c14-ba8f-0e79466609e8

---

On the preview side 

The auto-complete doesn't seem to kick in until after I blur or backspace from the input.

When I first focus I get a empty "grey" outline where the items will eventually end up.
https://github.com/user-attachments/assets/f5f21944-b8db-4fa4-81f5-1e8a0bcc7f84

CHANGELOG.md
```diff
@@ -38,71 +38,61 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [3.20.0](https://github.com/cds-snc/platform-forms-client/compare/v3.19.0...v3.20.0) (2024-08-26)
 
-
 ### Features
 
-* Updates Repeating Sets, Pages and Rules to work together on the Review page and form output ([#4186](https://github.com/cds-snc/platform-forms-client/issues/4186)) ([5422d6c](https://github.com/cds-snc/platform-forms-client/commit/5422d6c1800a5ff961270e1effc224718558926a))
```
timarney, line 44: Any idea why this is landing in your PR

I see the diff is the `*` getting changed to `-`

app/(gcforms)/[locale]/(form administration)/forms/page.tsx
```diff
@@ -75,9 +75,12 @@ export default async function Page({
       };
     });
 
-    const overdueTemplateIds = await getOverdueTemplateIds(
-      templates.map((template) => template.id)
-    );
+    // @todo - remove this when overdue redis cache is implemented
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/forms/page.tsx:78]: Let's get this out of your PR

Must have come in during a merge?

Last code for this is 

```
const overdueTemplateIds = await getOverdueTemplateIds(
      templates.map((template) => template.id)
    );
    ```

File: CHANGELOG.md (hunk 1 above)
Comment (ShadeWyrm) [reply CHANGELOG.md:44]: git checkout origin/develop -- CHANGELOG.md pulls the *'s changing into -'s as well. Super strange.

Comment (ShadeWyrm) [reply CHANGELOG.md:44]: Found the issue : It's the linter forcing itself onto Changelog.md again. Will fix!~

**Comment by timarney** [comment]
@ShadeWyrm tried testing on preview env --- logged and and or out and it doesn't seem to be working.
https://github.com/user-attachments/assets/e16c4291-44e2-4496-aa47-bab16c8b8e74

**Comment by timarney** [comment]
Not sure if this is related to partial lookup API but I'm struggling to get the addresses I'm searching for vs when I use the [Canada post example](https://www.canadapost-postescanada.ca/ac/support/examples/)

https://github.com/user-attachments/assets/b43cb2fa-f42b-448e-b30b-2de915009a97



Canada Post example ---

https://github.com/user-attachments/assets/9e1673c2-25df-4a88-a816-502c79ec4c03

File: app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/elements/element-dialog/descriptions/AddressComplete.tsx [hunk 3]
```diff
@@ -0,0 +1,43 @@
...
+        className="mb-4 mt-4"
+        alt={t("addElementDialog.addressComplete.title")}
+      />
+      <h3 className="mb-4">{t("addElementDialog.addressComplete.title")}</h3>
+      <p>{t("addElementDialog.addressComplete.description")}</p>
+
+      <ExampleWrapper className="mt-4">
+        <h4 className="mb-4">{t("addElementDialog.addressComplete.whatIsYourAddress")}</h4>
+        <div className="mb-6">
+          <AddressCompleteComponent id="test-address" {...props} />
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/elements/element-dialog/descriptions/AddressComplete.tsx:38]: Can we just pass the props in here ...

```
 <AddressCompleteComponent
            unitNumber={true}
            civicNumber={true}
            streetName={true}
            city={true}
            province={true}
            ```
            
  1) props  isn't re-used           
  2) with  destructuring as a dev I now have to refer back to the props to see what's being passed in vs just looking at the component

File: app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts [hunk 4]
```diff
@@ -125,6 +130,19 @@ const getAnswerAsString = (question: FormElement | undefined, answer: unknown):
...
   }
 
+  if (question && question.type === "addressComplete") {
+    if (!answer) {
+      return "";
+    }
+
+    if (question.properties.full === false) {
+      return answer as string;
+    } else {
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts:140]: Can get rid of the else here?

components/clientComponents/forms/AddressComplete/types.ts
```diff
@@ -0,0 +1,42 @@
...
+export interface AddressCompleteResult {
+  FieldGroup: string;
+  FieldName: string;
+  FormattedValue: string;
+  FieldType: string;
+  FieldSequence: number;
+}
+
+// Props for the AddressComplete component
+export interface AddressCompleteProps extends InputFieldProps {
```
timarney, line 23: Same type different name?

https://github.com/cds-snc/platform-forms-client/pull/4197/files#diff-59ba237409062af9738110bb0f8018b8a37719a559cc4e8c2dca1abf620a2006R37

connorscarolyns:
I don't think the full address field is working as intended. If I select full address, I expect to see one field that populates with all the parts of the address that I've chosen. Right now, even with the full address selected, all the parts of the address are showing in separate fields.

ShadeWyrm:
> On the preview environment I'm now able to find the address I'm looking for now but the typeahead (preview of possible addresses) doesn't seem to be working - still only see results when I backspace.
> 
>  Kapture.2024-09-18.at.08.48.00.mp4

There seems to be a delay in the component being "ready". I'm looking to see how to get rid of it - it's related to the feature flag deciding if the component is enabled or not.

components/clientComponents/forms/AddressComplete/types.ts
ShadeWyrm: One is used for the Input during customization of the component, the other is output which won't have the InputFieldProps. Other components had their type in the Responses/Review area, but I kept it in the component.

ShadeWyrm:
> On the preview environment I'm now able to find the address I'm looking for now but the typeahead (preview of possible addresses) doesn't seem to be working - still only see results when I backspace.
> 
>  Kapture.2024-09-18.at.08.48.00.mp4

Still investigating how I can speed it up - first load is a bit rough, but after that it's super fast, so I added a spinner until it's ready.

i18n/translations/fr/form-builder.json
```diff
@@ -44,6 +45,39 @@
       },
       "title": "Adresse"
     },
+    "addressComplete": {
+      "whatIsYourAddress": "Quelle est votre adresse?",
```
anikbrazeau, line 49:
```suggestion
      "whatIsYourAddress": "Adresse postale",
```

i18n/translations/fr/form-builder.json
```diff
@@ -44,6 +45,39 @@
...
+      "country": "Pays",
+      "description": "Remplir automatiquement les adresses à l'aide de l'API \"Compléter l'adresse\" de Postes Canada. Cela permettra aux remplisseurs de formulaires de remplir les champs d'adresse avec une précision de 95 %. ",
+      "label": "Adresse (auto-remplissage)",
+      "postal": "Code postal",
+      "province": "Province ou territoire",
+      "street": {
+        "description": "Indiquez le nom de rue et le numéro d'unité.",
+        "label": "Adresse municipale"
+      },
+      "title": "Adresse (auto-remplissage)",
```
anikbrazeau, line 60:
```suggestion
      "title": "Adresse validée",
```

i18n/translations/fr/review.json
```diff
@@ -4,5 +4,16 @@
   "edit": "Modifier",
   "goBack": "Retourner",
   "submit": "Soumettre",
-  "unknown": "Réponse inconnue"
+  "unknown": "Réponse inconnue",
```
anikbrazeau, line 7:
```suggestion
  "unknown": "Adresse introuvable",
```

i18n/translations/fr/form-builder.json
```diff
@@ -44,6 +45,39 @@
       },
       "title": "Adresse"
     },
+    "addressComplete": {
+      "whatIsYourAddress": "Quelle est votre adresse?",
+      "city": "Ville ou communauté",
+      "country": "Pays",
+      "description": "Remplir automatiquement les adresses à l'aide de l'API \"Compléter l'adresse\" de Postes Canada. Cela permettra aux remplisseurs de formulaires de remplir les champs d'adresse avec une précision de 95 %. ",
```
anikbrazeau, line 52:
```suggestion
      "description": "Le champ d'adresse validée est développé par Postes Canada et utilise l'API AdresseComplète pour vérifier les adresses au fur et à mesure qu'elles sont saisies. Utilisez-le pour une une adresse validée et précise.",
```

i18n/translations/fr/form-builder.json
```diff
@@ -44,6 +45,39 @@
...
+      "label": "Adresse (auto-remplissage)",
```
anikbrazeau, line 53:
```suggestion
      "label": "Adresse validée",
```

i18n/translations/fr/form-builder.json
```diff
@@ -44,6 +45,39 @@
...
+      "postal": "Code postal",
+      "province": "Province ou territoire",
+      "street": {
+        "description": "Indiquez le nom de rue et le numéro d'unité.",
```
anikbrazeau, line 57:
```suggestion
        "description": "Indiquez le nom et le numéro de rue.",
```

i18n/translations/en/review.json
```diff
@@ -4,5 +4,16 @@
   "edit": "Edit",
   "goBack": "Go back",
   "submit": "Submit",
-  "unknown": "Unknown answer"
+  "unknown": "Unknown answer",
```
anikbrazeau, line 7:
```suggestion
  "unknown": "Address not found",
```

i18n/translations/en/review.json
```diff
@@ -4,5 +4,16 @@
...
+  "addressComponents": {
+    "header": "Parts of the address to be included",
+    "desc": "Consider minimizing what is asked to only collect what is necessary.",
+    "unitNumber": "Unit number",
+    "civicNumber": "Civic number",
+    "streetName": "Street name",
+    "city": "City or town",
+    "province": "Province, territory or state",
```
anikbrazeau, line 15:
```suggestion
    "province": "Province, territory, or state",
```

i18n/translations/en/review.json
```diff
@@ -4,5 +4,16 @@
...
+    "postalCode": "Postal or Zip code",
```
anikbrazeau, line 16:
```suggestion
    "postalCode": "Postal or zip code",
```

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,39 @@
...
+      "startTyping": "Start typing an address to search",
+      "components": {
+        "header": "Parts of the address to be included",
+        "desc": "Consider minimizing what is asked to only collect what is necessary.",
+        "unitNumber": "Unit number",
+        "civicNumber": "Civic number",
+        "streetName": "Street name",
+        "city": "City or town",
+        "province": "Province, territory or state",
+        "postalCode": "Postal or Zip code",
```
anikbrazeau, line 77:
```suggestion
        "postalCode": "Postal or zip code",
```

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,39 @@
...
+      "customize": "Customize address format",
+      "startTyping": "Start typing an address to search",
+      "components": {
+        "header": "Parts of the address to be included",
+        "desc": "Consider minimizing what is asked to only collect what is necessary.",
+        "unitNumber": "Unit number",
+        "civicNumber": "Civic number",
+        "streetName": "Street name",
+        "city": "City or town",
+        "province": "Province, territory or state",
```
anikbrazeau, line 76:
```suggestion
        "province": "Province, territory, or state",
```

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,39 @@
...
+      "province": "Province or territory",
+      "street": {
+        "description": "Include street name and number.",
+        "label": "Street address"
+      },
+      "title": "Address (Auto-fill)",
+      "select": "Select which component parts of the address to include.",
+      "options": "Address format options",
+      "fields": "Address fields",
+      "fieldsDesc": "This will determine how the date is entered and how data is kept.",
```
anikbrazeau, line 64:
```suggestion
      "fieldsDesc": "This will determine how address data is stored.",
```

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,39 @@
...
+      "country": "Country",
+      "description": "Auto-fill addresses using Canada Posts Address Complete API. This will allow form fillers to complete address fields with 95% accuracy.",
+      "label": "Address (Auto-fill)",
+      "postal": "Postal code",
+      "province": "Province or territory",
+      "street": {
+        "description": "Include street name and number.",
+        "label": "Street address"
+      },
+      "title": "Address (Auto-fill)",
```
anikbrazeau, line 60:
```suggestion
      "title": "Verified address",
```

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,39 @@
...
       "title": "Address"
     },
+    "addressComplete": {
+      "whatIsYourAddress": "What is your address?",
+      "city": "City, town, or community",
+      "country": "Country",
+      "description": "Auto-fill addresses using Canada Posts Address Complete API. This will allow form fillers to complete address fields with 95% accuracy.",
+      "label": "Address (Auto-fill)",
+      "postal": "Postal code",
+      "province": "Province or territory",
```
anikbrazeau, line 55:
```suggestion
      "province": "Province, territory, or state",
```

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,39 @@
       },
       "title": "Address"
     },
+    "addressComplete": {
+      "whatIsYourAddress": "What is your address?",
+      "city": "City, town, or community",
+      "country": "Country",
+      "description": "Auto-fill addresses using Canada Posts Address Complete API. This will allow form fillers to complete address fields with 95% accuracy.",
+      "label": "Address (Auto-fill)",
```
anikbrazeau, line 53:
```suggestion
      "label": "Verified address",
```

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,39 @@
       },
       "title": "Address"
     },
+    "addressComplete": {
+      "whatIsYourAddress": "What is your address?",
```
anikbrazeau, line 49:
```suggestion
      "whatIsYourAddress": "Mailing address",
```

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,39 @@
...
+      "city": "City, town, or community",
+      "country": "Country",
+      "description": "Auto-fill addresses using Canada Posts Address Complete API. This will allow form fillers to complete address fields with 95% accuracy.",
```
anikbrazeau, line 52:
```suggestion
      "description": "Verified address is a component developed by Canada Post using the AddressComplete API to look up addresses as they are entered. Use this element for a more accurate and verified address.",
```

i18n/translations/en/form-builder-responses.json
```diff
@@ -271,5 +271,16 @@
...
-  "loading": "Loading..."
+  "loading": "Loading...",
+  "addressComponents": {
+    "header": "Parts of the address to be included",
+    "desc": "Consider minimizing what is asked to only collect what is necessary.",
+    "unitNumber": "Unit number",
+    "civicNumber": "Civic number",
+    "streetName": "Street name",
+    "city": "City or town",
+    "province": "Province, territory or state",
```
anikbrazeau, line 282:
```suggestion
    "province": "Province, territory, or state",
```

i18n/translations/en/form-builder-responses.json
```diff
@@ -271,5 +271,16 @@
...
+    "postalCode": "Postal or Zip code",
```
anikbrazeau, line 283:
```suggestion
    "postalCode": "Postal or zip code",
```

i18n/translations/en/form-builder.json
```diff
@@ -23,6 +23,7 @@
       "name": "Add full name",
       "firstMiddleLastName": "Add individual names",
       "address": "Add address",
+      "addressComplete": "Add address (Auto-fill)",
```
anikbrazeau, line 26:
```suggestion
      "addressComplete": "Add verified address",
```

i18n/translations/fr/form-builder.json
```diff
@@ -44,6 +45,39 @@
...
       "title": "Adresse"
     },
+    "addressComplete": {
+      "whatIsYourAddress": "Quelle est votre adresse?",
+      "city": "Ville ou communauté",
+      "country": "Pays",
+      "description": "Remplir automatiquement les adresses à l'aide de l'API \"Compléter l'adresse\" de Postes Canada. Cela permettra aux remplisseurs de formulaires de remplir les champs d'adresse avec une précision de 95 %. ",
+      "label": "Adresse (auto-remplissage)",
+      "postal": "Code postal",
+      "province": "Province ou territoire",
```
anikbrazeau, line 55:
```suggestion
      "province": "Province, territoire ou état",
```

i18n/translations/fr/form-builder.json
```diff
@@ -44,6 +45,39 @@
...
+      "street": {
+        "description": "Indiquez le nom de rue et le numéro d'unité.",
+        "label": "Adresse municipale"
+      },
+      "title": "Adresse (auto-remplissage)",
+      "select": "Sélectionner les éléments de l'adresse à inclure.",
+      "options": "Options de format d’adresse",
+      "fields": "Champs d’adresse",
+      "fieldsDesc": "Ceci détermine la manière dont la date est saisie et dont les données existent.",
```
anikbrazeau, line 64:
```suggestion
      "fieldsDesc": "Ceci détermine la manière dont les données d'adresse sont stockées.",
```

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,39 @@
...
+        "label": "Street address"
+      },
+      "title": "Address (Auto-fill)",
+      "select": "Select which component parts of the address to include.",
+      "options": "Address format options",
+      "fields": "Address fields",
+      "fieldsDesc": "This will determine how the date is entered and how data is kept.",
+      "fullAddress": "Full address",
+      "splitAddress": "Split address",
+      "customize": "Customize address format",
```
anikbrazeau, line 67:
```suggestion
      "customize": "Customize format",
```

i18n/translations/fr/form-builder.json
```diff
@@ -44,6 +45,39 @@
...
+      "fullAddress": "Adresse complète",
+      "splitAddress": "Adresse fractionnée",
+      "customize": "Adapter le format de l’adresse",
```
anikbrazeau, line 67:
```suggestion
      "customize": "Adapter le format",
```

anikbrazeau (changes requested):

anikbrazeau:
**Issues to investigate or consider**
- [ ] Verified address lookup/dropdown that includes all data: city, province, etc. to avoid errors
- [ ] Instruction/hint for how to search for address in "Test" preview and form filling (Start typing an address to search shows up in "Edit" but not in the output)
- [ ] Difficulty changing address from review page, glitches when trying to modify entity
- [ ] Redundancy of street address and unit/civic number and street name fields
- [ ] Combining "Address" and "Verified address" into one component of different variants

lib/types/form-types.ts
```diff
@@ -60,6 +70,8 @@ export interface ElementProperties {
   autoComplete?: string;
   dateFormat?: string;
   conditionalRules?: ConditionalRule[];
+  full?: boolean;
```
timarney, line 73: Calling this "full" doesn't given us enough context when working with a form element -- 

Can we add this onto the `AddressComponents` 

Thinking it would be nicer to access all via the address properties under something like properties.address.x

Maybe instead of `full` it would be a boolean for splitAddress = true or false 

if (
            question?.type === FormElementTypes.addressComplete &&
            question.properties.splitAddress === false
          ) {
          
          ... split into the pieces

anikbrazeau: Would changing the language to `combined` vs.`split` help? 

Also a thought -- could the component be made simpler by having less options? So that verified address is always full and combined for example? Just food for thought.

ShadeWyrm: I re-used a property from another field, but probably good to not do that. I'll split off into AddressComponents.

ShadeWyrm: > Would changing the language to `combined` vs.`split` help?
> 
> Also a thought -- could the component be made simpler by having less options? So that verified address is always full and combined for example? Just food for thought.

It would simpliify it - but the question I guess I'd have is "What do ours users need/want?" I'm not sure. :)

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,41 @@
...
+      },
+      "title": "Verified address",
+      "select": "Select which component parts of the address to include.",
+      "options": "Address format options",
+      "fields": "Address fields",
+      "fieldsDesc": "This will determine how address data is stored.",
+      "fullAddress": "Full address",
+      "fullAddressDesc": "Collect the address in one field, as one data point.",
+      "splitAddress": "Split address",
+      "splitAddressDesc": "Collect the address in multiple fields, as separate data points.",
```
anikbrazeau, line 68:
```suggestion
      "fields": "Address data",
      "fieldsDesc": "This determines how address data is stored in response data.",
      "fullAddress": "Entire address combined",
      "fullAddressDesc": "Data collected as one entity or line.",
      "splitAddress": "Split address by component parts",
      "splitAddressDesc": "Data separated in multiple entities or lines.",
```

i18n/translations/fr/form-builder.json
```diff
@@ -44,6 +45,41 @@
...
+      },
+      "title": "Adresse validée",
+      "select": "Sélectionner les éléments de l'adresse à inclure.",
+      "options": "Options de format d’adresse",
+      "fields": "Champs d’adresse",
+      "fieldsDesc": "Ceci détermine la manière dont les données d'adresse sont stockées.",
+      "fullAddress": "Adresse complète",
+      "fullAddressDesc": "Recueillir l'adresse avec un seul champ, comme une seule donnée.",
+      "splitAddress": "Adresse fractionnée",
+      "splitAddressDesc": "Recueillir l'adresse avec plusieurs champs, comme données distinctes.",
```
anikbrazeau, line 68:
```suggestion
      "fields": "Données d'adresse",
      "fieldsDesc": "Ceci détermine la manière dont l'adresse est stockée dans les données de réponses.",
      "fullAddress": "Adresse entière combinée",
      "fullAddressDesc": "Données recueillies en tant qu'entité ou ligne unique.",
      "splitAddress": "Adresse fractionnée en parties",
      "splitAddressDesc": "Données divisées en plusieurs entités.",
```

anikbrazeau (changes requested):

ShadeWyrm:
> Grabbed a video of what I'm seeing on the preview link
> 
> Looking for an address on main st in Grimsby
> 
> I'm unable to find the address on our version ---

Should now be all fixed up now!~

components/clientComponents/forms/Review/Review.tsx
```diff
@@ -53,6 +57,15 @@ function formatElementValues(element: ReviewElement) {
     );
   }
 
+  // Case of Address Complete
+  if (element.element?.type === FormElementTypes.addressComplete) {
+    if (element.element.properties?.full === false) {
+      return element.values as string; // We're a split address, broken into components.
+    } else {
+      return getAddressAsString(JSON.parse(element.values as string) as AddressElements);
```
thiessenp-cds, line 65: You may want to consider unlikely the case of invalid JSON. You could use the `safeJSONParse()`  util function. 
For example:
`getAddressAsString(safeJSONParse(element.values as string) as AddressElements)`

All it does is return undefined if there is a thrown JSON parse exception and leave it up to the caller to handle it. We(I) added this util recently and probably forgot to mention it :)

Oh and I see the formattedDate block above (unrelated to this PR) doesn't handle invalid JSON either. Maybe this could be a followup PR instead since there probably other cases.

components/clientComponents/forms/Review/Review.tsx
```diff
@@ -134,6 +186,16 @@ export const Review = ({ language }: { language: Language }): React.ReactElement
...
   useFocusIt({ elRef: groupsHeadingRef });
 
+  //Prior to useMemo, we grab the necessary translations for AddressComplete
+  addressCompleteStrings.unitNumber = t("addressComponents.unitNumber", { lng: language });
+  addressCompleteStrings.civicNumber = t("addressComponents.civicNumber", { lng: language });
+  addressCompleteStrings.streetName = t("addressComponents.streetName", { lng: language });
+  addressCompleteStrings.city = t("addressComponents.city", { lng: language });
+  addressCompleteStrings.province = t("addressComponents.province", { lng: language });
+  addressCompleteStrings.postalCode = t("addressComponents.postalCode", { lng: language });
+  addressCompleteStrings.country = t("addressComponents.country", { lng: language });
```
thiessenp-cds, line 196: This pattern of passing in language strings is unfortunate. I originally wrapped the related code in a useMemo but I didn't really see any performance difference. We (team) may want to think about a pattern/general idea when it makes sense to use useMemo. Also the React compiler will soon automatically add useMemo's where it makes sense in code. Sorry for rambling in your PR, just something to think about in the future as a team.

components/clientComponents/forms/Review/Review.tsx
```diff
@@ -134,6 +186,16 @@ export const Review = ({ language }: { language: Language }): React.ReactElement
...
+  //This is done here, as useTranslation is inacessible inside useMemo.
```
thiessenp-cds, line 197: We should note this example as a case to think about when the team talks about patterns. The useMemo here (unrelated to this PR) was more of a nice to have and probably doesn't make much of a performance difference (would have to test more to be sure). So it could probably be removed to make the above easier. Also the react compiler will soon automatically search code for and insert useMemo's where it makes sense. Just a note, to note :)

components/clientComponents/forms/Review/Review.tsx
ShadeWyrm: Yah, there's a set of error handling I want to add. :)

Added #4341

components/clientComponents/forms/AddressComplete/AdressComplete.tsx
```diff
@@ -0,0 +1,314 @@
+"use client";
```
timarney, line 1: Filename type `AdressComplete.tsx]

lib/types/form-types.ts
ShadeWyrm: Resolved through design change that will be occuring in a later PR.

Dockerfile.pr
```diff
@@ -4,6 +4,8 @@ FROM node:20 as build
 
 ENV NODE_ENV=production
 ENV NEXT_OUTPUT_STANDALONE=true
+ENV NEXT_PUBLIC_ADDRESSCOMPLETE_API_KEY=UR78-BU29-RU35-EP49 
+# Adress Complete API key is domain locked, and provided to the client browser, where the API is processed.
```
timarney, line 8: Typo - `Adress`

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/elements/element-dialog/ElementDialog.tsx
```diff
@@ -8,18 +8,27 @@ import { Groups } from "@lib/hooks/form-builder/useElementOptions";
...
   ref: React.RefObject<HTMLElement>;
 };
 
-const filterElementsByGroup = (elements: ElementOption[], selectedGroup: Groups | "all") => {
+const filterElementsByGroup = (
+  elements: ElementOption[],
+  selectedGroup: Groups | "all",
+  featureFlags: { [key: string]: boolean }
+) => {
+  const excludeId = featureFlags["addressComplete"] ? "address" : "addressComplete";
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/elements/element-dialog/ElementDialog.tsx:23]: Maybe rename this --- to excludeAddressComplete vs excludeId to make the filter a little more clear.

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts
```diff
@@ -240,6 +261,35 @@ export const getSubmissionsByFormat = async ({
             } as Answer;
           }
 
+          // Handle "Split" AddressComplete in a similiar manner to dynamic fields.
+          if (
+            question?.type === FormElementTypes.addressComplete &&
+            question.properties.full === false
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts:267]: Same comment as before

```
question.properties.full 
```

This should be tied back to address I can't tell looking at this code what full is

components/clientComponents/forms/AddressComplete/AddressComplete.tsx
```diff
@@ -0,0 +1,343 @@
...
+
+  //Address Complete elements
+  const [choices, setChoices] = useState([""]);
+  const [addressResultCache, setAddressResultCache] = useState<AddressCompleteChoice[]>([]); // Cache the results from the address search.
+
+  // Cache and Allow values
+  const [isReady, setIsReady] = useState(false);
+  const [allow, setAllow] = useState(false);
+  const [apiKey, setApiKey] = useState("");
+  //const [cache, setCache] = useState<{ [key: string]: AddressElements }>({});
```
timarney, line 28: If we're not using this let's remove it.

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts
Comment (ShadeWyrm) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts]: Missed that one!~ It should of been .addressComponents?.splitAddress instead of .full.

lib/cspScripts.ts
```diff
@@ -38,7 +38,7 @@ export const generateCSP = (): { csp: string; nonce: string } => {
     form-action 'self';
     frame-src www.googletagmanager.com;
     frame-ancestors 'none';
-    connect-src 'self' www.googletagmanager.com www.google-analytics.com;
+    connect-src 'self' www.googletagmanager.com www.google-analytics.com ws1.postescanada-canadapost.ca;
```
thiessenp-cds, line 41: Looks good to me but we may want to double check this with Bryan, the CSP expert :)

lib/managedData/data/countries.json
```diff
@@ -0,0 +1,1496 @@
+[
```
thiessenp-cds, line 1: I wonder if we could pull this in from a third-party so we don't have to manage it in the future? Though I guess the occasional copy paste wouldn't be a big deal either

components/clientComponents/forms/AddressComplete/utils.ts
```diff
@@ -0,0 +1,134 @@
+import { FormElement } from "@lib/types";
+import { AddressCompleteChoice, AddressCompleteResult, AddressElements } from "./types";
+import { Answer } from "@lib/responseDownloadFormats/types";
+
+const autoCompleteUrl =
+  "https://ws1.postescanada-canadapost.ca/AddressComplete/Interactive/AutoComplete/v1.00/json3.ws";
+const retriveAddressUrl =
+  "https://ws1.postescanada-canadapost.ca/AddressComplete/Interactive/RetrieveById/v1.00/json3.ws";
+
+// Function returns address complete list of choices.
```
thiessenp-cds, line 10: Have we thought about abuses for the calls to these APIs?

app/api/components/route.ts
```diff
@@ -0,0 +1,14 @@
+import { NextResponse } from "next/server";
+import { checkFlag } from "@formBuilder/actions";
+
+//Seperated from utils to avoid unavailable dependencies in testing and response downloads.
+export const allowAddressComplete = async () => {
+  return checkFlag("addressComplete");
+};
+
+// This route is used to allow Cypress Testing of components, and provide the API key for AddressComplete.
```
thiessenp-cds, line 9: I wonder if there is anything we should do to secure this endpoint? It just gets a flag and key so not from a privacy/* concern. But I wonder if it's worth mentioning to Pat - do we need to add any new rules/*

anikbrazeau:
### Suggested edits from testing:

- [ ] Combine "street number" "unit number" and "street name" into one: "street address" (ex: 3-123 Main St.)
- [ ] Include street suffix after street name in street address (ex: Ave. / St. / Cres.)
- [ ] Change "Postal code" to "Postal or zip code" to include American and other addresses
- [ ] Make all fields dynamically editable once looked up / populated so people can correct inacuracies
- [ ] Make "Canada" the default country selected to save most users from having to click/scroll if not limited to Canadian addresses
- [ ] Update text in modal to match what is in Pink/Purple in [Figma](https://www.figma.com/design/5PqqBQR0s0Q8wkvB75pq2C/Address-complete?node-id=4458-2427&node-type=section&t=TKBRL8eOy9jjrTHy-0)

components/clientComponents/forms/AddressComplete/AddressComplete.tsx
```diff
@@ -0,0 +1,343 @@
...
+      (country) => country[i18n.language as Language] === countryText
+    );
+    if (country) {
+      // set the country in the address object to the ID of the country.
+      setAddressData("country", country.id);
+    }
+  };
+
+  return (
+    <>
```
thiessenp-cds, line 180: For a followup PR. 
A fieldset requires a legend. Swapping the Description out for a Legend won't work because it may or may not be present. One idea is to have some screen reader only visible text generically describing the fields e.g. "Fill in the Address fields" or something like that.
Here is the WCAG bit if anyone is curious: https://www.w3.org/TR/WCAG20-TECHS/H71.html

components/clientComponents/forms/AddressComplete/utils.ts
ShadeWyrm: They need a key, and are rate limited by CanadaPost on their end. There's not a lot more we can do to prevent people from hitting their endpoints. 

They're also just available for people to hit on the internet right now, and publicly documented.

In an ideal world, they'd provide a "Server Side" implementation that let us secure everything, but as is - this is just really a client side helper thing.

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/elements/element-dialog/ElementDialog.tsx
```diff
@@ -8,18 +8,28 @@ import { Groups } from "@lib/hooks/form-builder/useElementOptions";
...
 };
 
-const filterElementsByGroup = (elements: ElementOption[], selectedGroup: Groups | "all") => {
+const filterElementsByGroup = (
+  elements: ElementOption[],
+  selectedGroup: Groups | "all",
+  featureFlags: { [key: string]: boolean }
+) => {
+  const excludeAddressComplete = featureFlags["addressComplete"] ? "address" : "addressComplete";
+
```
Comment (dsamojlenko) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/edit/components/elements/element-dialog/ElementDialog.tsx:24]: not sure about filtering by featureFlags inside a function intended to filterElementsByGroup.

this filter should probably be applied inside useElementOptions instead where featureFlags are used to enable/disable repeatingSets and fileInput.

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts
```diff
@@ -240,6 +261,35 @@ export const getSubmissionsByFormat = async ({
...
+            return {
+              questionId: question.id,
+              type: FormElementTypes.address,
+              questionEn: question?.properties.titleEn,
+              questionFr: question?.properties.titleFr,
+              answer: addressElements,
+            } as Answer;
+          }
+
+          // return the final answer object
```
Comment (dsamojlenko) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts:292]: I don't think you should be doing this formatting here, it should be handled in the getAnswerAsString function.

And this is currently broken- when I try to download an html formatted Response, I get the following error from the action: `ERROR: Could not create submissions in format html: Cannot read properties of undefined (reading 'en')`

Comment (dsamojlenko) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts:292]: I suppose it's because you want to sometimes return an object instead of a string? Either way I believe it'll be a string- as JSON will need to be serialized anyway.

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts
```diff
@@ -149,12 +167,14 @@ export const getSubmissionsByFormat = async ({
...
   lang,
   revalidate = false,
+  extraTranslations = {},
 }: {
   formID: string;
   ids: string[];
   format: DownloadFormat;
   lang: Language;
   revalidate?: boolean;
+  extraTranslations?: { [key: string]: { en: string; fr: string } };
```
Comment (dsamojlenko) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts:177]: Is there a reason you couldn't use `serverTranslation` wherever you needed to access these?

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/components/Dialogs/DownloadDialog.tsx
```diff
@@ -109,6 +109,19 @@ export const DownloadDialog = ({
...
+        en: t("addressComponents.civicNumber"),
+        fr: t("addressComponents.civicNumber"),
+      },
+      streetName: { en: t("addressComponents.streetName"), fr: t("addressComponents.streetName") },
+      city: { en: t("addressComponents.city"), fr: t("addressComponents.city") },
+      province: { en: t("addressComponents.province"), fr: t("addressComponents.province") },
+      postalCode: { en: t("addressComponents.postalCode"), fr: t("addressComponents.postalCode") },
+      country: { en: t("addressComponents.country"), fr: t("addressComponents.country") },
+    };
+
```
Comment (dsamojlenko) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/components/Dialogs/DownloadDialog.tsx:124]: Can't we use serverTranslation() instead of prop drilling these?

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/components/DownloadSingleButton.tsx
```diff
@@ -29,12 +29,35 @@ export const DownloadSingleButton = ({
...
+          fr: t("addressComponents.streetName"),
+        },
+        city: { en: t("addressComponents.city"), fr: t("addressComponents.city") },
+        province: { en: t("addressComponents.province"), fr: t("addressComponents.province") },
+        postalCode: {
+          en: t("addressComponents.postalCode"),
+          fr: t("addressComponents.postalCode"),
+        },
+        country: { en: t("addressComponents.country"), fr: t("addressComponents.country") },
+      };
```
Comment (dsamojlenko) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/components/DownloadSingleButton.tsx:52]: Same comment as above

components/clientComponents/forms/AddressComplete/AddressComplete.tsx
```diff
@@ -0,0 +1,309 @@
...
+      setAllow(jsonData.allowed);
+      setApiKey(jsonData.key);
+      setIsReady(true);
+    };
+
+    if (apiKey === "") {
+      // Don't recheck.
+      checkAllowed();
+    }
+  }, [apiKey]);
```
dsamojlenko, line 53: Shouldn't this just be controlled by the feature flag?
Is there a better way of providing the API key? It's not sensitive, could the production and staging keys just be embedded and switched based on APP_ENV?

dsamojlenko:
In a multi-page form, when you navigate away from a page with an address input, then back to that page- the Street Address is empty 
Screen Recording 2024-10-11 at 10 27 59 AM

components/clientComponents/forms/AddressComplete/AddressComplete.tsx
```diff
@@ -0,0 +1,309 @@
...
+        if (JSON.stringify(parsedValue) !== JSON.stringify(addressObject)) {
+          setAddressObject(parsedValue);
+        }
+      } catch (e) {
+        if (addressObject !== null) {
+          setAddressObject(null);
+        }
+      }
+    }
+  }, [field.value]);
```
dsamojlenko, line 73: As previously discussed, this useEffect is unnecessary. I implemented the same in the formattedDate component, but later realized that it was in conflict with the useEffect immediately following.

A better solution would be to set the addressObject initial state from the field value instead:

```
const [addressObject, setAddressObject] = useState<AddressElements | null>(
    field.value ? JSON.parse(field.value) : null
);
  ```

File: app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts (hunk 47 above)
Comment (ShadeWyrm) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts]: It's to return an object allowing a piggy backing off the same functionality as the dynamic rows.

File: components/clientComponents/forms/AddressComplete/AddressComplete.tsx [hunk 53]
```diff
@@ -0,0 +1,309 @@
...
+
+  // Update the field value when the address object changes
+  useEffect(() => {
+    const newValue = addressObject ? JSON.stringify(addressObject) : "";
+    // Only update if newValue is different from field.value
+    if (newValue !== field.value) {
+      helpers.setValue(newValue);
+    }
+  }, [addressObject, helpers]);
```
dsamojlenko, line 82: [Continuing from above] You can see the problem with this code if you update your useEffect dependencies to clear the typescript warning, the bug you were trying to fix comes back.

By initializing the object as described in the previous comment, you can remove the newValue/oldValue check from this second useEffect as well.

components/clientComponents/forms/AddressComplete/AddressComplete.tsx
ShadeWyrm: This was before the fun feature flag provider was made, will refactor.

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/components/Dialogs/DownloadDialog.tsx
```diff
@@ -109,6 +109,17 @@ export const DownloadDialog = ({
...
+      streetAddress: {
+        en: t("addressComponents.streetAddress"),
+        fr: t("addressComponents.streetAddress"),
+      },
+      city: { en: t("addressComponents.city"), fr: t("addressComponents.city") },
+      province: { en: t("addressComponents.province"), fr: t("addressComponents.province") },
+      postalCode: { en: t("addressComponents.postalCode"), fr: t("addressComponents.postalCode") },
+      country: { en: t("addressComponents.country"), fr: t("addressComponents.country") },
+    };
+
```
Comment (dsamojlenko) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/components/Dialogs/DownloadDialog.tsx:122]: This stuff shouldn't have to be scattered across unrelated components. Why can't you just i18n?

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/components/DownloadSingleButton.tsx
```diff
@@ -29,12 +29,27 @@ export const DownloadSingleButton = ({
...
+
```
Comment (dsamojlenko) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/components/DownloadSingleButton.tsx:45]: Same comment as above, this stuff shouldn't be repeated across multiple unrelated components. i18n utils should be used.

components/clientComponents/forms/AddressComplete/AddressComplete.tsx
```diff
@@ -0,0 +1,300 @@
...
+      if (
+        !addressResultCache.find((item: AddressCompleteChoice) => item.Id === responseData[i].Id)
+      ) {
+        newElements.push(responseData[i]);
+      }
+    }
+
+    if (newElements.length > 0) {
+      setAddressResultCache((prevCache) => [...prevCache, ...newElements]);
+    }
```
dsamojlenko, line 100: Seems this cache is not really behaving like a cache- ie, it's not preventing excessive API lookups, as every keypress still sends an API call. 

The cache here seems to have been implemented to address a shortcoming in the original Combobox component where it only accepts display strings with no value or id property, so when selecting an option, in onAddressSet we take the display string and do a lookup in the "cache" to get the id for the retrieval api call. 

Would prefer to fix this in the combobox component by adding an id property that you could then use directly instead of having to do this lookup. 

But maybe that's outside the scope of this PR for the moment.

components/clientComponents/forms/AddressComplete/AddressComplete.tsx
```diff
@@ -0,0 +1,300 @@
...
+    if (!allow) {
+      return;
+    } // Abandon if addressComplete is disabled.
+
+    const query = e.target.value;
+    const responseData = await getAddressCompleteChoices(
+      apiKey,
+      query,
+      addressObject?.country || "CAN"
+    );
```
dsamojlenko, line 84: Should we debounce this so it's not called on every keypress?

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts
```diff
@@ -240,6 +261,36 @@ export const getSubmissionsByFormat = async ({
...
+            const addressElements = [reviewElements];
+
+            return {
+              questionId: question.id,
+              type: FormElementTypes.address,
+              questionEn: question?.properties.titleEn,
+              questionFr: question?.properties.titleFr,
+              answer: addressElements,
+            } as Answer;
+          }
```
Comment (dsamojlenko) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/actions.ts:291]: Still think the better place to handle this is in getAnswerAsString. You're really just returning one of two string representations of the stored data.

components/clientComponents/forms/AddressComplete/AddressComplete.tsx
```diff
@@ -0,0 +1,300 @@
...
+      }
+
+      setIsReady(true);
+    };
+
+    if (apiKey === "") {
+      // Don't recheck.
+      checkAllowed();
+    }
+  }, [apiKey]);
```
dsamojlenko, line 61: will just need to add `featureFlags.addressComplete` to the dependency array here

components/clientComponents/forms/AddressComplete/AddressComplete.tsx
ShadeWyrm: 100% out of scope, but yes - that's what it's used for.

components/clientComponents/forms/AddressComplete/AddressComplete.tsx
```diff
@@ -0,0 +1,300 @@
...
+            <Combobox
+              id={`${name}-country`}
+              name={`${name}-country`}
+              choices={countryChoices}
+              onSetValue={(val) => setCountry(val)}
+              className={cn(meta.error && "gc-error-input")}
+              required={required}
+              defaultValue="Canada"
+              data-testid="addresscomplete-input-country"
+            />
```
dsamojlenko, line 215: Wondering why you chose to use Combobox here instead of the ManagedCombobox that you customized? 

I figured the purpose of ManagedCombobox was so you could customize it for your purposes and leave Combobox alone for now, but it looks like you've also made changes to Combobox.

ShadeWyrm: The ManagedCombobox was only to prevent issues around the choices being propagated at runtime, and avoid overcomplicating and adding functionality to the Combobox that broke cypress tests.

I could probably swap both to use Managed.

timarney: Ideally now that we know the managed state works we update the current combobox so we only need the one and not both to maintain.

Would suggest make a separate PR to update the current and pull it into this one now that it's been proved out.

timarney: @dsamojlenko perhaps in the same PR we could do the ID update.

components/clientComponents/forms/AddressComplete/AddressComplete.tsx
ShadeWyrm: I think that would undo some of the purpose of the control? The autocomplete has no cost associated - it's simply the result lookup when they click an address.

anikbrazeau:
Potential bug: I'm having a hard time switching the country and address once one gets selected? It seems hard to correct a potential mistake? Is this a bug?

components/clientComponents/forms/AddressComplete/utils.ts
```diff
@@ -0,0 +1,124 @@
...
+  params += "Key=" + encodeURIComponent(addressCompleteKey);
+  params += "&SearchTerm=" + encodeURIComponent(query);
+  params += "&Country=" + encodeURIComponent(countryCode);
+
+  const response = await fetch(autoCompleteUrl + params, {
+    headers: { "content-Type": "application/x-www-form-urlencoded" },
+    method: "POST",
+  });
+
+  const responseData = await response.json(); //Todo - Error Handling
```
timarney, line 26: Please create an issue for the todo

components/clientComponents/forms/AddressComplete/utils.ts
```diff
@@ -0,0 +1,124 @@
...
+  params += "Key=" + encodeURIComponent(addressCompleteKey);
+  params += "&Id=" + encodeURIComponent(selectedResult);
+  params += "&Country=" + encodeURIComponent(countryCode);
+
+  const response = await fetch(retriveAddressUrl + params, {
+    headers: { "content-Type": "application/x-www-form-urlencoded" },
+    method: "POST",
+  });
+
+  const responseData = await response.json(); //Todo - Error Handling
```
timarney, line 49: Please create an issue for the todo

lib/types/form-types.ts
timarney: full?: boolean;

> Resolved through design change that will be occuring in a later PR.

Please create an issue so we can track this back.

components/clientComponents/forms/AddressComplete/utils.ts
ShadeWyrm: Already exists : #4341

components/clientComponents/forms/AddressComplete/utils.ts
ShadeWyrm: #4341

components/clientComponents/forms/AddressComplete/utils.ts
timarney, line 26:
```suggestion
  const responseData = await response.json(); //Todo #4341  - Error Handling
```

components/clientComponents/forms/AddressComplete/utils.ts
timarney, line 49:
```suggestion
  const responseData = await response.json(); //Todo #4341 - Error Handling
```

timarney:
Added some issues that can be resolved in separate PRs

https://github.com/cds-snc/platform-forms-client/issues/4417
https://github.com/cds-snc/platform-forms-client/issues/4419
https://github.com/cds-snc/platform-forms-client/issues/4420

timarney:
Will look to approve / merge this after the release tomorrow.

i18n/translations/en/form-builder-responses.json
```diff
@@ -271,5 +271,14 @@
     "sending": "Sending...",
     "cancel": "Cancel"
   },
-  "loading": "Loading..."
+  "loading": "Loading...",
+  "addressComponents": {
+    "header": "Parts of the address to be included",
+    "desc": "Consider minimizing what is asked to only collect what is necessary.",
```
anikbrazeau, line 277: Do we still have/need these strings?

i18n/translations/en/form-builder.json
```diff
@@ -23,6 +23,7 @@
       "name": "Add full name",
       "firstMiddleLastName": "Add individual names",
       "address": "Add address",
+      "addressComplete": "Add verified address",
```
anikbrazeau, line 26:
```suggestion
      "addressComplete": "Add address",
```

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,40 @@
       },
       "title": "Address"
     },
+    "addressComplete": {
+      "whatIsYourAddress": "Mailing address",
+      "city": "City, town, or community",
+      "country": "Country",
+      "description": "Verified address is a component that uses the Canada Post AddressComplete API to look up addresses as they are entered. Use this element to collect a more accurate address.",
+      "label": "Verified address",
```
anikbrazeau, line 53:
```suggestion
      "description": "Address is a component that uses the Canada Post AddressComplete API to look up verified addresses as they are entered. Use this element to collect more accurate addresses.",
      "label": "Address",
```

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,40 @@
...
+      "postal": "Postal or zip code",
+      "province": "Province, territory, or state",
+      "street": {
+        "description": "Include street name and number.",
+        "label": "Street address"
+      },
+      "title": "Verified address",
+      "select": "Select which component parts of the address to include.",
+      "options": "Address format options",
+      "fields": "Data Format",
```
anikbrazeau, line 63:
```suggestion
      "fields": "Data format",
```

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,40 @@
...
+      "fieldsDesc": "This determines how the address appears in response data.",
+      "fullAddress": "Entire address combined",
+      "fullAddressDesc": "Data collected as one entity or line.",
```
anikbrazeau, line 66:
```suggestion
      "fullAddressDesc": "Data displays as one line.",
```

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,40 @@
...
+      "splitAddress": "Split address by component parts",
+      "splitAddressDesc": "Data separated in multiple entities or lines.",
```
anikbrazeau, line 68:
```suggestion
      "splitAddress": "Address split into parts",
      "splitAddressDesc": "Data separated in multiple lines.",
```

i18n/translations/en/review.json
```diff
@@ -4,5 +4,16 @@
   "edit": "Edit",
   "goBack": "Go back",
   "submit": "Submit",
-  "unknown": "Unknown answer"
+  "unknown": "Address not found",
+  "addressComponents": {
+    "header": "Parts of the address to be included",
+    "desc": "Consider minimizing what is asked to only collect what is necessary.",
```
anikbrazeau, line 10: Do we still have/need these strings?

i18n/translations/fr/form-builder-responses.json
```diff
@@ -271,5 +271,14 @@
     "sending": "Envoi en cours...",
     "cancel": "Annuler"
   },
-  "loading": "Chargement..."
+  "loading": "Chargement...",
+  "addressComponents": {
+    "header": "Parties de l’adresse à inclure",
+    "desc": "Essayez de minimiser ce qui est demandé pour ne recueillir que ce qui est nécessaire.",
+    "streetAddress": "Adresse",
```
anikbrazeau, line 278:
```suggestion
    "streetAddress": "Adresse municipale",
```

i18n/translations/fr/form-builder.json
```diff
@@ -23,6 +23,7 @@
       "name": "Ajouter un nom complet",
       "firstMiddleLastName": "Ajouter des noms individuels",
       "address": "Ajouter une adresse",
+      "addressComplete": "Ajouter une adresse validée",
```
anikbrazeau, line 26:
```suggestion
      "addressComplete": "Ajouter une adresse",
```

i18n/translations/fr/form-builder.json
```diff
@@ -44,6 +45,40 @@
       },
       "title": "Adresse"
     },
+    "addressComplete": {
+      "whatIsYourAddress": "Adresse postale",
+      "city": "Ville ou communauté",
+      "country": "Pays",
+      "description": "Le champ d'adresse validée utilise l'API AdresseComplète de Postes Canada pour vérifier les adresses au fur et à mesure qu'elles sont saisies. Utilisez-le pour recueillir une une adresse plus précise.",
+      "label": "Adresse validée",
```
anikbrazeau, line 53:
```suggestion
    "addressComplete": {
      "whatIsYourAddress": "Adresse postale",
      "city": "Ville ou communauté",
      "country": "Pays",
      "description": "Le champ d'adresse utilise l'API AdresseComplète de Postes Canada pour vérifier les adresses au fur et à mesure qu'elles sont saisies. Utilisez-le pour recueillir des adresses plus précises.",
      "label": "Adresse",
```

i18n/translations/fr/form-builder.json
```diff
@@ -44,6 +45,40 @@
...
+      },
+      "title": "Adresse validée",
+      "select": "Sélectionner les éléments de l'adresse à inclure.",
+      "options": "Options de format d’adresse",
+      "fields": "Format des données",
+      "fieldsDesc": "Ceci détermine la manière dont l’adresse apparaît dans les données de réponses.",
+      "fullAddress": "Adresse entière combinée",
+      "fullAddressDesc": "Données recueillies en tant qu'entité ou ligne unique.",
+      "splitAddress": "Adresse fractionnée en parties",
+      "splitAddressDesc": "Données divisées en plusieurs entités.",
```
anikbrazeau, line 68:
```suggestion
      "fullAddress": "Adresse entière combinée",
      "fullAddressDesc": "Données affichées en tant qu'une seule ligne.",
      "splitAddress": "Adresse fractionnée en parties",
      "splitAddressDesc": "Données divisées en plusieurs lignes.",
```

anikbrazeau (changes requested):

i18n/translations/en/form-builder.json
```diff
@@ -44,6 +45,40 @@
...
+      "country": "Country",
+      "description": "Address is a component that uses the Canada Post AddressComplete API to look up verified addresses as they are entered. Use this element to collect more accurate addresses.",
+      "label": "Address",
+      "postal": "Postal or zip code",
+      "province": "Province, territory, or state",
+      "street": {
+        "description": "Include street name and number.",
+        "label": "Street address"
+      },
+      "title": "Verified address",
```
anikbrazeau, line 60:
```suggestion
      "title": "Address",
```

i18n/translations/fr/form-builder.json
```diff
@@ -44,6 +45,40 @@
...
+      "country": "Pays",
+      "description": "Le champ d'adresse utilise l'API AdresseComplète de Postes Canada pour vérifier les adresses au fur et à mesure qu'elles sont saisies. Utilisez-le pour recueillir des adresses plus précises.",
+      "label": "Adresse",
+      "postal": "Code postal ou zip",
+      "province": "Province, territoire ou état",
+      "street": {
+        "description": "Indiquez le nom et le numéro de rue.",
+        "label": "Adresse municipale"
+      },
+      "title": "Adresse validée",
```
anikbrazeau, line 60:
```suggestion
      "title": "Adresse",
```

lib/types/form-types.ts
timarney:
```suggestion
```

Can we just remove this ? Where is it being used at this point?

timarney (approved):

timarney:
You can merge whenever the release is out.

PR #4199, opened 2024-08-16
fix: content - clarify limitations with reordering pages if branching is applied

anikbrazeau:
## Problem to solve

Small section where we refer to rules instead of branches, text doesn't make a lot of sense:

image

When clicking on the page flow icon in branching, already have branches applied to the page, get this message,

"This page has branching applied to options
Remove the branches to create a simple page to page connection"


thiessenp-cds (approved):
LGTM

PR #4201, opened 2024-08-20
fix: Adds focus heading behavior for forms with pages (groups)

thiessenp-cds:
# Summary | Résumé

Adds focus header behavior for forms with lengthy pages. Focussing the group header (H2) navigates the page to the top and also announces the change (heading title) for AT users.

## Test

Create a form with a few pages that have a lot of form elements. I found adding a few address component sets helpful for this. Then try navigating back and forth through the form. For each page navigation, the focus should start at the top of the form.

### DIFF
--- a/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/BackButton.tsx
+++ b/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/BackButton.tsx
@@ -4,7 +4,13 @@ import { Button } from "@clientComponents/forms";
 import { Language } from "@lib/types/form-builder-types";
 
 // Must be placed withing context of the GCFormsContext.Provider
-export const BackButton = ({ language }: { language: Language }) => {
+export const BackButton = ({
+  language,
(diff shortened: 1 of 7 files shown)


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/BackButton.tsx
```diff
@@ -4,7 +4,13 @@ import { Button } from "@clientComponents/forms";
 import { Language } from "@lib/types/form-builder-types";
 
 // Must be placed withing context of the GCFormsContext.Provider
-export const BackButton = ({ language }: { language: Language }) => {
+export const BackButton = ({
+  language,
+  callback,
+}: {
+  language: Language;
+  callback?: () => void;
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/BackButton.tsx:12]: Wondering if there's a better name for this maybe just onClick?

callback makes me think of something that's going to happen after a function call happens which isn't exactly what's happing here.

timarney (approved):

Issue #3877, opened 2024-06-21
Accents not showing up properly in Excel on Windows


jprince-cds:
## Description
When opening a CSV file that contains French accents (questions and/or responses), they don't show up properly on Windows. The accents get converted to 2 other symbols. The reason for this is the file is in Unicode format, while Windows Excel seems to be expected ANSI/iso-8859-1 encoding. 

## Steps to reproduce
1. On a Windows PC, double-click a CSV file containing responses with French accents. 
2. You will notice that accents are displayed incorrectly, for example **é** becomes **Ã©**

## Details
- **Browser**: Any
- **Operating system**: Windows
- **Environment**: Production
- **Language**: English
- **Viewport**: Any

## Expected behaviour
Accents should be displayed correctly.

Here is the workaround needed for the files to open up correctly in Excel on Windows.

Right-click the CSV file, Open With, choose Notepad.
Select File, Save as...
In the save dialog, change the encoding to ANSI.
Save the file and double-click to open in Excel. The accents are loading properly. 

## Screenshots or videos
<img width="768" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd2lEQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--67f60fb41dde58ce10e7194ade0054b109d3ab23/image.png" alt="image.png" />

jprince-cds:
[testing-sign-up-form-2024-06-11-responses-reponses.csv.zip](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd3VEQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--841b116bd2faa43c56173f27c8769e301fdd4650/testing-sign-up-form-2024-06-11-responses-reponses.csv.zip)

jprince-cds:
We should inform client in [Ticket 17757](https://cds-snc.freshdesk.com/a/tickets/17757) once this is resolved, as they are using this workaround now.

anikbrazeau:
Related to #3878

timarney:
How often are we seeing support requests for this type of issue come up?

Is this happening for all / most users or a few?


----

For now I found a couple of alternative ways to get excel to pick up the utf-8 encoded files properly at the file level vs making OS level changes.

[...]

PR #4237, opened 2024-08-29
Force UTF-8 encoding on unzipped CSV files

dsamojlenko:
# Summary | Résumé

Update to force UTF-8 encoding using BOM for unzipped CSV files


timarney:
https://github.com/cds-snc/platform-forms-client/issues/3877#issuecomment-2326603060

timarney (approved):

PR #4257, opened 2024-09-09
feat: update new Status#CreatedAt field when changing Status in DynamoDB Vault entries

craigzour:
# Summary | Résumé

Part 3 of https://github.com/cds-snc/platform-forms-client/issues/4287

- Adds or updates new `Status#CreatedAt` property every time an update is made to the legacy `Status` property on Vault `NAME#` items


bryan-robitaille (approved):
LGTM

PR #4231, opened 2024-08-28
[WIP] feat: Self-serve Manage Form Ownership

dsamojlenko:
# Summary | Résumé

Self-serve Manage Form Ownership

| Access from Responses page  | Manage form access dialog    | Invite form |
| ----------------------------------------------- | -------------------------------------------- | ----- |
| image | image | image |

| Re-invite | Cancel invitation |
| ------------------------- | ----------------- |
|  Screen Recording 2024-09-26 at 4 21 49 PM | Screen Recording 2024-09-26 at 4 26 24 PM |

| Remove existing owner | Adding invitees + validation |
| ----- | ----- | 
| Screen Recording 2024-09-26 at 4 27 47 PM | Screen Recording 2024-09-26 at 4 35 24 PM |

| Accept or decline an invitation |
| ----- |
| Screen Recording 2024-09-26 at 4 57 28 PM |

## Test instructions

Using email aliases (ie, firstname.lastname+alias@cds-snc.ca) can be very helpful in testing this PR. If you can run the service locally, consider the Notify Interceptor.

Note that invitations are tied to a user's email address, and can be sent to someone whether they have an existing account or not.

When adding invitees, the following validations are run on email addresses:
- Must be from same domain as logged-in user
- Must be a government email address
- Must not already have access to the form

### Inviting a non-Forms user
- If an invitee does not have an account already, they will be sent an email referring them to the Registration form.
- After successfully registering they will land on their /forms page, where all invitations are displayed.
- If they accept the invitation, a notification goes to all existing owners to notify of the ownership change.

### Inviting a Forms user
- If an invitee already has an account, they will be sent an email instructing them to login to Forms to accept the invitation
- After successfully logging in, they will land on their /forms page, where all invitations are displayed.
- If they accept the invitation, a notification is sent to all existing owners to notify of the ownership change.

### Removing an owner
In the Manage Ownership dialog, existing owners are displayed. To remove an owner, press the Delete button and then Confirm that you want to delete the user. 

You will not be able to remove an Owner if:
- That owner is you
- There is only one owner remaining on the form

When an owner is removed, notifications are sent to the remaining owners, as well as the owner that was removed to confirm the ownership change.

### Cancelling an invitation
In the Manage Ownership dialog, previously sent invitations are displayed. To cancel an invitation, press the Cancel button and Confirm that you want to cancel. The Invitation is deleted, and no notifications are sent.

### Re-sending expired or lost invitations
[...]

PR #4269, opened 2024-09-12
chore: Add Invitation model and migration

dsamojlenko:
# Summary | Résumé

In preparation for PR #4231 
Adds an Invitations prisma model and migration

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -151,7 +151,6 @@ model Invitation {
   id           String   @id @default(cuid())
   email        String
   expires      DateTime
-  role         String   // future use
   templateId   String
   template     Template @relation(fields: [templateId], references: [id], onDelete: Cascade, onUpdate: NoAction)
 }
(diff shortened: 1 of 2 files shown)


prisma/schema.prisma
```diff
@@ -145,3 +146,12 @@ model MagicLink {
   token      String   @unique
   expires    DateTime
 }
+
+model Invitation {
+  id           String   @id @default(cuid())
+  email        String
+  expires      DateTime
+  role         String   // future use
```
bryan-robitaille, line 154: Let's remove the `role` field for now since it won't be in use.  Once we get to that point we'll need to leverage the Privilege model as our project doesn't have 'roles' per se from the authorization sense.
```suggestion
```

bryan-robitaille (changes requested):
Let's just quickly remove the `role` field and we're good to go.

bryan-robitaille (approved):

Issue #4283, opened 2024-09-17
Element drag and drop in start group doesn't land items in correct order


timarney:
## Description

When dragging and dropping elements in the TreeView - start group dropping and element between other elements seems to put the element at the end of the list.  

There also seems to be an issue when dragging up.

Note: The same actions under a different page / group seem to work correctly 

https://github.com/user-attachments/assets/968761c9-91d4-4d35-8e45-c66b088b4c1c

### Steps to re-create

Add elements A, B, C to the start group

Drag A between B and C

PR #4284, opened 2024-09-17
fix: drag and drop for elements for start group

timarney:
# Summary | Résumé

Adjusts drag and drop target index to account for "ghost items".

Fixes: https://github.com/cds-snc/platform-forms-client/issues/4283


timarney:
- Need to handle when dragging from group X items into the start group

dsamojlenko (approved):

PR #4302, opened 2024-09-20
feat: client feature flags

thiessenp-cds:
# Summary | Résumé

Adds a React context provider to allow the client (components) to access feature flags without having to do prop drilling. 

Currently the getFlags() util method is only used in form-builder. A follow up commit could be to use this in other similar areas.

I also have a few ideas to make this a little more fancy but I'd rather keep this as simple as possible for the first iteration.

### DIFF
--- a/lib/cache/flags.ts
+++ b/lib/cache/flags.ts
@@ -108,3 +108,13 @@ const checkMulti = async (keys: string[]): Promise<{ [k: string]: boolean }> => 
 
   return Object.fromEntries(mapped);
 };
+
+export type Flags = {
+  [k: string]: boolean;
+};
+
+// NOTE: Is there any security risk in accessing all our feature flags?
+// Alternatively this could become a static "safe list" of public feature flags that is updated manually.
+export async function getAllFlags(): Promise<Flags> {
+  return checkMulti(await getKeys());
+}
--- a/lib/hooks/useFeatureFlags.tsx
+++ b/lib/hooks/useFeatureFlags.tsx
@@ -1,0 +1,38 @@
+"use client";
+
+import { Flags } from "@lib/cache/flags";
+import { createContext, useContext, useState } from "react";
+
+const featureFlagsContext = createContext({
+  flags: {} as Flags,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  setFlags: (flags: Flags) => {},
+  getFlag: () => {},
+});
+
+export const FeatureFlagsProvider = ({
+  children,
+  featureFlags,
+}: {
+  children: React.ReactNode;
+  featureFlags: Flags;
+}) => {
+  const [flags, setFlags] = useState(featureFlags);
+  return (
+    <featureFlagsContext.Provider value={{ flags, setFlags, getFlag: () => {} }}>
+      {children}
+    </featureFlagsContext.Provider>
+  );
+};
+
+export const useFeatureFlags = () => {
+  const { flags, setFlags } = useContext(featureFlagsContext);
+  return {
+    getFlag: (key: string) => {
+      return flags[key as keyof typeof flags];
+    },
+    setFlag: (key: string, value: boolean) => {
+      setFlags({ ...flags, [key]: value });
+    },
+  };
+};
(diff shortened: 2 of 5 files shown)


lib/cache/flags.ts
```diff
@@ -108,3 +108,13 @@ const checkMulti = async (keys: string[]): Promise<{ [k: string]: boolean }> =>
...
   return Object.fromEntries(mapped);
 };
+
+export type Flags = {
+  [k: string]: boolean;
+};
+
+// NOTE: Is there any security risk in accessing all our feature flags?
+// Alternatively this could become a static "safe list" of public feature flags that is updated manually.
+export async function getAllFlags(): Promise<Flags> {
```
timarney, line 118: Thinking it might be better to pass in an array of flags we want to use ['addressComplete', 'repeatingSets', 'groups'] etc...

thiessenp-cds: That would work to. I just worry about maintaining that array of flags. Or maybe I'm over thinking it?

thiessenp-cds: Thinking... actually on second thought I like the idea of an array of flags passed in and that list maintained E.g. in the layout/page level.

thiessenp-cds: I'm trying to think.. Assuming there is no risk in exposing feature flag names, is there a benefit to choosing which feature flags? 
E.g. imagine a developer is adding a new feature and wonders why that feature flag isn't in the global list of feature flags. I'm sure s/he'd figure it out but it could be frustrating :)

lib/hooks/useFeatureFlags.tsx
```diff
@@ -0,0 +1,33 @@
+"use client";
+
+import { Flags } from "@lib/cache/flags";
+import { createContext, useContext, useState } from "react";
+
+const featureFlagsContext = createContext({
```
timarney, line 6: To keep this consistent we should use `FeatureFlagsContext` here

lib/cache/flags.ts
timarney: Think this depends a bit 

i.e if we wrap the provider around the front facing form (which we'll be doing) ... we don't need to expose the values of all our flags.

lib/hooks/useFeatureFlags.tsx
thiessenp-cds: Nice, thanks - completely missed that. I guess because it's not a component React didn't complain about the casing.

lib/cache/flags.ts
timarney: i.e. does this open up issues 

Can I get access to feature via dev tools
https://github.com/user-attachments/assets/21c93672-ff71-4a6b-be7e-d3050e89b63c

thiessenp-cds: I wonder if all the feature flag names and status could become some kind of an attack vector? I can't think of anything but I guess it's safer to hide as much info as possible (the Java programmer side in me is in full agreement :)
Ok, a safe-list but then the client team will need to remember this. So for each use of the Provider (in each page/layout)

thiessenp-cds: Yup looks like the provider name and values are obfuscated

Dev
1Screenshot 2024-09-23 at 9 24 49 AM

vs. Staging
2Screenshot 2024-09-23 at 9 25 03 AM

timarney:
We'll need to update the form filler page as well as it uses 

const isAllowGrouping = await allowGrouping();

and passes to 

```
<FormWrapper
            formRecord={formRecord}
            currentForm={currentForm}
            allowGrouping={isAllowGrouping}
          />
```

thiessenp-cds:
> We'll need to update the form filler page as well as it uses
> 
> const isAllowGrouping = await allowGrouping();
> 
> and passes to
> 
> ```
> <FormWrapper
>             formRecord={formRecord}
>             currentForm={currentForm}
>             allowGrouping={isAllowGrouping}
>           />
> ```

Yup, that is one of a number of pages that access that feature flag that way. There is also other feature flags. My plan was to do the forms-forms page and other pages in a followup commit. To me, this seems like less work to review and less chance to miss something in a review.
Or do you see a benefit to adding the feature flag provider AND updating all uses of client feature flags in one PR?

timarney (approved):

PR #4329, opened 2024-09-25
chore: Add flag for repeating sets

timarney:
# Summary | Résumé

Adds a dedicated feature flag for repeating sets.


thiessenp-cds (approved):
Tested locally, works as expected. Also the new flag is off by default, so no danger to production. LGTM :)

Issue #2214, opened 2023-06-06
Improve GC Notify callback API retry logic


craigzour:
**User story**
The GC Notify callback API should improve its retry logic by adding a delay between retries (and/or) a maximum number of retries.

**Story context**
While working on https://github.com/cds-snc/platform-forms-client/pull/2202 we discussed possible improvements that could be implemented in order to avoid rapid/infinite retries.

**Acceptance criteria**
- [ ] Add a log when we schedule a retry for `technical-failure`
- [ ] Add maximum retry logic using Redis as a storage for `technical-failure`. Once we have reached the maximum then call `removeProcessedMark()` and stop retrying.
- [ ] Add a log when we have reached the maximum number of attempts for `technical-failure`
- [ ] `technical-failure` should be retried through the Dead Letter Queue instead of Reprocessing Queue
- [ ] Make sure handling of `temporary-failure` stays the same

srtalbot:
@craigzour - I think Notify implemented this? Can you confirm if they did?

craigzour:
@srtalbot I am unsure if you are referring to https://gcdigital.slack.com/archives/C03T4FAE9FV/p1723728078403769 ? If yes then no it does not solve this ticket.

Otherwise I may be not be aware about it so I will try to ask them.

craigzour:
Awaiting confirmation from GCNotify but from I got told it looks like we should not retry sending a form submission that got flagged as `technical-failure`.

PR #4336, opened 2024-09-26
feat: treat technical-failure as permanent when receiving callback from GCNotify

craigzour:
closes https://github.com/cds-snc/platform-forms-client/issues/2214 (see comments for more information)

# Summary | Résumé

- Adds `technical-failure` status (from GCNotify callback) to the list of statuses we should not try to reprocess submission for

### DIFF
--- a/app/api/notify-callback/route.ts
+++ b/app/api/notify-callback/route.ts
@@ -136,9 +136,9 @@ export const POST = middleware([validAuthorizationHeader()], async (req, props) 
     });
   }
 
-  if (deliveryStatus === "permanent-failure") {
+  if (deliveryStatus === "technical-failure" || deliveryStatus === "permanent-failure") {
     logMessage.warn(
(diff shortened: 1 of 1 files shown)


app/api/notify-callback/route.ts
```diff
@@ -136,9 +136,9 @@ export const POST = middleware([validAuthorizationHeader()], async (req, props)
     });
   }
 
-  if (deliveryStatus === "permanent-failure") {
+  if (deliveryStatus === "technical-failure" || deliveryStatus === "permanent-failure") {
```
bryan-robitaille, line 139: Just a comment but another pattern that allows for more statuses to be added in the future could be done like this:
```js
const terminalFailures = ["permanent-failure","technical-failure"];
if (terminalFailures.includes(deliveryStatus)){
// do not retry
}
```

bryan-robitaille (approved):
👍

PR #4340, opened 2024-09-26
fix: improves form timer (old)

thiessenp-cds:
# Summary | Résumé

Improves the form timer by handling the case of forms with groups. If a form has groups, the delay calculation includes required questions based on the group history. The time a user spends on a form is now also included in the calculation.
This is a high risk PR since it updates forms-forms submission logic.

## Test

The form timer should not display in the form-builder Preview. The form timer is only for forms-forms

### Groups Form
Create a form with groups and add a few required questions. View the form you created and open the dev console. Reload the form (to restart the timer) and quickly fill in the form, counting the number of required questions. When you get to the Review page, check the console and see if the final delay makes sense.

For example, a form that has 10 required questions and has been view for 5 seconds, should have a delay of 
2 + (10 - 5) * 2 = 12 seconds

Note that the countdown begins once the Review page loads. So if you linger on the Review page for e.g. 5 seconds and then click the submit button, count down delay should start at e.g. 7 seconds (with a total delay listed of 12 seconds)

### Non-Groups Form
Create a form without groups and add a few required questions. Following the same steps above. The difference is the time a form is viewed is not taken into account (existing behaviour - countdown begins when the single page form is shown).


components/clientComponents/forms/Form/Form.tsx
```diff
@@ -43,15 +43,28 @@ const SubmitButton: React.FC<SubmitButtonProps> = ({
...
   const formTimerEnabled = process.env.NEXT_PUBLIC_APP_ENV !== "test";
 
   // If the timer hasn't started yet, start the timer
-  if (!formTimerState.timerDelay && formTimerEnabled) startTimer(submitDelaySeconds);
+  if (!formTimerState.timerDelay && formTimerEnabled) {
+    // Calculated here to avoid being called on each checkTimer interval (every second)
+    const submitDelaySeconds =
+      secondsBaseDelay + getNumberOfRequiredQuestions() * secondsPerFormElement;
+
+    // TEMP START: For easier testing, to be removed before merge
```
thiessenp-cds, line 60: TODO: Do not merge until this is removed.

Abi-Nada:
I have a question - is it suppose to calculate the time spent on each required question throughout the entire form, prior to the Review Page? Or is it that you need to hang around the Review Page for a certain amount of time (per each required question) before you can submit?

(Asking because when I tested, I waited 10 seconds while filling out each required field. But then when I got to the Review Page, I still had to wait to submit the form?)

components/clientComponents/forms/Form/Form.tsx
```diff
@@ -189,9 +210,39 @@ const InnerForm: React.FC<InnerFormProps> = (props) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [formStatusError, errorList, lastSubmitCount, canFocusOnError]);
 
-  const numberOfRequiredQuestions = form.elements.filter(
-    (element) => element.properties.validation?.required === true
-  ).length;
+  // Done in a callback to allow calling when the submit button is clicked. Otherwise the block
+  // below would be called very time the form values change. Similar reason why useMemo is not used.
+  const getNumberOfRequriedQuestions = () => {
```
timarney, line 215: Let's look at how we can make this function testable via a unit test --passing in the params vs all the work being done inside.

i.e. I think the groupHistoryElements could be a stand-alone function and passed in.

components/clientComponents/forms/Form/Form.tsx
```diff
@@ -44,15 +44,28 @@ const SubmitButton: React.FC<SubmitButtonProps> = ({
...
   const secondsPerFormElement = 2;
-  const submitDelaySeconds = secondsBaseDelay + numberOfRequiredQuestions * secondsPerFormElement;
 
   const formTimerEnabled = process.env.NEXT_PUBLIC_APP_ENV !== "test";
 
   // If the timer hasn't started yet, start the timer
-  if (!formTimerState.timerDelay && formTimerEnabled) startTimer(submitDelaySeconds);
+  if (!formTimerState.timerDelay && formTimerEnabled) {
+    // Calculated here to avoid being called on each checkTimer interval (every second)
+    const submitDelaySeconds =
```
timarney, line 58: Wondering if this work should happen outside of the actual button ... pass the delay as a prop to the button.

Thoughts?

components/clientComponents/forms/Review/Review.tsx
```diff
@@ -61,7 +61,8 @@ function formatElementValues(element: ReviewElement) {
   return String(element.values);
 }
 
-function getReviewItemElements(
+// TODO: move into a helper file?
+export function getReviewItemElements(
```
timarney, line 65: Just looking at the name of this function 

What's a ReviewItem ? --- trying to remember why it's called that.

components/clientComponents/forms/Form/Form.tsx
thiessenp-cds: Yup that could easily pulled out into a function and the result of that function passed into the Submit Button

components/clientComponents/forms/Review/Review.tsx
```diff
@@ -1,3 +1,5 @@
+"use client";
```
thiessenp-cds, line 1: TODO: Come back to this. The build suddenly gave an error about there being no parent with a "use client". The useRef was highlighted as the cause. Why now suddenly? Also the parent does have a "use client".

lib/formContext.ts
```diff
@@ -867,3 +869,142 @@ export const decrementNextActionChoiceIds = (groups: GroupsType, choiceId: strin
...
+      requiredQuestionsCount,
+      delayFromFormData,
+      submitDelay,
+    });
+  }
+  return submitDelay;
+};
+
+// To make future debugging easier
+const ENABLE_SUBMIT_DELAY_DEBUGGING = true;
```
thiessenp-cds, line 982: This could also be an app setting

lib/formContext.ts
```diff
@@ -867,3 +869,142 @@ export const decrementNextActionChoiceIds = (groups: GroupsType, choiceId: strin
...
+  groupHistory: string[];
+  groups: GroupsType | object | undefined;
+  matchedIds: string[];
+  language: string;
+}) => {
+  return groupHistory
+    .map((groupId) => {
+      if (!groups) return [];
+      const group: Group = groups[groupId as keyof typeof groups] || {};
+      const groupElements = getReviewItemElements(
```
timarney, line 896: If group is {} can we avoid the getReviewElements call --- we know it doesn't have elements

lib/hooks/useGCFormContext.tsx
```diff
@@ -139,7 +148,7 @@ export const GCFormsProvider = ({
     return values.current as FormValues;
   };
 
-  // TODO: once groups flag is on, just use formHasGroups
+  // TODO: once groups flag is on, just use formHasGroups - tracked in #4407
   const groupsCheck = (groupsFlag: boolean | undefined) => {
```
timarney, line 152: Noting the flag will be always on however a legacy form will not have a groups prop or a groups layout etc...

lib/formContext.ts
```diff
@@ -867,3 +869,142 @@ export const decrementNextActionChoiceIds = (groups: GroupsType, choiceId: strin
...
+  requiredQuestionsCount: number;
+}) => {
+  const DEFAULT_DELAY = 4;
+  if (!Number.isInteger(requiredQuestionsCount) || !startTime || !currentTime) {
+    return DEFAULT_DELAY;
+  }
+
+  const timeElapsedSeconds = Math.floor((currentTime - startTime) / 1000);
+
+  let delayFromFormData = requiredQuestionsCount - timeElapsedSeconds;
```
thiessenp-cds, line 962: Or would no delay (button ready) make more sense?
So when the time elapsed is greater than the required questions resulting in a negative number (no delay), just submit the form (/button ready)?

lib/hooks/useFormTimer.ts
```diff
@@ -30,7 +30,9 @@ export const useFormTimer = (): [FormTimerStateType, FormTimerDispatchType] => {
   ): { canSubmit: boolean; remainingTime: number; timerDelay?: number; timeLock?: number } {
     switch (action.type) {
       case "START": {
-        if (!action.timerDelay) throw new Error("Missing timerDelay on start FormTimer action");
+        if (action.timerDelay === undefined || !Number.isInteger(action.timerDelay)) {
+          throw new Error("Missing timerDelay on start FormTimer action");
```
thiessenp-cds, line 34: I wonder if it would be better to just show the submit button vs. showing an error

lib/hooks/useFormTimer.ts
```diff
@@ -30,7 +30,9 @@ export const useFormTimer = (): [FormTimerStateType, FormTimerDispatchType] => {
   ): { canSubmit: boolean; remainingTime: number; timerDelay?: number; timeLock?: number } {
     switch (action.type) {
       case "START": {
-        if (!action.timerDelay) throw new Error("Missing timerDelay on start FormTimer action");
```
thiessenp-cds, line 33: This should be left as is, since 0 is intentionally the disabled state. So don't use 0 as the default

thiessenp-cds:
Moved to this PR that takes a simpler approach https://github.com/cds-snc/platform-forms-client/pull/4437

PR #4344, opened 2024-09-27
[Experimental] Notify Interceptor

dsamojlenko:
# Summary | Résumé

Ever work on a feature that sends a lot of emails?
Do you have to log out / in frequently in local dev?
Do you get tired of waiting on verification codes to arrive?
Is your email inbox always cluttered with test notifications?

## Introducing Notify Interceptor.
Think MailHog or MailCatcher but for our Notify integration.
**For local development only**

Screen Recording 2024-09-27 at 12 12 41 PM


timarney:
Might be nice if this was setup kind of like middleware 

if it exist run it if not pass it on for the next thing to handle type of thing.

PR #4345, opened 2024-09-27
fix:  Updates flags to be loaded from file

thiessenp-cds:
# Summary | Résumé

Updates flags to be loaded from file vs. manually maintaining the types list.

### DIFF
--- a/lib/cache/types.ts
+++ b/lib/cache/types.ts
@@ -1,8 +1,10 @@
-// TODO: in the future these could pulled in from default_flag_settings.json
-export const FeatureFlags = {
-  addressComplete: "addressComplete",
-  repeatingSets: "repeatingSets",
-} as const;
+// Load from file so we have a static list of flag names (vs. from Redis state would be important)
+import flagInitialSettings from "flag_initialization/default_flag_settings.json";
+
+export const FeatureFlags = {} as Record<string, string>;
+for (const key in flagInitialSettings) {
+  FeatureFlags[key] = key;
+}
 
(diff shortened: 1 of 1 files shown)


timarney:
We end up without the typing here ... ends up as just a Record any

Screenshot 2024-09-27 at 2 30 38 PM

vs the current lists out the keys for us


Screenshot 2024-09-27 at 2 31 31 PM

The typed const is useful as we can do things like

```
const allowRepeatingSets = getFlag(FeatureFlags.repeatingSets);
```

Also typescript tells us if we try to use a flag that has been removed.

timarney:
You might consider going the other way around ... update the places that use the json to grab the keys from the type.  Not sure if that will work but might be worth a shot.

thiessenp-cds:
> We end up without the typing here ... ends up as just a Record any
> 
> Screenshot 2024-09-27 at 2 30 38 PM
> vs the current lists out the keys for us
> 
> Screenshot 2024-09-27 at 2 31 31 PM
> The typed const is useful as we can do things like
> 
> ```
> const allowRepeatingSets = getFlag(FeatureFlags.repeatingSets);
> ```
> 
> Also typescript tells us if we try to use a flag that has been removed.
> 
> With just a record we end up with no types Screenshot 2024-09-27 at 2 36 06 PM

I remember trying this last week and I was able to see the types, I think. But as for the error you see, I don't see that. Though the type info, yup I'm no longer seeing that. Will look into it.

thiessenp-cds:
I don't think the JSON idea would work for dynamically creating a type definition that can be used in VSCode to show the available types.

PR #4361, opened 2024-10-07
chore: update various form components with updated GCDS styles

timarney:
# Summary | Résumé

Updates support forms , auth (login etc...) to use GCDS styles.


thiessenp-cds (approved):
I tested the updated pages with a few automated a11y scanners as well as with a few screen readers - VoiceOver on Chrome in OSX and Jaws and NVDA on Edge in windows. Works as excepted. LGTM :)

PR #4365, opened 2024-10-07
chore: fix centering for auth pages

timarney:
# Summary | Résumé

Fixes layout issue for login etc... panel centering.


thiessenp-cds (approved):
Works as intended. LGTM :)

PR #4383, opened 2024-10-09
fix: FormattedDate useEffect/initial state bug

### LINKED ISSUE #4380: Formatted date state + useEffect bug
## Description

I don't have full steps to re-produce this yet but have captured videos (see below) the issue happening on local dev and staging.

This is occurring in the useEffect call 

date-input

If / when this happens the input becomes unusable and in turn would prevent form submission if the field is required.

Potentially this only happens when multiple dates are on the page --- but that's unproven 

Staging
https://github.com/user-attachments/assets/98fc0a70-cb1c-4700-85ba-796288acbf08

Dev --- recorded a full run the issue is closer to the end.

https://github.com/user-attachments/assets/32a71877-1f65-4e65-9631-816ae1990941

dsamojlenko:
# Summary | Résumé

Fixes #4380


timarney (approved):

timarney:
Ref: 
https://github.com/cds-snc/platform-forms-client/issues/4380

PR #4386, opened 2024-10-09
chore: Add invitations lib to support Self-serve ownership feature.

dsamojlenko:
# Summary | Résumé

Adds the underlying invitations lib files and test suite.
This lib will support the [Self-serve ownership feature](https://github.com/cds-snc/platform-forms-client/pull/4392).


lib/invitations/acceptInvitation.ts
```diff
@@ -0,0 +1,144 @@
...
+/**
+ * Accept an invitation.
+ * User has created their account or logged into their existing account.
+ *
+ * @param ability (logged in user)
+ * @param invitationId
+ * @returns
+ */
+export const acceptInvitation = async (ability: UserAbility, invitationId: string) => {
+  // Retrieve the invitation
```
timarney, line 23: Wondering if we should grab the ability from 

```
```

vs passing it in ....

bryan-robitaille: The `authCheckAndThrow()` and `authCheckAndRedirect()` functions should only be used in Server Components because it leverages React cache.  Usage outside of a Server Component causes the main auth function to run on each call and recreates an ability and session.  I should update the comments in that code file and ensure doc strings are available that explain the caveats.

lib/invitations/inviteUserByEmail.ts
```diff
@@ -0,0 +1,208 @@
...
+  ability: UserAbility,
+  email: string,
+  formId: string,
+  message: string
+) => {
+  let invitation;
+
+  const sender = await getUser(ability, ability.userID);
+
+  if (!sender) {
```
bryan-robitaille, line 38: `getUser()` throws if no user is found.  You may want to add a `.catch()` statement and then throw your `UserNotFoundError()`

lib/invitations/inviteUserByEmail.ts
```diff
@@ -0,0 +1,208 @@
...
+    throw new UserNotFoundError();
+  }
+
+  const template = await getTemplateWithAssociatedUsers(ability, formId);
```
bryan-robitaille, line 42: I'm pretty sure the `getTemplateWithAssociatedUsers()` requires global User and Manage All Forms permissions.  If this function will be available to normal base users this may throw an `AccessControlError`.

lib/invitations/inviteUserByEmail.ts
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * Retrieve existing invitation for a form by email
+ *
+ * @param email string
+ * @param formId string
+ * @returns Invitation
+ */
+const _retrieveFormInvitationByEmail = async (email: string, formId: string) => {
+  const invitation = await prisma.invitation.findFirst({
```
bryan-robitaille, line 104: Looking at the retrieve, delete, and create invitation functions I might suggest using a `findMany` here instead.  This way if the system ends up in a weird state where multiple invitations for the same form and user exist we can handle it.
Another way to address this would be to set that the `templateId` and `email` must be unique on a record within the Primsma Schema.
```
model Invitation {
  id           String   @id @default(cuid())
  email        String
  expires      DateTime
  templateId   String
  template     Template @relation(fields: [templateId], references: [id], onDelete: Cascade, onUpdate: NoAction)
  @@unique([email, templateId])
}
```

lib/invitations/inviteUserByEmail.ts
```diff
@@ -0,0 +1,208 @@
...
+ * @param email
+ * @param formId
+ */
+export const inviteUserByEmail = async (
+  ability: UserAbility,
+  email: string,
+  formId: string,
+  message: string
+) => {
+  let invitation;
```
bryan-robitaille, line 34: We may want to add a type to `invitation` since it's known and we're already importing the type.

lib/invitations/inviteUserByEmail.ts
```diff
@@ -0,0 +1,208 @@
...
+  if (senderDomain !== recipientDomain) {
+    throw new MismatchedEmailDomainError();
+  }
+
+  // check if user is already invited to the form
+  const previousInvitation = await _retrieveFormInvitationByEmail(email, formId);
+  if (previousInvitation) {
+    invitation = previousInvitation;
+    // if invitation is expired, delete and recreate
+    if (previousInvitation.expires < new Date()) {
```
bryan-robitaille, line 70: What happens to invitations that are never accepted or denied? Do the entries stay in the DB until the form is deleted?

lib/invitations/cancelInvitation.ts
```diff
@@ -0,0 +1,52 @@
...
+      templateId: true,
+    },
+  });
+
+  if (!invitation) {
+    throw new InvitationNotFoundError();
+  }
+
+  checkPrivileges(ability, [
+    { action: "update", subject: { type: "FormRecord", object: { id: invitation.templateId } } },
```
bryan-robitaille, line 29: In order to verify if a User has the privilege to update a Form Record we must pass in the Form Record Object, or it's similie.
In this case we would need to get the Users associated with the template and then pass them in as the `object` being verified.
```js
{ action: "update", subject: { type: "FormRecord", object: { users:[/* users array from actual Form Record */] } } }
```

lib/invitations/cancelInvitation.ts
```diff
@@ -0,0 +1,52 @@
...
+  ]);
+
+  // If no invitation found, return an error
+  if (!invitation) {
```
bryan-robitaille, line 33: I believe this might be a duplicate of line 24

lib/invitations/inviteUserByEmail.ts
dsamojlenko: Updated in the second PR: https://github.com/cds-snc/platform-forms-client/pull/4392/files#diff-c0ef026f39b595cdd15c9a54804ed75e8e473ebc66487370e63ab189cfc4cffaR485-L483

lib/invitations/inviteUserByEmail.ts
dsamojlenko: Settled on updating the schema/db: https://github.com/cds-snc/platform-forms-client/pull/4386/commits/4317aaae1db3326119a164a7181f25f6925f9688

Will ensure error handling on creation is in place.

lib/invitations/inviteUserByEmail.ts
dsamojlenko: Good question. As it is, it would be up to the owner(s) on the form to cleanup old/expired invitations.

Open to suggestions- should we auto-clean based on some interval? Should we try to draw attention to particularly old invitations to encourage cleanup?

lib/invitations/cancelInvitation.ts
dsamojlenko: Updated to retrieve the template using getTemplateWithAssociatedUsers (using the templateId from the invitation) and then pass that in as the object https://github.com/cds-snc/platform-forms-client/pull/4386/commits/388eb61138cce0cc97bf2fc29564412a5e0b6177#diff-b24305f4af2ab739a4a431f95f5882531b0aa2127c0b11a945fc6eebb5c76d59R30-R37

lib/invitations/cancelInvitation.ts
dsamojlenko: Indeed (removed)

lib/invitations/acceptInvitation.ts
```diff
@@ -0,0 +1,144 @@
...
+    },
+    select: {
+      id: true,
+      name: true,
+      jsonConfig: true,
+      users: true,
+    },
+  });
+
+  const form = template?.jsonConfig as FormProperties;
```
timarney, line 117: Suggest we move this down under the !template

lib/invitations/acceptInvitation.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  if (!template) {
+    throw new TemplateNotFoundError();
+  }
+
+  const emailContent = ownerAddedNotification(form.titleEn, form.titleFr, userName);
```
timarney, line 123: Wondering if we should name this function to  ownerAddedTemplate or something --- easy enough to guess what it does based on the variable name but I had to click into the function to see what it's doing --- just returning a template string.

lib/invitations/inviteUserByEmail.ts
timarney: I would think a cron job to clean after x amount of time would make sense.

dsamojlenko: Do we have cron?

bryan-robitaille: Unfortunatley we don't have access to 'cron' which is why we tend to setup so many Lambda functions.  Definitely not urgent and I guess it really depends on how users behave in the app. For example do Forms actually get deleted or are the mostly abandoned? We could track this as out of scope but a follow up  with an issue that needs more research on?

dsamojlenko: I'll create an issue to track this, as you say- see how it's used and if it becomes a problem.

lib/invitations/acceptInvitation.ts
dsamojlenko: done here: https://github.com/cds-snc/platform-forms-client/pull/4392/commits/b2302a98eefcdd89ac38765765f18f7ffe08a125

Issue #4388, opened 2024-10-10
Repeating Set: Clicking "More" on a page Text creates an error


Abi-Nada:
## Description
When I add a Page Text to the Repeating Set and then click "More", I run into an Error.

## Steps to reproduce
1. Create a Repeating Set
2. Add an address form element
3. For the page text "What is your address?" click on "More"

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Staging
- **Language**: EN

## Expected behaviour
There isn't a "More" option typically on page texts, so I'm not quite sure what the expected behaviour should be within a set? Can we just remove it for Page texts?

<img width="1908" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWczQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--acc4dd7055f4f05c2ab46519a2a018cef8b0b8d5/Screenshot%202024-10-09%20at%209.43.33%E2%80%AFPM.png" alt="Screenshot 2024-10-09 at 9.43.33 PM.png" />

## Screenshots or videos

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWMzQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--724341af81dc6d0fa6883cb9ac6ed7985a3982da/Screen%20Recording%202024-10-09%20at%209.38.59%E2%80%AFPM.mov)

PR #4390, opened 2024-10-10
fix: Remove more menu for Page text element

timarney:
# Summary | Résumé

Fixes https://github.com/cds-snc/platform-forms-client/issues/4388


thiessenp-cds (approved):
When I follow the steps to reproduce the error, no error is shown. From the code it looks like if an error is thrown it is caught and ignored. This looks like the intended behaviour as an improvement/work in progress. LGTM :)

PR #4400, opened 2024-10-11
Update main branch in preperation for default branch switch

bryan-robitaille:
# Summary | Résumé
Aligning all repos to use "main" instead of develop


bryan-robitaille:
Need to take a different strategy because there has been drift between the two branches

PR #4437, opened 2024-10-18
fix: update form timer to handle groups

thiessenp-cds:
# Summary | Résumé

Improves the form timer by handling the case of forms with groups. If a form has groups, the delay calculation includes required questions based on the group history. The time a user spends on a form is now also included in the calculation.
This is a high risk PR since it updates forms-forms submission logic.

## Test

The form timer should not display in the form-builder Preview. The form timer is only for forms-forms

### Groups Form
Create a form with groups and add a few required questions. View the form you created and open the dev console. Reload the form (to restart the timer) and quickly fill in the form, counting the number of required questions. When you get to the Review page, check the console and see if the final delay makes sense.

For example, a form that has 10 required questions and has been view for 5 seconds, should have a delay of 
2 + (10 - 5) * 2 = 12 seconds

Note that the countdown begins once the Review page loads. So if you linger on the Review page for e.g. 5 seconds and then click the submit button, count down delay should start at e.g. 7 seconds (with a total delay listed of 12 seconds)

### Non-Groups Form
Create a form without groups and add a few required questions. Following the same steps above. The difference is the time a form is viewed is not taken into account (existing behaviour - countdown begins when the single page form is shown).


lib/hooks/useFormDelayContext.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+export const useFormDelay = () => {
+  const { formDelay, setFormDelay } = useContext(FormDelayContext);
+  return {
+    /**
+     * Adds the number of required questions in the current group to the form delay state.
+     * @param form the current form
+     * @param currentGroupId group Id of the current page
+     */
+    addRequiredQuestions: (form: FormProperties, currentGroupId: string) => {
```
thiessenp-cds, line 73: Rename to "updateFormDelay" and mention setting the time in the comments

lib/hooks/useFormDelayContext.tsx
```diff
@@ -0,0 +1,119 @@
...
+};
+
+export const calculateDelayWithGroups = (
+  startTime: number,
+  endTime: number,
+  requiredQuestions: number
+) => {
+  if (isNaN(startTime) || isNaN(endTime) || isNaN(requiredQuestions)) {
+    return -1;
+  }
```
craigzour, line 46: In which scenario we could end up getting inputs that are not valid numbers?

lib/hooks/useFormDelayContext.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const elapsedTime = Math.floor((endTime - startTime) / 1000);
+  const delayFromFormData = requiredQuestions - elapsedTime;
+  return calculateSubmitDelay(delayFromFormData);
+};
+
+export const calculateDelayWithoutGroups = (formElements: FormElement[]) => {
+  if (!Array.isArray(formElements)) {
+    return -1;
+  }
```
craigzour, line 55: Same question. I am probably missing some context or knowledge here I guess 😅

lib/hooks/useFormDelayContext.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const delayFromFormData = formElements.filter(
+    (element) => element.properties.validation?.required === true
+  ).length;
+  return calculateSubmitDelay(delayFromFormData);
+};
+
+// Turn on for local testing
+const debug = false;
```
craigzour, line 63: Not sure why we would need such boolean. From what I am seeing it is only used in one place to trigger a log.

Is not there an environment variable that we could use to detect whether we are in local or not? I thought there was one but I would have to search for it.

craigzour: I found `process.env.NODE_ENV === "development"` in the `logger.ts` file

lib/hooks/useFormDelayContext.tsx
```diff
@@ -0,0 +1,119 @@
...
+     * @returns delay in seconds or in the case of an error -1 is used to fallback to no delay
+     */
+    getFormDelay: (formElements: FormElement[], hasGroups: boolean) => {
+      try {
+        const endTime = Date.now();
+        const delay = hasGroups
+          ? calculateDelayWithGroups(formDelay.startTime, endTime, formDelay.requiredQuestions)
+          : calculateDelayWithoutGroups(formElements);
+
+        debug && logMessage.info(`Delay: ${delay}, formDelay: ${JSON.stringify(formDelay)}`);
```
craigzour, line 109: I wonder if could not just use the `logMessage.debug` in such case. This way it should only shows in development mode and you would not even need the `debug` boolean.

lib/hooks/useFormDelayContext.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+        // Avoid 0 because the SubmitButton relies on this to disable itself
+        return delay === 0 ? -1 : delay;
```
craigzour, line 112: I could not find where the `0` value would actually disable the button. Could you show me where the code is?

lib/hooks/useFormDelayContext.tsx
thiessenp-cds: I don't think there is a case. But I wanted to make sure, just encase, that this code would not stop a form from submitting. So anywhere there is -1, that will skip the form delay and allow the form to submit right away

lib/hooks/useFormDelayContext.tsx
thiessenp-cds: Right, it's not obvious. If you look in `useFormTimer.ts` on line 33. There the START case, uses !action.timerDelay to check if the form is disabled (disabled=0 see line 55) when trying to start. Where as  using -1 is truthy (will not be flagged as disabled) and would start the countdown with no delay

lib/hooks/useFormDelayContext.tsx
```diff
@@ -0,0 +1,116 @@
...
+    return -1;
+  }
+  const delayFromFormData = formElements.filter(
+    (element) => element.properties.validation?.required === true
+  ).length;
+  return calculateSubmitDelay(delayFromFormData);
+};
+
+export const useFormDelay = () => {
+  const { formDelay, setFormDelay } = useContext(FormDelayContext);
```
bryan-robitaille, line 63: It's usually good practice to check if the Context is available from where the hook is being called.  If the context doesn't exist, the calling node is not a child node of the context then we throw an error.  https://github.com/cds-snc/platform-forms-client/blob/8cdf199ef04cf38b10f5898c9e9a3d9a9ac3c2d8/app/(gcforms)/%5Blocale%5D/(form%20administration)/form-builder/%5Bid%5D/edit/logic/components/flow/provider/FlowRefProvider.tsx#L30

bryan-robitaille (approved):
👍

timarney (approved):
Approving for further testing on staging.

PR #4444, opened 2024-10-21
fix: Nested Addresses for AddressComplete

### LINKED ISSUE #4417: AddressComplete  - nested address
## Description

When searching for an address in some cases multiple  addresses can fall under the same address.

In this case the result ends up being `undefined undefined`

We should look to handle this use case.

Screenshot 2024-10-16 at 11 15 31 AM

ShadeWyrm:
Closes #4417

Changes:
-> API structure is diff from v1 and v2, fixed that.
-> API decides "Retrieve" vs "Find" completely arbitrarily, adds a regex check to pick the right one when the API fails to define itself correctly.
-> Add lookup function for 'Find'.

WIP : The "choices" isn't being updated with the Find return. Working on the why atm.

### DIFF
--- a/components/clientComponents/forms/AddressComplete/types.ts
+++ b/components/clientComponents/forms/AddressComplete/types.ts
@@ -6,8 +6,9 @@ export interface AddressCompleteChoice {
   Id: string;
   Text: string;
   Highlight: string;
+  Cursor: number;
   Description: string;
-  IsRetrievable: boolean;
+  Next: string;
 }
 
 // Address Lookup API returns an array of objects like:
--- a/components/clientComponents/forms/AddressComplete/utils.ts
+++ b/components/clientComponents/forms/AddressComplete/utils.ts
@@ -47,11 +47,33 @@ export const getSelectedAddress = async (
   });
 
   const responseData = await response.json(); //Todo #4341 - Error Handling
+
   const addressData = responseData.Items as AddressCompleteResult[];
 
   const addressComponents = await getAddressComponents(addressData, language);
 
   return addressComponents;
+};
+
+// Function returns the address set from a retreive.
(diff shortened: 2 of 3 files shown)


components/clientComponents/forms/AddressComplete/types.ts
```diff
@@ -6,8 +6,9 @@ export interface AddressCompleteChoice {
   Id: string;
   Text: string;
   Highlight: string;
+  Cursor: number;
   Description: string;
-  IsRetrievable: boolean;
+  Next: string;
```
timarney, line 11: Can we define the possible types here?

Retrieve 
Find

components/clientComponents/forms/AddressComplete/utils.ts
```diff
@@ -125,3 +147,9 @@ export const getAddressAsAnswerElements = (
 
   return answerArray;
 };
+
+// Helper function to test if the address has multiple results.
+export function matchesAddressPattern(input: string): boolean {
+  const pattern = /^.+,\s+[A-Z]{2}(?:,\s+[A-Z0-9]+)?\s+-\s+\d+\s+Addresses$/;
```
timarney, line 153:
```suggestion
  // See issue # 4417  
  const pattern = /^.+,\s+[A-Z]{2}(?:,\s+[A-Z0-9]+)?\s+-\s+\d+\s+Addresses$/;
```

timarney:
Any thoughts on how we can handle if the user clicks out - just looking for thoughts ... this PR is a big improvement so will look to get it merged --- juts the 2 minor comments.



Screenshot 2024-10-22 at 2 21 40 PM

timarney (approved):

PR #4464, opened 2024-10-24
fix: Update Regex for Nested Addresses to Match More Addressess

### LINKED ISSUE #4452: AddressComplete : Nested Address - Not displaying Full Amount of Addressess

ShadeWyrm:
Closes #4452

Issue was the RegEx was not returning correctly for certain addresses causing it to perform a normal search instead of a search by ID.

### DIFF
--- a/components/clientComponents/forms/AddressComplete/utils.ts
+++ b/components/clientComponents/forms/AddressComplete/utils.ts
@@ -150,7 +150,7 @@ export const getAddressAsAnswerElements = (
 
 // Helper function to test if the address has multiple results.
 export function matchesAddressPattern(input: string): boolean {
-  // See issue # 4417  
-  const pattern = /^.+,\s+[A-Z]{2}(?:,\s+[A-Z0-9]+)?\s+-\s+\d+\s+Addresses$/;
+  // Updated regex pattern to match addresses with various postal/zip code formats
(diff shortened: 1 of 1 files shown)


components/clientComponents/forms/AddressComplete/utils.ts
```diff
@@ -150,7 +150,7 @@ export const getAddressAsAnswerElements = (
 
 // Helper function to test if the address has multiple results.
 export function matchesAddressPattern(input: string): boolean {
-  // See issue # 4417  
-  const pattern = /^.+,\s+[A-Z]{2}(?:,\s+[A-Z0-9]+)?\s+-\s+\d+\s+Addresses$/;
+  // Updated regex pattern to make the comma after the city name optional
```
timarney, line 153: We should keep a ref to the PR number and or try to explain in a comment what this regex is doing (what it's for --- when does this situation occur)

For the next dev I won't know what this is for and likely not even what a nested address is.

ShadeWyrm: Let me know if the comments I added sufficiently provide enough information. :)

timarney (approved):

components/clientComponents/forms/AddressComplete/utils.ts
timarney: yup that works.

PR #4465, opened 2024-10-24
chore: Move key generation to inside modal / download 

timarney:
# Summary | Résumé

Update API key generation step to happen when the user presses download after agreeing.

### DIFF
--- a/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/api/components/ApiKeyButton.tsx
+++ b/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/api/components/ApiKeyButton.tsx
@@ -37,7 +37,7 @@ export const ApiKeyButton = ({
           loading={false}
           theme="primary"
           disabled={Boolean(keyId)}
-          onClick={async () => {
+          onClick={() => {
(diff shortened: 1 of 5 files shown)


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/api/components/ApiKeyButton.tsx
```diff
@@ -25,43 +21,23 @@ export const ApiKeyButton = ({
...
+      {showDelete && keyId ? (
         <DeleteKeyButton id={id} keyId={keyId} />
       ) : (
         <GenerateApiKeyButton
-          loading={generatingKey}
+          loading={false}
           theme="primary"
-          disabled={Boolean(keyId) || pendingKey !== null}
+          disabled={Boolean(keyId)}
           onClick={async () => {
```
Comment (craigzour) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/api/components/ApiKeyButton.tsx:40]: I am guessing we probably don't need the `async` here anymore?

```suggestion
          onClick={() => {
```

craigzour (changes requested):
Just a small comment

craigzour (approved):
LGTM!

PR #4477, opened 2024-10-28
chore: swap out classnames package for cn utility (clsx)

timarney:
# Summary | Résumé

We have multiple packages for merging classnames together.  

This PR:

- Updates components that still use classnames to use the `cn` utility 
- Removes the classnames package in favour of using `clsx` and `twMerge`

https://www.linkedin.com/pulse/understanding-differences-between-clsx-classnames-react-terranova-l5dxf

> clsx is a more modern, lightweight version but with the same functionality.


thiessenp-cds (approved):
LGTM :)

PR #4479, opened 2024-10-28
chore: Use jsx from React

timarney:
# Summary | Résumé

Usage of global JSX is deprecated

This updates the file using it to use from React (not global).
 
Screenshot 2024-10-28 at 8 39 45 AM


thiessenp-cds (approved):
LGTM :)

PR #4483, opened 2024-10-28
fix: Adds an error message to the api create dialog

thiessenp-cds:
# Summary | Résumé

Adds an error message to the API create a key dialog. We may want to iterate on this in the future e.g. adding more error messages, adding a support link etc.

# Test

When an error occurs, the dialog should show an error.
Screenshot 2024-10-28 at 12 12 09 PM


i18n/translations/fr/form-builder.json
```diff
@@ -645,6 +645,12 @@
           "label": "Confirmation",
           "required": "(obligatoire)",
           "description": "Saisissez <strong>ACCEPTE</strong> dans la case ci-dessous pour confirmer que vous comprenez vos responsabilités et que vous acceptez les <a href='canada.ca' target='_blank'>procédures de sécurité requises</a> lors de l'enregistrement et du partage de cette clé."
+        },
+        "error": {
+          "createFailed": {
```
timarney, line 650: @anikbrazeau flagging a couple of new strings here :)

timarney:
@thiessenp-cds pushed some minor updates 

- Moved the Alert up 
- Added some margin to the bottom
- Updated some string text

Screenshot 2024-10-29 at 8 45 18 AM

timarney (approved):
Approving.

Given we'll iterate on this further we can get the translations in via a follow-up PR

PR #4493, opened 2024-10-28
fix: fixes an issue with the close date dialog

thiessenp-cds:
# Summary | Résumé

Sets the close form dialog's default date to now. 

Previously when no date was set, the dropdown for January was not changed, so the value was not being set for this case.


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/close/ClosingDateDialog.tsx
```diff
@@ -26,7 +26,8 @@ export const ClosingDateDialog = ({
   const dialogRef = useDialogRef();
   const [hasErrors, setHasErrors] = useState(false);
 
-  const [month, setMonth] = useState<string>("");
+  // Default to January ("1") for the case of no month selection, which would not trigger a change event (not set any value for it)
+  const [month, setMonth] = useState<string>("1");
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/close/ClosingDateDialog.tsx:30]: We should update to default to the current date

Comment (thiessenp-cds) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/close/ClosingDateDialog.tsx:30]: I like that. This way the user will be closer to a future date and not be put in a past date by default.

timarney (approved):
Tested on the preview URL works as expected.

PR #4514, opened 2024-10-29
fix: addressComplete wcag legend

### LINKED ISSUE #4405: AddressComplete : WCAG Fieldset Legend

ShadeWyrm:
Adds missing design system css for fieldset / fieldset:legend.

Closes #4405


styles/_fieldset.scss
```diff
@@ -0,0 +1,73 @@
+@layer reset, default, disabled, focus;
+:root {
```
timarney, line 2: Can you use 

@import "../node_modules/@cdssnc/gcds-tokens/build/web/css/components/fieldset.css";

I did that with the text input etc...

Would prefer if we can just get the vars from them vs the hardcoded.

thiessenp-cds (commented):
The fieldset part looks great. 

I noticed that some aria-labelledby id's don't match the related label id. But that can be for another PR. This is happening in a number of places outside addressComplete.

I think the design system import looks good. But I'll leave that to Tim who has more experience with that :)

components/clientComponents/forms/AddressComplete/AddressComplete.tsx
```diff
@@ -244,7 +249,7 @@ export const AddressComplete = (props: AddressCompleteProps): React.ReactElement
           </div>
         )}
         {!props.canadianOnly && (
-          <div className="mb-6">
+          <div className="mb-6 mt-4">
             <Label htmlFor={`${name}-country`} className="gc-label">
```
timarney, line 253:
```suggestion
            <Label htmlFor={`${name}-country`}>
```

timarney: Screenshot 2024-10-29 at 3 01 08 PM

timarney (approved):

timarney (approved):

PR #4522, opened 2024-10-30
fix: Allow apostrophes in email addresses

### LINKED ISSUE #4494: Unable to create an account with email address containing apostrophe
User is trying to create an account using this pattern:

First.L'ast@cbsa-asfc.gc.ca
(notice the apostrophe above)

The system indicates this email address is not valid, but client mentions it is valid. 

The validation may have changed over time, because we see some accounts with apostrophes in their email address.

Reported in [Freshdesk #19690](https://cds-snc.freshdesk.com/a/tickets/19690)

dsamojlenko:
# Summary | Résumé

Fixes #4494 

On login, registration, and reset-password we were using the valibot email validator which fails on email addresses with apostrophes, which should be valid.

In most places we were additionally using a custom validator, isValidGovEmail which checks for a valid government email domain, but also checks the name portion of the email address, and correctly passes addresses with apostrophes.

This PR removes the valibot check and just uses the isValidGovEmail check instead.


timarney:
Would be worth taking a look at https://www.geeksforgeeks.org/how-to-write-email-validation-test-cases/ to see if there's other test cases we could pull in.

Abi-Nada:
QA tested on the PR environment and it LGTM! I tested:
- Signing up with a new email that included an apostrophe
- Logging in (+ requesting a new 2FA code a couple times)
- Resetting Password

Also double checked to see if the same email would work on Staging to validate the current issue (it doesn't, as expected - so the fix solves it woooo):
Screenshot 2024-10-31 at 8 05 28 AM

bryan-robitaille (commented):
From a security standpoint all good on my end.

timarney (approved):

Issue #4519, opened 2024-10-30
Memory Leak - Contact Us


bryan-robitaille:
Starting October 25th the majority of POST requests to `/contact` have ended with the Load Balancer timing out for the application to respond and returning a 504 to the client.  These timeouts correspond to activity in the memory utilization where the app continually utilizes reserved memory capacity until erroring out and restarting.

dsamojlenko:
Taking a closer look at the Contact form to see if there's any error handling or additional validation we can do, specifically around the freshdesk api call.

dsamojlenko:
https://github.com/cds-snc/platform-forms-client/pull/4529

Added some client-side validation to those forms

anikbrazeau:
Closed for now. We will wait and observe to see if this happens again: https://gcdigital.slack.com/archives/C07U6P00CTD/p1730476841165639

PR #4529, opened 2024-10-30
fix: Add client side validation to public forms

dsamojlenko:
# Summary | Résumé

Related: #4519 
While investigating the recent memory leak issues, there seems to be some correlation to Contact form submissions with large amounts of invalid data. 

Noted that this form (along with Support and Unlock publishing) rely exclusively on server-side validation of form inputs.

Given that Contact and Support are public-facing unauthenticated forms, it's good practice to validate both client-side and server-side. Catching validation errors client-side will prevent unnecessary round trips to the server.

Unclear if this is the specific cause of the memory leak issue, but given the correlation between Contact form submits and the recent spikes, it's a possibility to be considered, and as mentioned is just good practice.

## Testing
This change affects the Contact, Support, and Unlock publishing form.
To test, visit the forms and submit some information.
The form should behave as normal.

To get a better sense of what's changed, open the Network tab of the Developer tools in Chrome.
- Fill the form with some invalid data (bad email, missing required input, etc)
- When you submit, you should see the validation errors appear, but you should not see any activity in the Network tab.
- If you fix the errors in the form and submit again, assuming you've fixed all the errors you will see the submission to the server in the Network tab.
- If you were to compare against the current version of the forms, you would see the form submission sent to the server on every submit valid or not.


anikbrazeau:
Tested and seeing some activity is it what's expected?
Screenshot 2024-10-30 at 5 25 33 PM
Screenshot 2024-10-30 at 5 22 40 PM
Screenshot 2024-10-30 at 5 22 04 PM

dsamojlenko:
> Tested and seeing some activity is it what's expected? Screenshot 2024-10-30 at 5 21 56 PM Screenshot 2024-10-30 at 5 25 33 PM Screenshot 2024-10-30 at 5 22 40 PM Screenshot 2024-10-30 at 5 22 04 PM

yes, that's unrelated.

craigzour (approved):
LGTM! I tested it and it does work as expected :)

app/(gcforms)/[locale]/(support)/unlock-publishing/components/client/Success.tsx
```diff
@@ -1,9 +1,10 @@
-import { serverTranslation } from "@i18n";
+"use client";
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(support)/unlock-publishing/components/client/Success.tsx:1]: The "use client" is not required because `UnlockPublishingForm` is already a Client Component.

bryan-robitaille (dismissed):
Looks good.

craigzour (approved):

PR #4551, opened 2024-10-31
feat: review page for api delivery

thiessenp-cds:
# Summary | Résumé

Adds a new review page for the API mode.

A followup PR could consider what to do when a user directly enters in a URL (e.g. bookmark) to a form state that is not valid. For example, if there is an API key for the form and the user goes to responses/new, it will work but they'll get the wrong view. And vice versa if a user doesn't have an API key and goes directly to /api-responses/.

API Responses - with signed-off responses
Screenshot 2024-11-04 at 3 27 37 PM

API Responses - with no signed-off responses
Screenshot 2024-11-04 at 3 25 17 PM

# Test

Open a form and submit a few responses. Then go the responses page and download and confirm a few.
Then open that form's setting page and generate an API key. The click the responses link. The url should now end with /api-responses and the responses page should only show the responses you confirmed.


dsamojlenko:
Why duplicate the Responses component and create a new path/link? why not just hide the elements you don't need when the API is enabled? By losing the navigation tabs, you've also lost the button to "Manage access" dialog.

timarney:
@dsamojlenko 

> Why duplicate the Responses component and create a new route? 

The future state starts to build up quite a different interface so show / hiding is going to get harder as we go.

Screenshot 2024-11-05 at 7 21 10 AM

Also the "new" tab / route isn't where we want to land for api --- it's the confirmed view so again the routing doesn't match up with the current. 

We're still kicking around the best setup --- which avoids a bunch of checks and redirects and conditionals.

thiessenp-cds:
> Why duplicate the Responses component and create a new route? why not just hide the elements you don't need when the API is enabled? By losing the navigation tabs, you've also lost the button to "Manage access" dialog.

Sure, the main reason is to avoid a lot of if-conditions in the current responses code, making it more difficult to read. The code dose introduce new pages and a new route but a majority of the new code is just "wrappers" composing components etc. IMO this is easier to read and maintain.

timarney:
The review environment isn't working as `checkKeyExists` calls `getZitadelClient`


Screenshot 2024-11-05 at 8 13 24 AM

timarney:
Also re-thinking if we should create a provider for the key check.

thiessenp-cds:
After a call with Tim, I'm going to park this PR for now and try a new approach on a new branch

thiessenp-cds:
Decided to go with other approach to simplify routing (https://github.com/cds-snc/platform-forms-client/pull/4585) - closing this branch

PR #4555, opened 2024-11-01
try: React compiler

timarney:
# Summary | Résumé

Just testing


timarney:
Closing until we have a path forward for Next 15

PR #4560, opened 2024-11-01
fix: Updates save date with correct date conversion

thiessenp-cds:
# Summary | Résumé

Fixes a bug where the save date function was not first converting the date to an ISO date which caused a date miss-match in prisma etc

Here is the related [production thread chat](https://app.slack.com/client/T2G2S06PM/C05G766KW49?selected_team_id=T2G2S06PM).

Closes #4559

PR #4563, opened 2024-11-04
fix: Adds date ISO check to closing a form

thiessenp-cds:
# Summary | Résumé

Adds a date ISO check when closing a form. 
Previously it was possible to send a non ISO date to the server action which would (silently) fail setting in Prisma.  The check now returns an error when the server action receives a non ISO date.

This is a followup PR to #4560 

## Test

Try updating the close form message and saving it. You should not see a toast error.


lib/utils/date/isValidISODate.ts
```diff
@@ -0,0 +1,8 @@
+export const isValidISODate = (dateString: string) => {
```
timarney, line 1: Did a review today --- I want to do some more testing here --- I think there's some edge casees

app/(gcforms)/[locale]/(form administration)/form-builder/actions.ts
```diff
@@ -264,10 +264,10 @@ export const closeForm = async ({
     const { ability } = await authCheckAndThrow();
 
     // closingDate: null means the form is open, or will be set to be open
-    // closingDate: (now/past date) means the form is closed
-    // closingDate: (future date) means the form is scheduled to close in the future
+    // closingDate: a current or past date means the form is closed
+    // closingDate: a future date means the form is scheduled to close in the future
 
-    if (closingDate && !isValidDateString(closingDate)) {
+    if (closingDate && !isValidISODate(String(closingDate))) {
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/actions.ts:270]: Thinking this should live at the `updateClosedData` level vs here.

thiessenp-cds:
Ready for retesting. I think this is the only way to see the error:

Screenshot 2024-11-06 at 9 44 10 AM

lib/templates.ts
```diff
@@ -1513,6 +1514,10 @@ export const updateClosedData = async (
   ]);
   let detailsData: ClosedDetails | null = null;
 
+  if (!isValidISODate(String(closingDate))) {
+    throw new Error("Invalid ISO Date");
```
timarney, line 1518: We should add the date that was trying to be passed to the error so we can see it / and fix as needed,

timarney:
This is failing when you update a message for an open form

Believe it's because your always checking for a iso date even when the closing date is passed as `null` 
Screenshot 2024-11-13 at 9 13 58 AM

timarney (approved):

PR #4566, opened 2024-11-04
chore: Create zip file with Readme for API key generation

timarney:
# Summary | Résumé

- Adds .zip file creation
- Adds Readme file


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/api/templates/Readme.md
```diff
@@ -0,0 +1 @@
+# Readme
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/api/templates/Readme.md:1]: @anikbrazeau 

Do we have any content for this Readme yet?

If not we can look to merge as is and update it later.

PR #4569, opened 2024-11-04
chore: Remove revalidate New tab on download

dsamojlenko:
# Summary | Résumé

We don't need to revalidate the New tab after download anymore.
Previously, the page needed to be revalidated so that the table would rerender.
The code in this area has been refactored since then, and this is not necessary.

Impetus to remove is that this code is erroring under Next 15 and investigation revealed it's unnecessary and can be removed.


timarney:
Tested moving form "downloaded" + sign-off which should move to ` Signed off for removal`

The screen didn't update so maybe an issue ?

dsamojlenko:
checked that on Staging, and that's existing. Believe there used to be a refresh there, can review later.

timarney:
Will close --- we can just take from the Next 15 branch.

timarney (approved):

PR #4578, opened 2024-11-05
fix: Only grab children for answers using the copy row function [Dynamic Rows & Address Complete]

### LINKED ISSUE #4534: AddressComplete: HTML response - Copy and paste table pastes values twice
## Description
When using the HTML Official record, copy answers to clipboard function, pasting into excel causes the vales to paste into a single cell and multiple cells at the same time.

## Steps to reproduce

- Open an HTML record that contains an address component
- Use the copy answer to clipboard button to copy the table
- Paste it into excel
- See who the answers are mixed in with the questions and have columns of their own.

## Details
- **Browser**: (Chrome)
- **Operating system**: (Mac)
- **Environment**: (Staging)
- **Language**: (English)
- **Output**: (HTML official reciept)

## Expected behaviour
Answer to paste in the same way they are displayed on screen

## Additional context
Any other information about the problem here.

ShadeWyrm:
Closes #4534 

The copy paste function for responses was set to grab all dd elements instead of the child most, resulting in any content with dynamic rows to double paste values with that button.

Fixes the issue for both the dynamic rows and addressComplete components. I didn't find a bug associated with dynamic rows to close.


timarney:
@connorscarolyns looks like you opened the issue can you re-test?

connorscarolyns:
Yes this seems to be fixed now

timarney (approved):

PR #4580, opened 2024-11-05
fix: Invitation emails should be case insensitive

dsamojlenko:
# Summary | Résumé

Previously case differences in email addresses would be treated as unique, and invitations would not show up when they should.

This PR makes the query for invitations and the input validation on email addresses case insensitive.
Also changed the name of the `Invitations/index.tsx` file to `Invitations/Invitations.tsx` to make it easier to find.


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/components/ManageFormAccessDialog/ManageUsers.tsx
```diff
@@ -31,7 +31,12 @@ export const ManageUsers = () => {
     let valid = true;
 
     // User already has access
-    if (usersWithAccess.find((user) => user.email === email && !hasOwnProperty(user, "expired"))) {
+    if (
+      usersWithAccess.find(
+        (user) =>
+          user.email.toLowerCase() === email.toLowerCase() && !hasOwnProperty(user, "expired")
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/components/ManageFormAccessDialog/ManageUsers.tsx:37]: Need to trim here?

Comment (dsamojlenko) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[statusFilter]/components/ManageFormAccessDialog/ManageUsers.tsx:37]: The email was being trimmed before being passed into the function, but nothing wrong with belt+suspenders.

timarney (approved):

PR #4599, opened 2024-11-06
chore: Split up key exists function + add error handling

timarney:
# Summary | Résumé

- Breaks the db lookup for checkKeyExists into a standalone function
- Wraps a try catch around the machine code lookup so we can handle the error at the UI level.


thiessenp-cds (approved):
LGTM :)

PR #4614, opened 2024-11-06
chore: AddressComplete - Turn off placeholder text when addresscomplete component is off.

### LINKED ISSUE #4518: AddressComplete - turning flag off after it was on
## Description

- Create a form including AddressComplete with the feature flag on
- Turned flag off
- Complete form
- Check response data

With flag off  --- after it was on.

1) 
A country select shows

Screenshot 2024-10-30 at 10 48 34 AM

Response has `-` for value
Screenshot 2024-10-30 at 10 53 11 AM

2) 

Street address -- is using "typeahead" component and says start typing to search.  Given the flag is off this field will never have a dataset.  If the user is expecting a search to happen it won't.

Screenshot 2024-10-30 at 10 55 53 AM

ShadeWyrm:
Closes #4518


timarney (approved):
Tested locally working as expected.

PR #4634, opened 2024-11-07
chore: Pass the submitForm server action as a prop

dsamojlenko:
# Summary | Résumé

Attempting to fix the issue with server actions on Next 15
https://github.com/vercel/next.js/discussions/58431#discussioncomment-7620263


timarney (approved):
This better work ;)

PR #4648, opened 2024-11-13
chore: remove api access route

timarney:
# Summary | Résumé

Removes API access route and adds ability for user with Manage All forms to use the Response delivery screen to create API key.


thiessenp-cds (approved):
LGTM :)

PR #4655, opened 2024-11-13
feat: adds throttling UI

thiessenp-cds:
# Summary | Résumé

Adds the throttling rate UI to the Form Management screen. This PR goes out to @srtalbot [🎉](https://emojipedia.org/party-popper)

Throttling Pre-populates the weeks (if a rate is set)
Screenshot 2024-11-14 at 1 57 14 PM

Schedule the throttling rate
Screenshot 2024-11-18 at 10 57 39 AM

Make the rate permanent (disables weeks input)
Screenshot 2024-11-18 at 10 59 19 AM

Remove throttling rate (by leaving the form empty and submitting -or- clicking the reset link-button)
Screenshot 2024-11-18 at 10 58 11 AM


lib/cache/throttlingCache.ts
```diff
@@ -0,0 +1,65 @@
+"use server";
+import { logMessage } from "@lib/logger";
+import { getRedisInstance } from "../integration/redisConnector";
+import { getWeeksInSeconds } from "@lib/utils/date/dateConversions";
+
+export const getThrottling = async (
+  formId: string
+): Promise<{ rate: string | null; expires: number }> => {
+  const getParameter = `rate-limit:${formId}`;
```
craigzour, line 9: You could declare the key prefix as a const variable at the top of the file and reuse it across all your functions. This way, if we ever need to change it, there will only one line of code to update.

Example: https://github.com/cds-snc/forms-api/blob/main/src/lib/rateLimiting/tokenBucketProvider.ts#L14

lib/cache/throttlingCache.ts
```diff
@@ -0,0 +1,65 @@
...
+  try {
+    const redis = await getRedisInstance();
+    const weeksInSeconds = getWeeksInSeconds(weeks);
+    const expiresDateInSeconds = Date.now() + weeksInSeconds;
+    logMessage.info(
+      `Scheduling throttling for ${formId} for ${weeks} weeks and expires on ${new Date(
+        expiresDateInSeconds
+      )}`
+    );
+    await redis.setex(modifyParameter, expiresDateInSeconds, "high");
```
craigzour, line 34: I guess we could also apply my previous suggestion to the `high` value.

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/ThrottlingRate.tsx
```diff
@@ -0,0 +1,205 @@
...
+import { formClosingDateEst } from "@lib/utils/date/utcToEst";
+import { logMessage } from "@lib/logger";
+
+export const ThrottlingRate = ({ formId }: { formId: string }) => {
+  const {
+    t,
+    i18n: { language },
+  } = useTranslation("admin-settings");
+  const hasHydrated = useRehydrate();
+  const [hasPreLoaded, setHasPreLoaded] = useState(false);
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/ThrottlingRate.tsx:28]: Can we change this to 

```
const [isLoading, setIsLoading] = useState(false);
```

Preloading makes me think we're loading ahead of time

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/ThrottlingRate.tsx
```diff
@@ -0,0 +1,205 @@
...
+    setSubmitting(true);
+    try {
+      if (permanent) {
+        await permanentThrottling(formId);
+        setSuccess("permanent");
+        return;
+      }
+
+      if (weeks) {
+        await scheduledThrottling(formId, Number(weeks));
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/ThrottlingRate.tsx:75]: Wonder if this should be called something more related to setting an expiry 

setThrottlingExpiry or something like that

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/ThrottlingRate.tsx
Comment (thiessenp-cds) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/ThrottlingRate.tsx:28]: This is used for loading the setting. It kind of is loading data ahead of time. What about `loadingSetting`?

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/ThrottlingRate.tsx
```diff
@@ -0,0 +1,205 @@
...
+    i18n: { language },
+  } = useTranslation("admin-settings");
+  const hasHydrated = useRehydrate();
+  const [hasPreLoaded, setHasPreLoaded] = useState(false);
+  const [submitting, setSubmitting] = useState(false);
+
+  const [weeks, setWeeks] = useState("");
+  const [weeksDisabled, setWeeksDisabled] = useState(false);
+  const [permanent, setPermanent] = useState(false);
+  const [success, setSuccess] = useState("");
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/ThrottlingRate.tsx:34]: We should add a type here to indicate the types of possible success states

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/ThrottlingRate.tsx
Comment (thiessenp-cds) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/ThrottlingRate.tsx:75]: I went back and forth on naming this schedule, expiry, futureDate.. I'll go with `expiry` then

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/ThrottlingRate.tsx
```diff
@@ -0,0 +1,220 @@
...
+        }
+      } catch (error) {
+        toast.error(t("throttling.error"));
+      } finally {
+        setLoadedSetting(true);
+      }
+    };
+    setLoadedSetting(false);
+    getThrottlingSetting();
+  }, [formId, t]);
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/ThrottlingRate.tsx:119]: Let's set the translation outside of the useEffect so we don't have `t` as a dep --- I know we have done this before but we should try to avoid.

timarney:
We should fix this jumping --- prob don't need the spinner but up to you
https://github.com/user-attachments/assets/070d16b6-5640-452d-9daa-65989067e15d

timarney:
Checking the `Permanently increase throttling rate` and than unchecking should wipe the existing value if there is one
https://github.com/user-attachments/assets/631c5d56-dcc1-40a5-b3a4-d79462ca4788

thiessenp-cds:
> We should fix this jumping --- prob don't need the spinner but up to you https://github.com/user-attachments/assets/070d16b6-5640-452d-9daa-65989067e15d

I've seen that on localhost. I guess for users with fast connections any spinner/loader would have that interaction. So we'd want to fix *all* loaders an spinners. That's a PR in itself IMO. What do you think?

thiessenp-cds:
> Checking the `Permanently increase throttling rate` and than unchecking should wipe the existing value if there is one https://github.com/user-attachments/assets/631c5d56-dcc1-40a5-b3a4-d79462ca4788
> 
> The date should also reflect the saved value when unchecking (without saving)

I thought about this and didn't think it was worthwhile. But I guess if someone accidentally checks permanent then their old date would be lost. Could be unexpected. Sure, I'll add that

timarney (approved):

thiessenp-cds:
Future work could be to update the button loader to avoid the spinner loading "flash" that is especially visible when an element is beside the button loader.

Some fixes could include: 
- only showing the loader after a certain amount of time
- adding a width in the button to compensate for the spinner
- adding a margin beside the button that shows/hides depending on whether the spinner is visible

PR #4659, opened 2024-11-14
chore: Add view for API submissions available

timarney:
# Summary | Résumé

Adds Card view for when API submissions are available but haven't been retrieved.

For example 

A new draft form is created.
A few submissions are submitted.

At this point nothing is confirmed.

Previously the no responses available screen would show --- which isn't accurate they just haven't been retrieved via the API to show.


thiessenp-cds (approved):
LGTM :)

PR #4672, opened 2024-11-19
fix: Updates duplicate element to use numbers

thiessenp-cds:
# Summary | Résumé

Instead of using the hard coded "copy" text, a numerical count was used to keep track of duplicated rich text elements in the form builder.

Another option would have been to pass to the related function the "copy" or "duplicate" text in both languages. Though "duplicate" is probably not ideal since that would require a gender in French.

Screenshot 2024-11-19 at 3 42 07 PM

### DIFF
--- a/lib/store/useTemplateStore.tsx
+++ b/lib/store/useTemplateStore.tsx
@@ -371,9 +371,12 @@ const createTemplateStore = (initProps?: Partial<InitialTemplateStoreProps>) => 
                 const element = JSON.parse(JSON.stringify(state.form.elements[elIndex]));
                 element.id = id;
                 if (element.type !== "richText") {
-                  element.properties[state.localizeField("title")] = `${
-                    element.properties[state.localizeField("title")]
-                  } copy`;
+                  const regExRemovePastCount = /\s\(\d{1,3}\)$/;
+                  const title = element.properties[state.localizeField("title")].replace(
+                    regExRemovePastCount,
+                    ""
+                  );
+                  element.properties[state.localizeField("title")] = `${title} (${elIndex + 1})`;
                 }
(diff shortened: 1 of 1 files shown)


i18n/translations/fr/form-builder.json
```diff
@@ -621,6 +621,7 @@
   "selectAutocompleteHint": "Cette option est utile lorsque les personnes qui remplissent le formulaire fournissent des informations personnelles telles que leur adresse, courriel ou numéro de téléphone.",
   "selectElement": "Sélectionnez un type d'élément",
   "send": "Envoyer",
+  "copy": "[FR]copy",
```
thiessenp-cds, line 624: @anikbrazeau this string is used in the title text when an element is duplicated e.g. "My element copy"

lib/store/useTemplateStore.tsx
```diff
@@ -363,17 +363,18 @@ const createTemplateStore = (initProps?: Partial<InitialTemplateStoreProps>) =>
               const elIndex = get().form.elements.findIndex((el) => el.id === elId);
               return get().form.elements[elIndex]?.properties.choices?.[choiceIndex];
             },
-            duplicateElement: (itemId, groupId = "") => {
+            duplicateElement: (itemId, groupId = "", copyEn = "", copyFr = "") => {
```
timarney, line 366: Wondering if we should pass the full title in...

The element that's calling duplicate already knows it's own title so we can clear out the work from the template store. 



duplicateElement: (itemId, groupId = "", titleEn = "", titleFr = "") => {

thiessenp-cds: In the template store the element doesn't have the latest id/val which makes it difficult to do the lookup for the element there.

timarney (approved):

Issue #4287, opened 2024-09-18
Migration index


srtalbot:

anikbrazeau:
Related issue: https://github.com/cds-snc/platform-forms-client/issues/3051

craigzour:
Deployment plan has been updated after I discovered a scenario that was not originally taken into account. @srtalbot @Abi-Nada

srtalbot:
Part 2 is now in prod!

On part 3 of the deployment plan. Testing before merging to staging.

craigzour:
Part 3 is now released in Production.

Will now be testing part 4 in Staging.

craigzour:
Part 4 migration script was successfully run in Production on Saturday (Nov 30th)

craigzour:
Added new step 5 that will need to be released before the old step 5 (which is now step 6).

craigzour:
Part 6 has been deployed to Production

craigzour:
Part 7 is currently running in Staging and should be ready to be released to Production.

craigzour:
Part 7 > Infra is now running in Production

craigzour:
Part 8 is now running in Staging and awaiting Production release

craigzour:
Part 8 is now running in Production.

The plan is now to wait for at least a week to see if everything works fine with the full integration of the new Index. After that we can proceed with Part 9 which will result in the old Index being completely deleted from DynamoDB.

craigzour:
Final part is now running in Staging. It will automatically be included in the next Production release.

PR #4678, opened 2024-11-20
chore: add migration script to add CreatedAt property to Vault CONF# items

craigzour:
# Summary | Résumé

Part 2 of https://github.com/cds-snc/platform-forms-client/issues/4287

- Create migration script to add `CreatedAt` property to existing Vault `CONF#` items. The value assigned to that property is taken from the original associated `NAME#` item.


bryan-robitaille (approved):
Tested and everything works as designed.

PR #4679, opened 2024-11-20
chore: add migration script to add StatusCreatedAt property to Vault NAME# items

craigzour:
# Summary | Résumé

Part 4 of https://github.com/cds-snc/platform-forms-client/issues/4287

- Adds or updates `Status#CreatedAt` property (using the existing Status and CreatedAt properties from the same item) in NAME# items.


bryan-robitaille (approved):
👍

PR #4681, opened 2024-11-20
fix: Fixes Review Page showing the "Go back" button twice

thiessenp-cds:
# Summary | Résumé

This bug happens when a form group has no "nextAction" in the JSON. The current logic will then show both the GroupsBackButton and the submit BackButton. This PR fixes this case where both back buttons will show at the same time.

Follow up: send support the updated (fixed) JSON: [give-feedback-2024-11-19 (updated).json](https://github.com/user-attachments/files/17907154/give-feedback-2024-11-19.updated.json)

NOTE:
When a form is in this state, the current logic would also not allow the form to Submit.
Currently the form-builder editor tree view should handle the case of automatically auto-flowing and fixing any pages not pointing to a next action. If we can learn how the user got the form into this state then we could try to debug it.
One idea is to add an additional validation when a form is being saved. The validation would check each group, and make sure it has next action, if it does not, then a minor error message could show.

## Test

If you load the original form JSON (below), the back button show only display once on the "Give feedback" page. Get to this page by selecting "yes" on the "Would you like to receive a response from us?" question.

[give-feedback-2024-11-19 (updated).json](https://github.com/user-attachments/files/17907154/give-feedback-2024-11-19.updated.json)

Screenshot 2024-11-25 at 12 54 47 PM (fix)

Note that this only fixes the two back buttons showing at once. The form is still in a broken state and will not be able to submit (if you click submit, nothing will happen).
Below is an updated (fixed) version of the form that will work correctly. (note how this version has a "continue" button and not a broken submit button)

[give-feedback-2024-11-19 (original).json](https://github.com/user-attachments/files/17907149/give-feedback-2024-11-19.original.json)

Screenshot 2024-11-25 at 1 11 19 PM


Abi-Nada:
QA tested groups and non-group forms to ensure all works well, and LGTM

timarney (approved):

PR #4684, opened 2024-11-21
chore: review page refactor

thiessenp-cds:
# Summary | Résumé

Refactor notes
- simplified data structure and allowed sub components to handle "printing" themselves (vs one big DS..)
- (CTD) components "printing" themselves gives a lot of flexibility for customizing layouts 
- moved testable/shareable code into helper (utils) files
- fixed a number of existing bugs that were difficult to spot

Other Quick UX Fixes
- [x] remove the dashes that appear for page text
- [x] If a page doesn’t contain questions, do not show it
- [x] checkboxes & attestation, put options on separate line, not comma separated
- [x] Present address the same regardless on the output selection - as split
- [x] Repeating sets add title and use instance name as label

Followup PRs beyond the scope of this PR
- [ ] "Edit" button updated to. "Go to {page} to edit" 
- [ ] pages with no questions (no input) to say "Go to page", since there is no editing


components/clientComponents/forms/Review/reviewUtils.ts
```diff
@@ -0,0 +1,208 @@
+import { AddressComponents, FileInputResponse, FormElement, FormElementTypes } from "@lib/types";
```
thiessenp-cds, line 1: Note to self: the goal of this utils file is to use as a placeholder and in a followup PR move much/most of the below to general utils. Ideally this file will be deleted in the future.
The below code was used in ways it was never mean to be and the original should have been refactored at that time IMO.

components/clientComponents/forms/Review/EditButton.tsx
```diff
@@ -0,0 +1,32 @@
+import { useGCFormsContext } from "@lib/hooks/useGCFormContext";
+import { Button } from "@clientComponents/globals";
+import { Theme } from "@clientComponents/globals/Buttons/themes";
+import { ReviewItem } from "./reviewUtils";
+
+export const EditButton = ({
+  reviewItem,
```
timarney, line 7: Can we just pass the review item id here --- looks like that's all we're using.

components/clientComponents/forms/Review/EditButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+}: {
+  reviewItem: ReviewItem;
+  theme: Theme;
+  children: React.ReactElement | string;
+  onClick?: () => void;
+}): React.ReactElement => {
+  const { setGroup, clearHistoryAfterId } = useGCFormsContext();
+  return (
+    <Button
+      type="button"
```
timarney, line 20: This already default to button

components/clientComponents/forms/Review/QuestionsAnswersList.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { ReviewElement, ReviewItem } from "./reviewUtils";
+
+export const QuestionsAnswersList = ({
+  reviewItem,
+}: {
+  reviewItem: ReviewItem;
+}): React.ReactElement => {
+  return (
+    <dl className="my-10">
+      {Array.isArray(reviewItem.elements) &&
```
timarney, line 12: We should early return a null above the `dl` here if we're not working with an array ---  doesn't make sense to me to output an empty `<dl`

components/clientComponents/forms/Review/QuestionsAnswersList.tsx
```diff
@@ -0,0 +1,28 @@
...
+        reviewItem.elements.map((reviewElement) => {
+          if (Array.isArray(reviewElement.values)) {
+            return (
+              <SubElements key={randomId()} elements={reviewElement.values as ReviewElement[]} />
```
timarney, line 16: Why not use the review item id here vs random?

components/clientComponents/forms/Review/QuestionsAnswersList.tsx
```diff
@@ -0,0 +1,28 @@
...
+            );
+          }
+          return (
+            <div key={randomId()} className="mb-8">
```
timarney, line 20: Same here why a random id?

components/clientComponents/forms/Review/reviewUtils.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    // Handle any Sub Elements. Note Sub Elements = Dynamic Rows = Repeating Sets
+    if (element?.type === FormElementTypes.dynamicRow) {
+      resultValues = [];
+      const parentId = element.id;
+      const parentTitle = element.properties?.[getLocalizedProperty("title", lang)];
+      const subElements = element.properties?.subElements;
+      // Use FormValues as the source of truth and for each FormValue value, map the related
+      // subElement title to the FormValue value
+      const subElementValues = (formValues[parentId] as string[]).map(
```
timarney, line 99: Wondering if we can break this up ---

We have 2 maps inside of a map

components/clientComponents/forms/Review/reviewUtils.ts
```diff
@@ -0,0 +1,208 @@
...
+
+        const addressFormValue = formValues[elementId] as string;
+        const addressValues = JSON.parse(addressFormValue) as AddressElements;
+        const canadaOnly = element.properties?.addressComponents?.canadianOnly;
+
+        const titleSet = {
+          streetAddress: parentTitle + " - " + addressCompleteStrings.streetAddress,
+          city: parentTitle + " - " + addressCompleteStrings.city,
+          province:
+            parentTitle +
```
timarney, line 143: If we make these template string does it make it easier / less lines of code ?

thiessenp-cds: Yup sounds good. My hope is to move this code out and for it to be used as more of a customization. (future refactor) I never intended the code to be used this way.

components/clientComponents/forms/Review/reviewUtils.ts
thiessenp-cds: Peter decides to add a 3 map in the maps because he likes maps in maps. (just kidding :)

components/clientComponents/forms/Review/QuestionsAnswersList.tsx
thiessenp-cds: I ran into cases where the Id between the element(s) and sub element(s) would be the same - messing up reacts "caching". Why the randomId()

components/clientComponents/forms/Review/FormElements/BaseElementArray.tsx
```diff
@@ -0,0 +1,36 @@
...
+}): React.ReactElement => {
+  if (!formItem) {
+    return <></>;
+  }
+
+  return (
+    <div className="mb-8">
+      <dt className="mb-2 font-bold">{formItem.label}</dt>
+      <dd>
+        {Array.isArray(formItem.values) ? (
```
timarney, line 18: Should we catch this higher up ?

  if (!formItem || !Array.isArray(formItem.values)) {

timarney:
For repeatings sets can we use the customized name?

Screenshot 2024-12-23 at 1 16 29 PM

timarney (approved):

PR #4691, opened 2024-11-21
feat: replace old Status index with new StatusCreatedAt one

craigzour:
# Summary | Résumé

Part 5 of https://github.com/cds-snc/platform-forms-client/issues/4287

- Replaces use of old `Status` index with new `StatusCreatedAt` one
- Loads the `RemovalDate` piece of information for each submission listed under `Signed off for removal` tab in an asynchronous way. This is due to the new index not returning the `RemovalDate` data in its projected attributes.

# Tests

- On the response page (in all tabs): make sure you can still see all form submissions you were getting before testing the new index. The only difference should be that items are now ordered by `CreatedAt` date. It means that the oldest response should be at the top of the list.
- On the response page (tab: Signed off for removal): items should asynchronously load their "Next step" piece of information.

Testing options:
- If you want to keep your existing data then yes you will need to deploy the latest infra version and run the two following migration scripts on your local DynamoDB Vault table
https://github.com/cds-snc/platform-forms-client/tree/main/utils/createdAtVaultConfItemsMigration
https://github.com/cds-snc/platform-forms-client/tree/main/utils/statusCreatedAtMigration
- If you want to test with a fresh set of data you can just deploy the new infra and create new form responses. In this case the old data scenario can be tested on Staging with the PR preview env.


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[[...statusFilter]]/components/NextStep.tsx
```diff
@@ -22,14 +24,44 @@ const ExclamationText = ({
...
+        setLabel(getRemovalByMessage(value));
+      } else {
+        setLabel("-");
+      }
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  if (label === null) {
+    return <Skeleton count={1} className="my-4 ml-4 w-[300px]" />;
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/responses/[[...statusFilter]]/components/NextStep.tsx:51]: 👍

timarney (dismissed):
Nice work on this + all the previous steps to get to this point.

thiessenp-cds (approved):
I tested the responses tabs and the submissions are in order.
Testing the download formats, the submissions could be out of order. I added a quick sort() call on that to make sure they'd be in order. Kind of out scope of this PR but nice to fix now when it's obvious.
Nice work! LGTM :)

Issue #4378, opened 2024-10-09
API V2.1.0 - Machine readable tags or IDs


srtalbot:

connorscarolyns:
@thiessenp-cds @timarney we'll definitely need to talk this one through together to better understand the problems we are trying to solve. Do you think we should include one of our prototype partners who expressed this need to get a better idea of their needs?

Abi-Nada:
Connect with @bryan-robitaille and @craigzour first on ideas to get the ball rolling

connorscarolyns:
Proposed approach:

Phase 1 - 
Made sure current question IDs are static and consistent when forms are downloaded and re-uploaded

Phase 2- 
Create a custom editable tag that users can add to their questions to enable consistency between different forms that use the same field types, like "Name"

@srtalbot @timarney @bryan-robitaille @Abi-Nada

connorscarolyns:
@samsadasivan will be consulted on design when interface changes are ready.

connorscarolyns:
@Abi-Nada should we split this into two tickets? one for the static ID work and one for the editable tags?

timarney:
I think we should look at the `editable tags` first as that might help guide the path for the static ID work.

Looking to outline both before we make schema updates.

cc: @samsadasivan @connorscarolyns

anikbrazeau:
Figma: https://www.figma.com/design/zlh8BVQyDczHvXo4xXUrb6/GC-Forms-API-interface?node-id=4766-37281&t=vowAaPR0FqR7LOHp-4

timarney:
- [ ] update element more menu to allow tag input (default to uuid)
- [ ] update template store to include tag / uuid


<hr>

- [ ] update increment id to store "next id"
- [ ] update template store to use "next id"


<hr>

- [ ]  test with repeating sets
- [ ] test with address complete

dsamojlenko:
Work is underway here: https://github.com/cds-snc/platform-forms-client/pull/4694

connorscarolyns:
One use case I'm aware of is to provide a way to standardize questions across forms so that if different forms are going into one system, all first names, last names, and phone numbers can be scripted once to land in the right place without having to rewire each one individually.

anikbrazeau:
As a form builder, I want to change the form question content and the order of the form without breaking the API integration, so that I can keep tinkering with the form even once the API is set up.

[...]

PR #4694, opened 2024-11-21
feat: Configurable question tags

dsamojlenko:
# Summary | Résumé

#4378 "Machine readable tags or IDs":
- Adds a tag property to the form schema (under properties)
- Defaults this tag to a UUID when adding a new form element in the form-builder
- Adds a configuration component in MoreDialog to configure the tag name
- Modifies transformFormResponses to support tagged responses
- Adds test coverage for tagged responses

## About tagged responses
A tagged response has the following structure:
```
{
  tag: "...",
  answer: [string]
}
```

The answer string can sometimes be a serialized object - in the cases of FormattedDate and AddressComplete, they will be serialized JSON. Checkboxes also render as a serialized object/array of values.

When responses are downloaded as HTML or CSV they pass through the fetchSubmissions server action, where the response will be extracted and returned without the tag.

When responses are downloaded from the API, they will be returned following the object structure above.

If a form was created before this feature rolls out, but modified after- pre-existing inputs will not have tags, any newly added inputs will have tags. Tags can be added to pre-existing inputs by opening the MoreDialog, editing the tag, and saving.

If a form has a mix of tagged and untagged inputs, the responses to that form will have mixed tagged/untagged responses.


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/components/dialogs/MoreDialog/QuestionTagOptions.tsx
```diff
@@ -0,0 +1,52 @@
...
+              ...item.properties,
+              tag: e.target.value,
+            },
+          });
+        }}
+      />
+      <div>
+        <InfoDetails summary={t("How tags can help")}>
+          <div className="mb-8 mt-4 border-l-3 border-gray-500 pl-8">
+            <p className="mb-4 text-sm">{t("text description")}</p>
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/components/dialogs/MoreDialog/QuestionTagOptions.tsx:46]: Need a proper description here

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/components/dialogs/MoreDialog/QuestionTagOptions.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  if (item.type === FormElementTypes.richText) {
+    return null;
+  }
+
+  return (
+    <section className="mb-4 mt-8">
+      <Label htmlFor="">{t("Question tag")}</Label>
+      <Hint>
+        {t("Adding a question tag allows the question to contain a permanent identifier.")}
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/components/dialogs/MoreDialog/QuestionTagOptions.tsx:26]: Need proper i18n here

timarney:
@bryan-robitaille --- here's the "Configurable question tags" PR for you to look at -- at some point.

PR #4704, opened 2024-11-25
fix: update to stop Jaws screen reader re-announcing the form label on every input change

thiessenp-cds:
# Summary | Résumé

This update fixes an error where the Jaws screen reader would re-announce a form label on every input change. The ticket mentioned a text input but the same error was on a textarea, search dropdown,.. This error did not happen on the Support form so I suspect it is Form-Forms only. Also this behaviour only happens in Windows+Jaws and not in MacOS+VoiceOver or Windows+NVDA.

The fix removes a top level live-region on the main form element. I suspect this was placed there as a temporary fix and later noted that it should be removed. The tricky part with this is it is difficult to say whether any dynamic updates on a form still depend on this live-region.  My intuition would be none but this should be tested.

_Note: A deeper problem is why a text input change is causing a DOM update on every key stroke? I can see in the React developer tools that it looks like the form itself is re-rendered on each input change. I suspect we need to take a look into how we use context/providers and update form values. This would probably be a big lift._

## Test

Anyone willing to QA, try a few forms running VoiceOver and verify any dynamic updates are announced.

Peter can test Windows+Edge+Jaws in his VM.

### DIFF
--- a/components/clientComponents/forms/Form/Form.tsx
+++ b/components/clientComponents/forms/Form/Form.tsx
@@ -251,8 +251,6 @@ const InnerForm: React.FC<InnerFormProps> = (props) => {
               handleSubmit(e);
             }}
             noValidate
-            // TODO move this to each child container but that I think will take some thought.
-            aria-live="polite"
           >
             {isGroupsCheck &&
               isShowReviewPage &&


timarney (approved):

craigzour (approved):
I tested your PR. It appears to not be saying "You are currently on a text area" every time I type a new character (which is what is happening without your fix).

Let me know if this is what needed to be confirmed.

PR #4707, opened 2024-11-26
refactor: replace use of Status with new Status#CreatedAt attribute when requesting Vault items

craigzour:
# Summary | Résumé

Part 8 of https://github.com/cds-snc/platform-forms-client/issues/4287

- Replaces any reference to the old `Status` with new `Status#CreatedAt` attribute when requesting Vault items.


timarney (dismissed):
Tested on PR review env.

PR #4708, opened 2024-11-26
fix: for Radio button / group unexpected behaviour

thiessenp-cds:
# Summary | Résumé

A radio/checkbox group does not need a tabindex on a container, this can cause unpredictable behavior in assistive technologies.

### DIFF
--- a/components/clientComponents/forms/FormGroup/FormGroup.tsx
+++ b/components/clientComponents/forms/FormGroup/FormGroup.tsx
@@ -22,8 +22,6 @@ export const FormGroup = (props: FormGroupProps): React.ReactElement => {
       data-testid="formGroup"
       className={classes}
       aria-describedby={ariaDescribedBy}
-      // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
-      tabIndex={0}
     >
       {children}
     </fieldset>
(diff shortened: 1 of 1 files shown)


components/clientComponents/forms/FormGroup/FormGroup.tsx
```diff
@@ -22,8 +22,6 @@ export const FormGroup = (props: FormGroupProps): React.ReactElement => {
       data-testid="formGroup"
       className={classes}
       aria-describedby={ariaDescribedBy}
-      // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
-      tabIndex={0}
```
timarney, line 26: Any idea why this was here to begin with?

thiessenp-cds: Good question. On further testing I can see this was used by the form validation skip ahead links. The problem with tabindex=0 is it is then in the tab order which is confusing for assistive technology users because tabbing to a radio group will announce it etc. but the user is not really in the radio group yet. A nice compromise is to use tabindex=-1 so the form group is still programmatically focussable and it is not in the natural tab order.

timarney (approved):

PR #4709, opened 2024-11-27
chore: Add ability for gc forms admin to switch to api mode for live form

timarney:
# Summary | Résumé

Adds admin screen to allow generating an Api key for a live form.

Noting this will show for user with Manage all forms + a published form.

Screenshot 2024-11-27 at 10 20 22 AM


thiessenp-cds (approved):
LGTM :)

PR #4763, opened 2024-12-03
Reapply "feat: replace old Status index with new StatusCreatedAt one (#4691)" (#4761)

craigzour:
# Summary | Résumé

Part 6 of https://github.com/cds-snc/platform-forms-client/issues/4287

- Replaces use of old `Status` index with new `StatusCreatedAt` one
- Loads the `RemovalDate` piece of information for each submission listed under `Signed off for removal` tab in an asynchronous way. This is due to the new index not returning the `RemovalDate` data in its projected attributes.

# Tests

- On the response page (in all tabs): make sure you can still see all form submissions you were getting before testing the new index. The only difference should be that items are now ordered by `CreatedAt` date. It means that the oldest response should be at the top of the list.
- On the response page (tab: Signed off for removal): items should asynchronously load their "Next step" piece of information.

Testing options:
- If you want to keep your existing data then yes you will need to deploy the latest infra version and run the two following migration scripts on your local DynamoDB Vault table
https://github.com/cds-snc/platform-forms-client/tree/main/utils/createdAtVaultConfItemsMigration
https://github.com/cds-snc/platform-forms-client/tree/main/utils/statusCreatedAtMigration
- If you want to test with a fresh set of data you can just deploy the new infra and create new form responses. In this case the old data scenario can be tested on Staging with the PR preview env.


thiessenp-cds (approved):
Re-test and works as expected. LGTM :)

PR #4766, opened 2024-12-03
fix: followup to screen reader re-announcing the form label on every input change

thiessenp-cds:
# Summary | Résumé

This is a followup PR to the previous fix #4704. The form live-region is added back and to stop the "noisy" form labels, a live=off is added around them.

TL/DR: Should work fine.
This appears to work well with screen readers. It is generally a best practice to have any attributes like live-regions rendered as close as possible to the original rendered HTML. This is because if the DOM becomes complex, there is a good chance an AT will ignore any dynamic changes in the DOM unless told about them. But in this case, a live-region is added on the form element that is in the original-ish HTML. So the AT is paying attention to this element. A few children below are the dynamically added show-hide elements with labels that have an overriding live=off.  AT should pick these up and AT are getting better and better at noticing dynamic DOM changes.
Also the form validation elements are outside the label and won't be impacted.

A follow up question/investigation could be looking into why a form label is re-rendered when the related input is updated. The label and input are siblings (not nested in each other) so it is not related to a parent component re-rendering. Or at least not in an obvious way. I suspect this has something to do with Formik. I left this as a TODO after spending a few hours looking into it.


timarney (approved):
Let's merge so we can further test with the a11y folks.

PR #4770, opened 2024-12-03
chore: update add / remove sub choices for repeating sets

timarney:
# Summary | Résumé

Updates Sub options to use id vs index for adding / removing sub elements

Seeing an issue on staging ---doesn't seem to happen on local dev but was really slow.
Screenshot 2024-12-03 at 1 55 40 PM


thiessenp-cds (approved):
LGTM :)

PR #4795, opened 2024-12-06
chore: add path helper to template store

timarney:
# Summary | Résumé

Updates template store to include path helper.

The template store already has access to the `elements` -- with this update we can let the template store do the work and avoid pulling the elements into place where we just use them for a lookup.


thiessenp-cds (approved):
LGTM :)

PR #4799, opened 2024-12-09
chore: move store setup to standalone function

timarney:
# Summary | Résumé

Moves init store out to a standalone function to start to cut down the size of `useTemplateStore`.


thiessenp-cds (approved):
LGTM :)

PR #4802, opened 2024-12-10
chore: add move helpers

timarney:
# Summary | Résumé

Breaks move up and move down out into helper methods.


thiessenp-cds (approved):
LGTM :)

PR #4803, opened 2024-12-10
chore: move template storage options

timarney:
# Summary | Résumé

Move template store options file to standalone function.


thiessenp-cds (approved):
LGTM :)

PR #4811, opened 2024-12-10
chore: Add temporary session storage for form progress

timarney:
# Summary | Résumé

https://github.com/user-attachments/assets/0c9859ed-7e9d-4e93-a3b4-eecc200836a1

Update to handle missing Sever Action ID error

- Detects the error
- Saves currently filled data to a browser session
- User receives an error prompting them to reload and re-submit.
- Reloads the form when error occurs --- at this point the new server id should exist
- Saved data is restored 
- Saved data is cleared

## Notes

Partially based on original PR

https://github.com/cds-snc/platform-forms-client/pull/1557


components/clientComponents/forms/Form/Form.tsx
```diff
@@ -405,15 +416,22 @@ export const Form = withFormik<FormProps, Responses>({
       if (result.error) {
         if (result.error.message.includes("FileValidationResult")) {
           formikBag.setStatus("FileError");
+        } else if (result.error.message.includes("Failed to find Server Action")) {
+          throw new Error("Failed to find Server Action");
```
timarney, line 420: @dsamojlenko we can likely just re-throw vs creating a "new" error here

Thoughts?

dsamojlenko (dismissed):

bryan-robitaille (approved):
Nice work!

PR #4812, opened 2024-12-10
chore: break out save for purpose option

timarney:
# Summary | Résumé

Fixes small issue currently on staging server where the Delivery option was trying to be updated when saving a form purpose.

This PR splits the saves 1 for the Delivery option and 1 for the form purpose.


bryan-robitaille:
Should both save buttons become disabled or only one?
delivery_option

timarney:
If published both 

For the top --- if no changes have been made.

bryan-robitaille (approved):
Nice 👍

PR #4813, opened 2024-12-10
chore: save purpose option to template store

timarney:
# Summary | Résumé

Fixes issue where most recent saved value was not saved back to the template store.


craigzour (approved):
Confirmed through PR env that the bug is now fixed.

PR #4829, opened 2024-12-12
feature: In memory cached Ability and Authenticated Action wrapper.

bryan-robitaille:
# Summary | Résumé

## Authenticated Action
A new wrapper function that takes an async function as an argument but first ensures a session is present.  It replaces the common pattern of using `authCheckAndThrow`.  Similar to all the new functions introduced in the Privilege Refactor this change will be brought in incrementally.
```js
export const AuthenticatedAction = <Input, Output>(action: (args: Input) => Promise<Output>) => {
  return async (...args: unknown[]) => {
    const session = await auth();
    if (session === null) {
      const language = await getCurrentLanguage();
      redirect(`/${language}/auth/login`);
    }
    return action(...(args as Parameters<typeof action>));
  };
};
```

## InMemoryCache
A new Last Recently Used (LRU) cache implementation that can be used in the application.  The global ability cache is an instance of the InMemoryCache.  A regular cache like Redis could not be used to cache the ability as the ability contains properties that cannot be serialized.

```js
export class InMemoryCache {
  private cache: Map<string, { value: unknown; expireAt: number }>;
  private defaultTTL: number;
  private maxSize: number;
  constructor(maxSize = 200, defaultTTL = 60) {
    // default TTL is 60 seconds
    this.cache = new Map();
    this.defaultTTL = defaultTTL;
    this.maxSize = maxSize;
  }

  private staticSize() {
    while (this.cache.size >= this.maxSize) {
      const iterator = this.cache.entries();
      this.cache.delete(iterator.next().value[0]);
    }
  }

  public set(key: string, value: unknown, ttl = this.defaultTTL) {
    const now = Date.now();
    // convert ttl to milliseconds
    const expireAt = now + ttl * 1000;

    this.staticSize();

    this.cache.set(key, { value, expireAt });

    setTimeout(() => {
      const val = this.cache.get(key);
      if (val && val.expireAt <= Date.now()) {
        this.cache.delete(key);
      }
    }, ttl);
  }

  public get(key: string) {
    const now = Date.now();
    const cachedItem = this.cache.get(key);

    if (cachedItem && cachedItem.expireAt > now) {
      return cachedItem.value;
    } else {
      this.cache.delete(key);
      return null;
    }
  }

  public clear() {
    this.cache.clear();
  }

  public cleanup() {
    const now = Date.now();
    for (const key in this.cache) {
      const cachedItem = this.cache.get(key);
      if (cachedItem && cachedItem.expireAt <= now) {
        this.cache.delete(key);
      }
    }
  }
}
```

## Cached `ability`
The `ability` object can now be retrieved globally server side with the `getAbility` function.  The function is responsible to create and cache the ability object for a user session.  This increases performance by not computing the expensive `createMongoAbility` on every invocation  where `ability` is required.

```js
// Maximum number of items to store in the cache: 100
// TTL for items in the cache: 1 minutes
const abilityCache = new InMemoryCache(100, 60);

export const getAbility = async (): Promise<UserAbility> => {
  const session = await auth();
  if (!session) throw new Error("No session found");
  // If the privileges are cached, we can use them as a check that no changes have been made
  // and it is safe to use the cached ability
  const cachedPrivilegesRules = await privilegeCheck(session.user.id);
  const cachedAbility = abilityCache.get(session.user.id);
  if (cachedAbility && cachedPrivilegesRules !== null) {
    logMessage.debug(`Using cached ability for user ${session.user.id}`);
    return cachedAbility as UserAbility;
  } else {
    logMessage.debug(`Creating new ability for user ${session.user.id}`);
    const ability = createAbility(session);
    abilityCache.set(session.user.id, ability);
    return ability;
  }
};
```

## Remove requirement to pass `ability` to CRUD layer functions.
One of the major achievements of using a cached ability is that it removes the requirement of creating the ability at the entry point (server action) and passing it through all sub function invocations to the CRUD layer.  Fetching the ability for privileges checks is now handled directly within the `privilege` library when `authorization` helpers are called.

### Previously
```js
export const deleteForm = async (id: string) => {
  await deleteTemplate(ability, id);
// ...
};

export async function deleteTemplate(
  ability: UserAbility,
  formID: string
): Promise<FormRecord | null> {
    await authorization.canDeleteForm(ability, formID);
// ...
};
```
### Now
```js
export const deleteForm = AuthenticatedAction( async (id: string) => {
  await deleteTemplate(id);
// ...
});

export async function deleteTemplate(
  formID: string
): Promise<FormRecord | null> {
    await authorization.canDeleteForm(formID);
// ...
};
```

## Get current language utility helper
New utility function in i18n server lib that can provide the current user language.  Useful for creating redirects from server side.
```js
export async function getCurrentLanguage() {
  const path = headers().get("x-path") ?? "";
  const pathLang = pathLanguageDetection(path, languages);
  const cookieLang = cookies().get("i18next")?.value;
  return pathLang || cookieLang || languages[0];
}
```

## Updates `AccessControlError` to add properties to fetch user ID
Adds a method on the `AccessControlError` class to fetch the user ID from the current ability.  It is useful when creating Access Denied audit log events.  The `logEvent` function has also been updated to accept either a `string` or `Promise<string>` as the userId argument.

```js
export class AccessControlError extends Error {
  public user: {
    id: Promise<string>;
  } = {
    id: getAbility()
      .then((ability) => ability.userID)
      .catch(() => "unauthenticated"),
  };

  constructor(message: string = "AccessControlError") {
    super(message);
    Object.setPrototypeOf(this, AccessControlError.prototype);
  }
}
```

### Example Usage
```js
  // Can a user view any Template
    const { user } = await authorization.canViewAllForms().catch((e) => {
      logEvent(e.user.id, { type: "Form" }, "AccessDenied", "Attempted to access All System Forms");
      throw e;
    });
```

## Updates template lib functions and invocations 
This PR implements the above functionality for the `templates` lib and it's subsequent invocations throughout the codebase.  Future PRs will cover the remaining libraries.

Ensure all template functionalities behave properly.  
A user can create, update, and delete a template.

All other libraries were out of scope for this initial implementation

# Pull Request Checklist


lib/auth/index.ts
```diff
@@ -41,3 +41,17 @@ export {
...
 
 export { GET, POST, auth, signIn, signOut } from "./nextAuth";
+
+import { getAbility } from "@lib/privileges";
+
+// Creates a new custom Error Class
+export class AccessControlError extends Error {
+  public user: {
+    id: Promise<string>;
+  } = { id: getAbility().then((ability) => ability.userID) };
```
craigzour, line 51: From what I am seeing in the review process (so far), we could have simplified the `AccessControlError` object by passing the user identifier in the constructor. That would avoid having to deal with a promise.

Maybe there is one case (I did not see yet) that forced you to do it this way though.

bryan-robitaille: You're exactly right, my initial implementation did have the userID passed in as an arg to the constructor.  The issue was with the current implementation of AccessControlError across the app.  I'm hoping once the whole app is converted it'll be able to be simplified.

timarney:
Looks like you'll need a small adjustment for Next 15 to use async headers

https://nextjs.org/docs/app/api-reference/functions/headers

Screenshot 2024-12-18 at 11 20 29 AM

timarney (approved):

craigzour (approved):
Great work :)

PR #4841, opened 2024-12-16
Fix: Pin to Node Alpine 3.20

bryan-robitaille:
# Summary | Résumé
Pinning to Node Alpine 3.20 until Prisma 6.1.0 is released and implemented.

Fixes SSL location issue between Prisma and  Node Alpine > 3.20

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:22-alpine
+FROM node:22-alpine3.20
 
 ARG USERNAME=node
 ARG USER_UID=1000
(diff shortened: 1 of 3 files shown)


timarney:
There's some more alpine references in the dev container docker file --- just for consistency. 

Screenshot 2024-12-16 at 10 51 29 AM

bryan-robitaille:
> There's some more alpine references in the dev container docker file --- just for consistency.
Bah... no one uses dev containers ;)

I updated and pushed commit.

timarney (approved):

PR #4888, opened 2024-12-31
fix: Updates how a description is added to a checkbox or radio

thiessenp-cds:
# Summary | Résumé

Originally for any checkbox or radio, a description was repeated for each one. This fix removes the repeated description. This works because the description is linked from the container fieldset to the description using an aria-describedby. The original code did not break WCAG but was bad UX for screen reader users.

Note:
**This doesn't solve the other problem of very verbose page/group loads**. We need to think of a strategy how to announce adding dynamic groups of stuff to a form. I suspect we could update the page title and use the live region announce component or similar. But then we'd want to remove the main live=polite on the main form element and that would require a lot of testing and thinking about edge cases e.g. show-hide rules.

## Test

Look at the source code for a checkbox or radio with and without a description added using the form-builder. If there is a description, it should only appear once (not in each input=radio/checkbox)

### DIFF
--- a/components/clientComponents/forms/MultipleChoiceGroup/MultipleChoiceGroup.tsx
+++ b/components/clientComponents/forms/MultipleChoiceGroup/MultipleChoiceGroup.tsx
@@ -10,28 +10,16 @@ interface MultipleChoiceGroupProps extends InputFieldProps {
 }
 
 export const MultipleChoiceGroup = (props: MultipleChoiceGroupProps): React.ReactElement => {
-  const { className, choicesProps, type, ariaDescribedBy } = props;
+  const { className, choicesProps, type } = props;
 
   // field contains name, value, onChange, and other required Form attributes.
   const [field, meta] = useField(props);
 
   const choices = choicesProps.map((choice, index) => {
     return type == "checkbox" ? (
-      <Checkbox
-        {...choice}
-        key={index}
-        name={field.name}
-        className={className}
-        {...(ariaDescribedBy && { ariaDescribedBy: ariaDescribedBy })}
-      />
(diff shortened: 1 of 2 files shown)


components/clientComponents/forms/MultipleChoiceGroup/MultipleChoiceGroup.tsx
```diff
@@ -22,15 +22,15 @@ export const MultipleChoiceGroup = (props: MultipleChoiceGroupProps): React.Reac
         key={index}
         name={field.name}
         className={className}
-        ariaDescribedBy={ariaDescribedBy}
+        {...(ariaDescribedBy && { ariaDescribedBy: ariaDescribedBy })}
```
thiessenp-cds, line 25: I can't think of a case where the Radio or Checkbox input's would need an aria-describedby. The description of the input I would think would be in it's related label and for anything lengthy and generic in the fieldset's legend and or description. But I guess if a specific checkbox really needed a lot of explaining then maybe it could be used. But I wonder if that checkbox would be packed with too much info/context.
So I left it in to be used optionally for future cases I can't think of.

timarney, line 25: Should we just remove this ?

I don't see any code using the ariaDescribedBy for MultipleChoiceGroup

thiessenp-cds: Sure. On second thought, leaving that there may cause more harm than good.

timarney (approved):

PR #4897, opened 2025-01-02
fix: Updates number component to have more consistent UX

thiessenp-cds:
# Summary | Résumé

Updates the number and date components to no longer use input type=number for better UX.

Before

https://github.com/user-attachments/assets/8db075fb-a961-4379-91c9-d1e6156fcaa8

After

https://github.com/user-attachments/assets/09115e40-f1a3-4815-9ba9-65ca348ea7e2

## Background

There is no perfect solution for inputting numbers.

```
input type=number
Pros
- restricts only numbers entered by default
- announced as stepper (not number) still more descriptive than text
Cons
- UX issues https://tinyurl.com/2p9tm5vk
- AT quirks e.g. VO announce incorrectly beyond 10
- min-max values not enforced, only flagged as invalid through form validation

input type=text (+semantics title=number..)
Pros
- better UX https://tinyurl.com/2p9tm5vk
- can add specific semantics
- mobile keypay with inputmode=numeric
Cons
- type=text + semantics is a bit of a hack
- probably need/want a description e.g. https://design-system.service.gov.uk/patterns/dates/ 
- custom JS and semantics so must maintain and test
```

## Test

For both the number and date components:
- try entering characters, these should not be able to be entered
- try entering numbers, the complete number should be read out, so 111 as one hundred and eleven

Open a screen reader like VoiceOver
- tab into a number field, "number" should be read out along with the other info


timarney:
Looks like it's accepting text chars ?

Screenshot 2025-01-02 at 2 27 34 PM

thiessenp-cds:
> Looks like it's accepting text chars ?
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Constraint_validation
> 
> Screenshot 2025-01-02 at 2 27 34 PM

Form validation would fail if it's not a number like in your example. Do you think more immediate validation or constraints would be helpful?

timarney:
I would say we'll need a product decision on this one


cc: @Abi-Nada @srtalbot

thiessenp-cds:
> I would say we'll need a product decision on this one
> 
> cc: @Abi-Nada @srtalbot

I added a way to restrict user input to a number. Easy to remove if we decide not to include this.

connorscarolyns:
I think it works ok, It will be interesting to see how it tests. The screen reader announcing might be a bit confusing because it says it's a text field and to enter text, but nothing about it only accepting numbers.

connorscarolyns:
I also notice that the UK does not use the pattern attribute, and has added a spellcheck false. Do we have reasoning for or against these differences?

thiessenp-cds:
The pattern attribute is used by JavaScript to validate what’s entered as a number but I can move/hide this in the JavaScript.

We can add spellcheck though if we restrict to only numbers it may not have much value. Still doesn’t hurt.

I think that’s a good point that the input is not announcing itself as a number. So a user could try to enter characters that are not a number and have no idea why it wasn’t entered.

One solution would be to add an aria-label and override the input semantics. This way it would announce as a number.
E.g. without an aria-label

Screenshot 2025-01-08 at 10 46 24 AM


E.g. with an aria-label=”number” 
So the accessible name is changed to "number {label text}, edit text"

Screenshot 2025-01-08 at 10 46 47 AM


I wonder if the solution is becoming complex with the added JavaScript to restrict a number and aria to redefine an input text as a number? Vs. Just checking if an input is a number on submit and hopefully that input has hints it should be a number. I’m kind of on the fence.

thiessenp-cds:
@connorscarolyns actually what if we were to go ahead with the JavaScript to restrict entering to a number and adding an aria-label to identify the input as a number and see what happens in future user tests?

connorscarolyns:
I like your approach. You may be right about overcomplicating things, but user testing will show if there are issues. I'm fine with it how it is and will log any further concerns when we retest the next time.

timarney:
@thiessenp-cds further discussed with @srtalbot 

We're good to merge this when your back.

timarney (dismissed):

components/clientComponents/forms/FormattedDate/FormattedDate.tsx
```diff
@@ -147,12 +147,15 @@ export const FormattedDate = (props: FormattedDateProps): React.ReactElement =>
...
                 {t(`formattedDate.${part}`)}
               </label>
+              <div id={`${id}-description-month`} hidden>
+                {t("number")}
+              </div>
               <input
                 name={`${name}-${part}`}
                 id={`${name}-${part}`}
-                type="number"
-                min={1}
```
thiessenp-cds, line 154: The min and max here and below only had meaning with type=number. And even with type=number the min and max was only restricted when using the stepper with a mouse.

components/clientComponents/forms/FormattedDate/FormattedDate.tsx
```diff
@@ -147,12 +147,15 @@ export const FormattedDate = (props: FormattedDateProps): React.ReactElement =>
               <label className="mb-2" htmlFor={`${name}-${part}`}>
                 {t(`formattedDate.${part}`)}
               </label>
```
thiessenp-cds, line 149: The aria-described by with the "number" description may be overkill. I've never seen this done before on an input but it may help to identify the input as accepting numbers.  We could try with user testing.

connorscarolyns (approved):
Fields seem to be functioning well with VoiceOver. We'll validate these changes in our next Fable testing round. Most likely after the audit.

components/clientComponents/forms/TextInput/TextInput.tsx
```diff
@@ -59,17 +59,48 @@ export const TextInput = (
...
+        {...(type === "number" && {
+          // For mobile phones to switch the keypad to numeric
+          inputMode: "numeric",
+          "aria-describedby": `${id}-description-number`,
+          // Note: "onBeforeInput" and e.data could also be used but I'm not sure how cross-browser
+          // consistent it is
+          onKeyDown: (e) => {
+            // Allow control keys
+            if (
+              e.key.includes("Arrow") ||
```
timarney, line 90: Did you find any posts about a11y + using arrow keys with the number input ?

With a "browser" / html number input I can press and hold arrow up or down to increment decrement numbers.  Does removing this functionality cause issues ?

components/clientComponents/forms/FormattedDate/FormattedDate.cy.tsx
```diff
@@ -177,29 +177,4 @@ describe("<FormattedDate />", () => {
     cy.get("[data-testid=description]").should("exist");
     cy.get("[data-testid=description]").contains("This is a description");
   });
-
```
timarney, line 180: Why was this removed?

components/clientComponents/forms/FormattedDate/FormattedDate.tsx
```diff
@@ -184,16 +191,15 @@ export const FormattedDate = (props: FormattedDateProps): React.ReactElement =>
...
+              <div id={`${id}-description-day`} hidden>
+                {t("number")}
+              </div>
               <input
                 name={`${name}-${part}`}
                 id={`${name}-${part}`}
-                type="number"
-                min={1}
-                max={
```
timarney, line 192: Won't removing this cause issues?

components/clientComponents/forms/FormattedDate/FormattedDate.cy.tsx
thiessenp-cds: The test no longer applies because the max/min attributes were removed from the formatted date inputs

components/clientComponents/forms/FormattedDate/FormattedDate.tsx
thiessenp-cds: With a type=text the min/max doesn't really make sense. And even on a type=number the min/max only impact when increasing/decreasing a number using the stepper controls

components/clientComponents/forms/TextInput/TextInput.tsx
thiessenp-cds: Actually that was one of the UX issues with the input=number. Because the stepper is controlled by the mouse, if the input is focussed a user can easily accidentally scroll and increase/decrease the number without meaning to. 

The type=number is accessible and meets WCAG. It's just the UX issues e.g. https://tinyurl.com/2p9tm5vk

The people I respect in the accessibility community seem to be pretty neutral on this.

Though MDN recommends text unless you have a specific reason for the spin button integer entry:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#:~:text=as%20explained%20above).-,Accessibility,to%20numbers%20and%20associated%20characters.

components/clientComponents/forms/FormattedDate/FormattedDate.tsx
timarney: Let's make sure to get Dave's input on this part.

components/clientComponents/forms/FormattedDate/FormattedDate.cy.tsx
timarney: @thiessenp-cds Dave pushed a commit --- there is error handling in place when you submit.

Please open an issue to create a new Cypress test that will capture the error for a leap year either at the summary or field level.

timarney: Never mind looks like Dave does have some unit tests for this.

dsamojlenko (approved):

timarney (approved):

PR #4905, opened 2025-01-03
feat: handle dynamic content case by case

thiessenp-cds:
# Summary | Résumé

The current strategy for handling dynamic content in a form has been to use a parent live-region. While this technically works it makes dynamic updates for AT very verbose resulting in bad UX. For example, when a large show-hide block is shown, all the content is announced. Or if a dropdown is dynamically populated with a large list, the entire list is read out on load.

This solution handles each type of dynamic update on a case by case basis.

### Show-Hide

Uses the live message tool to announce when a sub element is activated. Because the sub form element could have exist on the current or another page, a generic message is used.
The shown element title is announced.

Note:
The disclosure pattern would be ideal for this situation but because of the impressive flexibility of show-hide conditionals, the pattern could not work e.g. activating a control that shows/hides a form element on a page that is not visible.

### Repeating Sets 

Uses the live message tool to announce when a Repeating Set is added. Because what is added can be any form element the message announced is generic.

Future Thought:
Should the removal of a Repeating Set be announced? The accessibility best practice is generally not to announce dynamic content removals. In this case though it may make sense.

### Dynamic Text

Both the Input and TextArea components had character length messages inserted dynamically along with their container div that had a live region. At best this will give unpredictable AT behavior. The fix was to have the container div with the live region in DOM long before the dynamic character message was added.

### Any Others / Cleanup?

A past hack on form labels with a live=off was removed because this PR should fix the original hack with a live=polite on the form container.  All hacks gone :)

## Testing

Try running VoiceOver on a form with a few form elements that have dynamic updates: e.g. show-hide, dynamic rows, form validation 

Here is a simple test showing show-hide and dynamic rows

https://github.com/user-attachments/assets/8e5d6824-232d-46c5-a509-1449b54fc5f0


components/clientComponents/forms/ConditionalWrapper/ConditionalWrapper.tsx
```diff
@@ -38,6 +40,8 @@ export const ConditionalWrapper = ({
   if (hasMatchedRule) {
     // Ensure rules for elements tied to the element are also matched.
     if (checkRelatedRulesAsBoolean(formRecord.form.elements, rules, matchedIds)) {
+      // Announce conditional shown. The check above ensures the element is on the current page
+      Event.fire(EventKeys.liveMessageObject, { key: "conditionalActivated", obj: element });
```
timarney, line 44: Do we need the entire element here or just the Title?

i18n/translations/en/common.json
```diff
@@ -111,7 +111,8 @@
   },
   "dynamicRow": {
     "add": "Add",
-    "delete": "Delete"
+    "delete": "Delete",
+    "addedMessage": "Added a copy of {{title}} ({{count}})"
```
timarney, line 115: Should this try to use rowTitle?

timarney:
Doing some initial testing with VoiceOver for repeating sets --- not hearing any announcement for  ` "addedMessage": "[FR]Adding {{rowLabelTitle}} - {{rowLabelCount}} for {{rowTitle}}"`

components/clientComponents/forms/ConditionalWrapper/Announce.tsx
```diff
@@ -0,0 +1,37 @@
...
+}: {
+  children: React.ReactNode;
+  element: FormElement;
+  lang: string;
+}) => {
+  const { t } = useTranslation();
+  const { Event } = useCustomEvent();
+  const messageRef = useRef("");
+
+  const maxTitleLength = 25;
```
timarney, line 22: Is this too low?

i18n/translations/fr/common.json
```diff
@@ -111,7 +111,8 @@
   },
   "dynamicRow": {
     "add": "Ajouter",
-    "delete": "Supprimer"
+    "delete": "Supprimer",
+    "addedMessage": "Fieldset {{rowTitle}} - {{count}} was added to the form"
```
timarney, line 115: We need the strings here before we can merge.

anikbrazeau, line 115:
```suggestion
    "delete": "Supprimer",
    "addedMessage": "Le champ {{rowTitle}} - {{count}} a été ajouté au formualaire"
```

i18n/translations/en/common.json
```diff
@@ -214,5 +215,8 @@
     "text2": "{{month}} {{day}}, {{year}} at {{hour}}:{{minute}} ET."
   },
   "cancel": "Cancel",
-  "number": "Number"
+  "number": "Number",
+  "conditional": {
+    "activated": "Field {{title}} was added to the form"
```
anikbrazeau, line 220: In French this has {{name}} rather than title? Is that intentional?

anikbrazeau, line 220:
```suggestion
    "activated": "The form element {{title}} was added to the form"
```

i18n/translations/fr/common.json
```diff
@@ -111,7 +111,9 @@
   },
   "dynamicRow": {
     "add": "Ajouter",
-    "delete": "Supprimer"
+    "delete": "Supprimer",
+    "addedMessage": "Le champ {{rowTitle}} - {{count}} a été ajouté au formulaire",
+    "removedMessage": "[FR]Fieldset {{rowTitle}} - {{count}} was removed from the form"
```
anikbrazeau, line 116:
```suggestion
    "removedMessage": "Le champ {{rowTitle}} - {{count}} a été supprimé du formulaire"
```

components/clientComponents/forms/DynamicRow/DynamicRow.tsx
```diff
@@ -60,13 +61,15 @@ export const DynamicGroup = (props: DynamicGroupProps): React.ReactElement => {
...
   const [hasReachedMaxNumberOfRows, setHasReachedMaxNumberOfRows] = useState<boolean>(false);
+  const { Event } = useCustomEvent();
 
   const { t } = useTranslation();
 
   useEffect(() => {
     if (focusedRow.current !== null) {
       try {
-        rowRefs.current[focusedRow.current].current?.focus();
+        // TODO: does focusing just the legend reduce verbosity? (fix other issue from Carolyn?)
```
timarney, line 71: Can we confirm / remove this todo ?

timarney:
Did some testing seems to be working quite well nice work.

Couple of questions:

When using a show / hide radio

The first time the field gets added it does announce 

- When I toggle it back off (swap the radio values) when I swap to a value that turns the field back on it does announce again --- it that what we're expecting?

- For the character count ... when I type and I exceed the amount the announcement of "you've exceeded" keeps getting cut / off (iterupted) by the value that was typed in the input --- did you notice the same and if yes anything we can do about that?

thiessenp-cds:
> Did some testing seems to be working quite well nice work.
> 
> Couple of questions:
> 
> When using a show / hide radio
> 
> The first time the field gets added it does announce
> 
>     * When I toggle it back off (swap the radio values) when I swap to a value that turns the field back on it does announce again --- it that what we're expecting?
> 
>     * For the character count ... when I type and I exceed the amount the announcement of "you've exceeded" keeps getting cut / off (iterupted) by the value that was typed in the input --- did you notice the same and if yes anything we can do about that?

For 1) As far as I know. In a show hide, the shown element is announced however many times it is hidden and later shown. This matches what is happening visually. @connorscarolyns does this sound right?

For 2) I'm not sure I understand. Is this related to this PR?

timarney (approved):

timarney (approved):

PR #4938, opened 2025-01-08
fix: Fix case of an empty formatted date on a groups page

thiessenp-cds:
# Summary | Résumé

Fix case of an empty formatted date on a groups page

## Test

Create a form. On the start page add a formatted date. Also add a formatted date to a page.

Navigate to the form and in both cases leave the date inputs empty and navigate to the Review page. The formatted date element should be listed with a "-".

Do the same but with date inputs filled in. The dates should show as expected on the Review page


timarney (approved):
Let's update the description with the ticket # so we can track this back.

thiessenp-cds:
Support Tickets
-  #20758
- #20755

PR #4947, opened 2025-01-09
bug: ensure array is sent for checkbox elements when they are in a hidden state

timarney:
# Summary | Résumé

Update to ensure when emptying out values the emptied value remains the same type i.e. array or string of the incoming value.

## Testing

- Create 2 checkbox lists 
- Make the second one a show hide based on one of the values in the first.
- Toggle the second list to show and set a value
- Toggle the second list to hide ... in this case the value should be "empty" when sending the payload as it's hidden,
- Ensure the value sent is an array for the checkbox type


thiessenp-cds (approved):
Works as expected. LGTM :)

craigzour (approved):

Issue #4865, opened 2024-12-19
Repeating set changes not being saved [https://assistance.cds-snc.ca/a/tickets/20618]


mo-cds:
## Description
When user tries to save changes to repeating set, the changes don't get saved.
https://assistance.cds-snc.ca/a/tickets/20618

## Steps to reproduce
1. Go to 'Q3A Allergies'
2. Click on 'Customize repeating set'
3. Edit the text fields.
4. Click save at the end of the screen
5. Click "Save Draft" at the top of the form.
6. Test the new changes 
7. See error

## Details
- **Browser**: Chrome
- **Operating system**: Mac, Windows
- **Environment**: Production
- **Language**: English, French
- **Viewport**: any screen size
- **Output**:[detailed-health-questionaire-testing-2024-12-18 (2).json.zip](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMGcyQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b7c6cf85116d0b6eb21cda2d3cde85c2c5af1148/detailed-health-questionaire-testing-2024-12-18%20%282%29.json.zip)

## Expected behaviour
It's expected for the changes to be saved after modifying the repeating set.

## Screenshots or videos
(https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBencyQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0fbe1490519e07d2234cc7ff6edfe4d252193d49/Recording-20241219_115410.webm)

PR #4967, opened 2025-01-13
fix: Repeating set changes not being saved

timarney:
# Summary | Résumé

Fixes: https://github.com/cds-snc/platform-forms-client/issues/4865


Abi-Nada:
QA Tested on the client's form, and the fix works!

bryan-robitaille (approved):

PR #4974, opened 2025-01-14
chore: add validate template check

timarney:
# Summary | Résumé

Adds schema validation check for template create.


craigzour (dismissed):
Since the check does not happen in the server action (which is a good thing) you will have to add it to the `updateTemplate` lib function that can also be called by the `createOrUpdateTemplate` server action.

timarney:
@craigzour pushed some updates --- feel free to push changes as you see fit.

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/publish/Publish.tsx
```diff
@@ -161,12 +161,16 @@ export const Publish = ({ id }: { id: string }) => {
...
         id,
         name: getName(),
         formConfig,
       });
 
+      if (operationResult === null) {
+        throw new Error("Failed to publish form");
+      }
+
       router.push(`/${i18n.language}/unlock-publishing`);
```
Comment (craigzour) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/publish/Publish.tsx:174]: @timarney  Not sure how this piece of code worked before. Maybe it never really worked? Since it was in a try/catch block I assume it was expected that the server action would throw something but it was not the case.

Comment (timarney) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/publish/Publish.tsx:174]: Ya your right this must have been missed along the way

timarney:
@craigzour your updates look good.

craigzour:
@timarney  I am also good with the changes and I could find any problem while testing locally.
Will wait for a review from @bryan-robitaille as it touches an important feature.

lib/templates.ts
```diff
@@ -129,6 +138,17 @@ export async function createTemplate(command: CreateTemplateCommand): Promise<Fo
     throw e;
   });
 
+  const validationResult = validateTemplate(command.formConfig);
+
+  if (!validationResult.valid) {
+    logMessage.warn(
+      `[templates][createTemplate] Form config is invalid. ${JSON.stringify(
```
craigzour, line 145: The goal here is just to monitor how often it happens. We can remove it in the future.

craigzour:
@timarney Will be reviewing and probably modifying some code here because server action signatures have been changed and if we keep it this way then we may end-up having the same compatibility problem between the web app and the NextJS server that we had with the `submitForm` server action.

craigzour:
@timarney I just updated the code to make sure we keep backward compatibility with people running the old version of the code in their existing browser tabs. Let's wait for https://github.com/cds-snc/platform-forms-client/pull/4978 to be merged before we merge this.
When we do merge it let's make sure we test that scenario where we start to create/update a template before the new code is deployed and finish it once the deployment is complete.

Issue #2106, opened 2023-05-15
Language Switch Form State


bryan-robitaille:

srtalbot:
Re-tested and still current

timarney:
Experimental PR in progress here:

https://github.com/cds-snc/platform-forms-client/pull/4989

srtalbot:
The tricky bit is how we save values for checkboxes. We save by value, not ID. We have to save data to keep them. 

Link to the save progress discovery brief: https://docs.google.com/document/d/1TdkwKDL3arHm9COMOB1bIaycYUBBEq2cp3gMc8UFC7A/edit?tab=t.0#heading=h.av9ts8u3psrn

timarney:
After doing some further exploration

Have concluded this work is better suited to fall under the save progress epic.

 
https://github.com/user-attachments/assets/7bade555-76df-4add-a4f8-129acef22133

For https://github.com/cds-snc/platform-forms-client/pull/4989 I continued the work done in save to session - 

Issue encountered.

For elements with choices we store the value in the current language.  When the language gets swapped we need to look for the corresponding fr value based on the en value.  This might be considered a risk as technically we're altering user inputted data.  Also if there's ever a bug in the code that ends up failing to match the values we can run into various situations.

timarney:
I added notes above @srtalbot 

Suggesting this we looked as in the save progress epic

Calling out this note

> This might be considered a risk as technically we're altering user inputted data. 

Will let you move this card around as you see fit.

timarney:
Update https://github.com/cds-snc/platform-forms-client/pull/4989

srtalbot:
This is not save progress, but is a small recovery of data for a single session. 

Testing instructions:
- Fill in a longer form
- Switch from FR to EN (or vice versa)
- See if the data stays.
- Test all components 

@anikbrazeau

anikbrazeau:
Testing and works great for the most part: 

- published form with most form elements (attestation causing issues --- will create a ticket)
- multi-page forms with review page (sometimes stays on page, sometimes takes you back to beginning start page but keeps all answers in both cases --- will create a ticket)
- draft form testing page language toggle
- error messages for empty fields
- confirmation page language switch

timarney:
Okay thanks - I will go ahead and merge as the current results in requiring a full re-fill after that we can work ticket by ticket as per above.

PR #4989, opened 2025-01-16
feat: Session storage for front facing forms

timarney:
# Summary | Résumé

- Updates "save to session" code to move beyond handling for server id changes
- Adds `beforeunload` handlers to save form progress to a session so it can be restored
- Adds ability to toggle language from a saved session 

Fixes: https://github.com/cds-snc/platform-forms-client/issues/2106 
> Language switch clears Form State removes data

https://github.com/user-attachments/assets/3d4c579a-2427-4c80-97ef-b939f2e4029b

## Notes

This PR purposely doesn't restore the "current page" --- that can be added easily but we'll want "full form validation" vs "current page" validation first.


timarney:
Closing this 

left notes here:

https://github.com/cds-snc/platform-forms-client/issues/2106#issuecomment-2602801924

components/clientComponents/globals/Header/LanguageToggle.tsx
```diff
@@ -21,16 +20,22 @@ const LanguageToggle = () => {
...
 
+  const handleClick = useCallback(() => {
+    // Dispatch beforeunload event
+    const event = new Event("beforeunload", { bubbles: true, cancelable: true });
+    window.dispatchEvent(event);
+  }, []);
+
   return (
-    <Link
+    <a
```
timarney, line 30: Swapped to a regular link as we want to ensure a full re-load here.

lib/utils/saveProgress.ts
```diff
@@ -0,0 +1,78 @@
...
+};
+
+export const saveProgress = (
+  language: string = "en",
+  { id, values, history, currentGroup }: Options
+) => {
+  if (typeof sessionStorage === "undefined") {
+    return false;
+  }
+
```
thiessenp-cds, line 26: Suggestion for a future iteration. I wonder if we want any error catching here? I guess in the case of an error we'd just show an empty form. Or perhaps an error + empty form would be more helpful.

thiessenp-cds (approved):
Awesome work! LGTM :)

craigzour:
Works well! Nice job :)

PR #5000, opened 2025-01-17
fix: Save on preview to prevent error on uploaded form file

dsamojlenko:
# Summary | Résumé

When uploading a form file, the user is directed to the Preview page.
This page does not autosave, so if the user then went straight to Settings and tried to make a change, it would throw an error.

This PR just adds the SaveButton to the Preview page so that an autosave is triggered when navigating away.


timarney:
Fixes

```
Error in privilege check: PrismaClientKnownRequestError:
Invalid prisma.template.findUniqueOrThrow() invocation:
An operation failed because it depends on one or more records that were required but not found. Expected a record, found none. data: [{"action":"update","subject":{"type":"FormRecord","scope":{"subjectId":""}}}]
```

timarney (approved):
Tested on preview link --- working as expected (auto-saves from preview).

PR #5020, opened 2025-01-21
chore: remove locale prop for Language toggle

timarney:
# Summary | Résumé

The locale prop is not supported for the app router - this PR removes it.

```
The `locale` prop is not supported in `next/link` while using the `app` router. Read more about app router internalization: https://nextjs.org/docs/app/building-your-application/routing/internationalization
```

Console warning:

Screenshot 2025-01-21 at 11 42 43 AM


thiessenp-cds (approved):
Nice and because the lang={language} is on the anchors they will/should be announce with proper pronunciation by screen readers.
LGTM :)

PR #5021, opened 2025-01-21
fix:  Correct "required", "no selection", "Select file" in preview being only english.

### LINKED ISSUE #4683: Form preview doesn't show all elements in the selected language
## Description

When previewing a form, the language toggle is used to preview the form in either English or French. However, some elements on the previewed form respond to the whole page's language instead of the preview language. 

## Steps to reproduce

1. Create a form with required elements and dropdowns.
2. Preview the form and switch the preview toggle to French.
3. Some elements such as "Required" or "No selection" are showing up in English instead of French.

Other case:

1. Create a form with required elements and dropdowns.
2. Switch the whole page to French (top right).
2. Preview the form and switch the preview toggle to English. 
3. Some elements such as "Pas de sélection" are showing up in French instead of English.

## Expected behaviour

When previewing the form, all elements should match the language selected for the preview. 

## Additional context

We received multiple Freshdesk tickets worrying about this.

ShadeWyrm:
Fixes #4683

### DIFF
--- a/app/(gcforms)/[locale]/(user authentication)/components/client/forms/Dropdown.tsx
+++ b/app/(gcforms)/[locale]/(user authentication)/components/client/forms/Dropdown.tsx
@@ -36,8 +36,9 @@ export const Dropdown = (props: DropdownProps): React.ReactElement => {
     ariaDescribedBy,
     validationError,
     onChange,
+    lang,
   } = props;
-  const { t } = useTranslation("common");
+  const { t } = useTranslation("common", { lng: lang });
   const classes = cn("gc-dropdown", className);
 
   const initialDropdownOption = <option value="">{t("dropdown-initial-option-text")}</option>;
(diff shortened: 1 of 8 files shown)


timarney:
Looks good --

Noticed while checking this 

The file upload which is only on for admins continues to have the issue -- so we'll want to fix that in this PR


Screenshot 2025-01-22 at 7 20 51 AM

---

Maybe a follow-up PR but looks like the address isn't working for required?

Screenshot 2025-01-22 at 7 17 58 AM

Screenshot 2025-01-22 at 7 18 14 AM

timarney:
Retested --- looks like the upload "required" is still in fr --- the rest of the text did swap.

Screenshot 2025-01-22 at 2 38 00 PM


--


Also looks like "complete the required field" messages all turn up in EN are you seeing the same?
Screenshot 2025-01-22 at 2 39 20 PM

Issue #3881, opened 2024-06-24
Missing translation ID for error with mix of EN/FR


anikbrazeau:
## Description
Getting error with mix of EN/FR content when adding a section on staging in French

## Steps to reproduce
1. Go to staging
2. Toggle to French
3. Add new section "Nouvelle section"
4. See error with mix of EN and FR

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Staging
- **Language**: French
- **Viewport**: Desktop
- **Output**: /clx3n99lp00006k8zyut3taq0

## Expected behaviour
1. Create new section as expected
2. If something goes wrong, error message in language of interface

## Screenshots or videos

<img width="576" alt="Screenshot 2024-06-24 at 8 56 53 AM" src="https://github.com/cds-snc/platform-forms-client/assets/38330843/8719d6c8-3059-4e91-9412-dbffde2d7a89">

anikbrazeau:
This no longer persists with this specific set of steps to reproduce, can add a new section in French, but I wonder if the English error message with the French button text would still appear if a similar situation was triggered?

Abi-Nada:
"To do" on localization for this error message:

<img width="1507" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBei9RQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7c0e90f843279ae63fba8ac6bd19cbcc68cf3699/Screenshot%202024-07-10%20at%2012.33.17%E2%80%AFPM.png" alt="Screenshot 2024-07-10 at 12.33.17 PM.png" />

Abi-Nada:
De-scoping from the conditional logic work, to be prioritized as a bug outside of this feature

anikbrazeau:
EN: `Something went wrong `
FR: `Quelque chose a mal tourné`

EN: `Try again`
FR: `Veuillez réessayer`

Abi-Nada:
PR has been merged into Staging, going out in tomorrow's release

PR #5022, opened 2025-01-22
fix: global error handling

timarney:
# Summary | Résumé

Fixes:
https://github.com/cds-snc/platform-forms-client/issues/3881

Before:

Screenshot 2025-01-22 at 2 07 44 PM

Uses default Error panel with support link off for global error message.

Screenshot 2025-01-22 at 2 04 12 PM

Screenshot 2025-01-22 at 2 05 47 PM

## Testing 

Used a Fake Error 

See:

https://github.com/cds-snc/platform-forms-client/pull/5022/commits/143871f5a96174ff570db643ef06e69aea53fef1

### DIFF
--- a/app/error.tsx
+++ b/app/error.tsx
@@ -1,38 +1,13 @@
 "use client";
-import { Button } from "@clientComponents/globals";
 import { ErrorPanel } from "@clientComponents/globals/ErrorPanel";
 import { logMessage } from "@lib/logger";
 import { useEffect } from "react";
 
-// This is the app catch all error component. If an error is thrown and an ErrorBoundary does not
-// exist in that route, the error will bubble up the tree and if none are found, end here.
-export default function Error({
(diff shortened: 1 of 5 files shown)


timarney:
https://github.com/vercel/next.js/discussions/68048#discussioncomment-10119674

"An error.js boundary will not handle errors thrown in a layout.js component in the same segment because the error boundary is nested inside that layouts component.

To handle errors for a specific layout, place an error.js file in the layouts parent segment.
To handle errors within the root layout or template, use a variation of error.js called app/global-error.js. "
"The root app/error.js boundary does not catch errors thrown in the root app/layout.js or app/template.js component.

To specifically handle errors in these root components, use a variation of error.js called app/global-error.js located in the root app directory."

dsamojlenko (approved):

Issue #4771, opened 2024-12-03
Moving questions across pages impacts linear branching


Abi-Nada:
## Description
If I have a linear form with multiple pages and I move questions to a different page, it removes the path of my linear branching without me knowing.

As a user, you might not be using branching and just want to move questions to different pages. However, this action entails then having to go to Branching and manually set the linear order of your form.

## Steps to reproduce
1. Create a form with 2 pages + questions on each page
2. Click on 'Branching' and observe the path (linear flow)
3. Go back to 'Pages', and swap questions from each page.
4. Go back to 'Branching'
5. See error: you'll now have to update your branching to go back to being a linear flow

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Staging
- **Language**: EN

## Expected behaviour
Ideally if you're moving questions across pages, it shouldn't alter the route of pages.

## Screenshots or videos

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOTBBQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6c438a65ea87e48910a6b5703aebf15b74e80414/Screen%20Recording%202024-12-03%20at%202.42.09%E2%80%AFPM.mov)

timarney:
https://github.com/cds-snc/platform-forms-client/pull/5024 should fix the issue

timarney:
https://6d2zs5syturpa2t3pbovjmo5640gfkjp.lambda-url.ca-central-1.on.aws/

PR #5024, opened 2025-01-23
chore: fix auto-flow issue

timarney:
# Summary | Résumé

Update to move where we call auto-flow.  Auto-flow was not running when dragging questions between pages.

Fixes: https://github.com/cds-snc/platform-forms-client/issues/4771


thiessenp-cds (approved):
LGTM :)

PR #5028, opened 2025-01-23
fix: Fix the validation message translations not occuring for Preview of Forms

### LINKED ISSUE #5027: "Complete the required field to continue." - Translation doesn't Occur on Preview
A continuation of #5021 and discovered during fixing #4683.

Image

ShadeWyrm:
Closes #5027

### DIFF
--- a/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/Preview.tsx
+++ b/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/Preview.tsx
@@ -78,8 +78,13 @@ export const Preview = ({
     }));
 
   const { t } = useTranslation(["common", "form-builder", "form-closed"]);
+
   const language = translationLanguagePriority;
   const currentForm = getRenderedForm(formRecord, language);
+
+  const translatedT = useTranslation(["common", "form-builder", "form-closed"], {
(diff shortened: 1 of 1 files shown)


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/Preview.tsx
```diff
@@ -78,9 +78,14 @@ export const Preview = ({
     }));
 
   const { t } = useTranslation(["common", "form-builder", "form-closed"]);
+
   const language = translationLanguagePriority;
   const currentForm = getRenderedForm(formRecord, language);
 
+  const translatedT = useTranslation(["common", "form-builder", "form-closed"], {
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/Preview.tsx:85]: Thinking we should move this up under the other `{t}`

At a glance it's easy to miss that we now have t and translatedT

Do you think the array we pass on will ever be different between the 2 ?  Might want to pull out to a const and pass in the same to both?

Comment (ShadeWyrm) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/Preview.tsx]: Nope! It should always be the same, so updated with your suggestions. :)

timarney (approved):

PR #5066, opened 2025-01-29
fix: sort layout and filter out richtext when generating CSVs

craigzour:
# Summary | Résumé

This is a variant to the fix proposed in https://github.com/cds-snc/platform-forms-client/pull/5054

- Reverts a temporary patch applied to CSV output for form responses
- Removes unnecessary use of `sortByGroups` (because it is already sorted before) when ordering elements for CSV generation
- Filters out `richText` elements before generating CSVs

### DIFF
--- a/lib/responseDownloadFormats/csv/index.ts
+++ b/lib/responseDownloadFormats/csv/index.ts
@@ -2,7 +2,7 @@ import { createArrayCsvStringifier as createCsvStringifier } from "csv-writer";
 import { FormResponseSubmissions } from "../types";
 import { FormElementTypes } from "@lib/types";
 import { customTranslate } from "@lib/i18nHelpers";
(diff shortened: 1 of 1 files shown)


Abi-Nada:
Tested the PR and LGTM! ✅

Testing coverage:
- Tested a multipage form that had a show/hide field that included a Repeating Set. Included multiple responses (some responses had chosen the repeating set, some didn't), and validated that the CSV headers appeared correctly ✅
- Tested on my multipage form that was misaligned in prod this morning: validated that it appeared correctly in this PR  ✅
- Tested on a single page form with a show/hide field + repeating set. Downloaded 2 responses without selecting the show/hide+repeating set, validated the csv looked good. Then downloaded 3 responses with a mix of repeating sets and no repeating sets: columns still look good! ✅

timarney (approved):

PR #5067, opened 2025-01-30
chore: Add the migration to produce a perm template to user view object.

ShadeWyrm:
A required addition to handle an issue caused by AWS Services not allowing uppercase column names, and Prisma using uppercase column names for Many-To-Many relation tables.

### DIFF
--- a/prisma/migrations/20250130165334_create_template_user_view/migration.sql
+++ b/prisma/migrations/20250130165334_create_template_user_view/migration.sql
@@ -1,0 +1,3 @@
+CREATE VIEW "TemplateUserView" AS
+SELECT "A" as templateId, "B" as userId
+FROM public."_TemplateToUser";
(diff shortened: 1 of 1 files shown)


prisma/migrations/20250130165334_create_template_user_view/migration.sql
```diff
@@ -0,0 +1,3 @@
+CREATE VIEW "TemplateUserView" AS
+SELECT "A" as templateId, "B" as userId
+FROM public."_TemplateToUser";
```
bryan-robitaille, line 3: Just curious if we need the `public` in `public."_TemplateToUser"`?
Quickly looking through the other prisma migrations the generated SQL usually doesn't reference the `public` schema.

ShadeWyrm: Looks like you're right!~ Cleaned up.

craigzour (approved):
LGTM!

Issue #5079, opened 2025-02-03
Form group mixing Page Text component and Repeating Set that includes a Radio component produces invalid JSON template


craigzour:
## Description
An error happens when you try to modify the list of options on a Radio component that is in a Repeating set part of a form group that has a Page Text component

## Steps to reproduce
1. Create a form
2. Add a Repeating set component
3. Add a Radio component in the Repeating set one
4. Set a new option for that radio component
5. Add a Page Text component outside of the Repeating set
6. Move up the Page Text component so that it is placed before the Repeating set
7. Add a new option to the Radio component
8. Save the form and you should see the error

It appears that the option added during step 7 is being included in the definition of the Page Text component which result in an invalid JSON template.
````
{
      "id": 2,
      "type": "richText",
      "properties": {
        "subElements": [
          {
            "properties": {
              "choices": [
                null,
                null,
                {
                  "en": "3"
                }
              ]
            }
          }
        ],
        "choices": [
          {
            "en": "",
            "fr": ""
          }
        ],
        "titleEn": "",
        "titleFr": "",
        "validation": {
          "required": false
        },
        "descriptionEn": "test",
        "descriptionFr": "",
        "placeholderEn": "",
        "placeholderFr": ""
      }
    }
````

craigzour:
@timarney FYI: While trying to get a fix in the JSON template of that impacted client I noticed that the same bug can also appear in more components than just Page Text.

````
{
      "id": 70,
      "type": "textField",
      "properties": {
        "choices": [
          {
            "en": "",
            "fr": ""
          }
        ],
        "titleEn": "blabla",
        "titleFr": "",
        "validation": {
          "type": "number",
          "required": false
        },
        "subElements": [
          null,
          null,
          null,
          null,
          null,
          {
            "properties": {
              "choices": [
                null,
                null,
                null,
                null,
                {
                  "en": "5"
                }
              ]
            }
          }
        ],
        "descriptionEn": "Enter a number",
        "descriptionFr": "Saisissez un chiffre",
        "placeholderEn": "",
        "placeholderFr": ""
      }
    },
````

anikbrazeau:
Perhaps related to #4708

PR #5080, opened 2025-02-03
bug: fix index issue for sub elements

timarney:
# Summary | Résumé

Fixes: https://github.com/cds-snc/platform-forms-client/issues/5079

Update sub elements to do a lookup for the parent index vs passing in so it stays in sync.


Abi-Nada:
I seem to run into an issue adding radio buttons to a Repeating Set and then saving on a Draft form:

https://github.com/user-attachments/assets/f5bb8fe1-dbcd-4556-b81f-687ee98f50a2

craigzour (approved):

PR #5119, opened 2025-02-13
refactor: introduce new connectors package

craigzour:
# Summary | Résumé

- Create new [gcforms/connectors](https://www.npmjs.com/package/@gcforms/connectors) package. It holds two initial connectors: `gc-notify-connector` and `postgres-connector`
- Rework existing `gcforms-types` package to create new [gcforms/types](https://www.npmjs.com/package/@gcforms/types) one
- Leverages new [gcforms/connectors](https://www.npmjs.com/package/@gcforms/connectors) common package when interacting with GC Notify service

# Tests

- Make sure features involving GC Notify work fine. For example: sign-in to received a 2FA code, share a form, ...

# Note

- After we merge this PR we can remove the legacy [@cdssnc/gcforms-types](https://www.npmjs.com/package/@cdssnc/gcforms-types) package


craigzour:
@timarney I could not keep the NPM workspace pattern that you used for the `types` package due to incompatibilities with the Lambda functions (leveraging the same shared packages) that are based on ModuleJS vs CommonJS for the web application. There might be a way to make it work but I have not found it yet 😞 .

lib/tests/validation/validation.test.js
```diff
@@ -411,13 +411,13 @@ describe("Test getErrorList function", () => {
...
     document.body.appendChild(errorElement);
-    errorElement.focus = jest.fn();
+    const spyFocus = jest.spyOn(errorElement, 'focus');
     const spy = jest.spyOn(document, "getElementById").mockImplementation(() => {
       return errorElement;
     });
 
     setFocusOnErrorMessage(props, "gc-form-errors");
-    expect(errorElement.focus).toHaveBeenCalled();
+    expect(spyFocus).toHaveBeenCalled();
```
craigzour, line 420: I could not understand why this specific test suddenly stopped working. The error was that the `focus` function could not be assigned anything because it only implemented a `getter`.

timarney:
> @timarney Due to incompatibilities with the Lambda functions that are based on ModuleJS vs CommonJS for the web application I could not keep the NPM workspace pattern that you used for the `types` package.
> 
> There might be a way to make it work but I have not found it yet 😞 (I spent too much on this problem 😅 ).

Right --- ya I think the only thing we'll need to make devs aware of if you cmd clicking to edit a type it will land you in the `dist` folder.

**Current pre PR**

https://github.com/user-attachments/assets/4062e10a-ac60-4540-a99b-3617480fe8b2

craigzour:
> > @timarney Due to incompatibilities with the Lambda functions that are based on ModuleJS vs CommonJS for the web application I could not keep the NPM workspace pattern that you used for the `types` package.
> > There might be a way to make it work but I have not found it yet 😞 (I spent too much on this problem 😅 ).
> 
> Right --- ya I think the only thing we'll need to make devs aware of if you cmd clicking to edit a type it will land you in the `dist` folder.
> 
> **Current pre PR**
> 
>  src.mp4

You are right! In development mode we may be able to use https://docs.npmjs.com/cli/v9/commands/npm-link so that we can make changes to the package without having to publish it. To be tested..

bryan-robitaille (approved):
Tested and everything seems to function as it should.
We can explore dev ops automation and local dev in future iterations.

PR #5121, opened 2025-02-14
fix: Prevent changes to groups that would lead to a broken form/flow

### LINKED ISSUE #5031: Exit page conversion causes branching logic to break and view collapses
## Description

Turning a page into an exit page collapses everything in the branching visualization.

When adding an exit page to help people leave the form flow, the logic/branching breaks. It seems to happen if you convert a page to an exit page at the end of the flow, that connects to the review page, it breaks the logic. 

With this action being irreversible. This can cause frustration and for work/time to be lost. The collapsed visualization also makes it impossible to assess/see/fix the issue.

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Production
- **Language**: English
- **Viewport**: Desktop

## Expected behaviour
If someone converts a page to an exit page, it should be taken out of the flow entirely.

## Screenshots or videos

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL1dUQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--50d62df05f98446f789efd339da768a322d1f76a/Screen%20Recording%202025-01-20%20at%209.49.44%E2%80%AFAM.mov)

dsamojlenko:
# Summary | Résumé

Fixes #5031 

There are certain circumstances where the flow can be broken if a page is deleted or set as an exit page.

This PR adds some checks before setting a nextAction or deleting a group:
- Prevent converting to exit page if this page is pointing to Review and is the only page pointing to Review
- Prevent deleting a page if it is the only page pointing to Review


i18n/translations/en/form-builder.json
```diff
@@ -1218,7 +1218,9 @@
       "description": " It seems the items you are moving have display or branching rules applied. Would you like to overwrite those rules and apply a linear autoflow?",
       "cancel": "No, move without changing rules",
       "confirm": "Yes, apply linear autoflow"
-    }
+    },
+    "cannotDeleteGroup": "Cannot delete page as it will lead to a broken form",
+    "cannotSetNextAction": "Cannot set next action to exit as it will lead to a broken form"
```
anikbrazeau, line 1223:
```suggestion
    },
    "cannotDeleteGroup": "This page cannot be deleted as it would break the form logic.",
    "cannotSetNextAction": "This page cannot go to an exit page as it would break the form logic"
```

timarney (approved):

PR #5148, opened 2025-02-20
fix: a11y updates for front facing form(s)

thiessenp-cds:
# Summary | Résumé

A11Y updates for front facing form and save progress dialog. Most of these issues existed before the feature and will exist even without the feature ever shipping.


timarney:
Might want to rename this to 

fix: a11y updates for front facing form or something as you mention "these existed before the feature" and will exist even without the feature ever shipping.

timarney (approved):

PR #5173, opened 2025-02-28
chore: app settings provider

thiessenp-cds:
# Summary | Résumé

Adds an app settings provider and utility hook. Also updates hCaptcha to use the app setting provider to show it works.

A followup PR could be to replace other instances where prop drilling was used to get an app setting.


thiessenp-cds:
A fun experiment but this probably doesn't fit with app settings since a session is required and would not fit well with areas, like forms-forms that have no login.

PR #5193, opened 2025-03-04
Internal packages setup

dsamojlenko:
# Summary | Résumé
Modifies the way "internal packages" are configured.

These are packages that are kept in the packages/ directory, have their own versioning, and are published to NPM for use in other Forms-related repositories (api, terraform, etc). 

Current packages: types, connectors.

The goal is to make it so that when you are working locally in this repository, you can make changes to the internal packages that are available for use immediately while you are working on a change PR, and can subsequently be published to NPM for use in other repositories. This is accomplished through the use of Yarn Workspaces described below.

## Internal package conventions
This PR introduces some proposed packages conventions.

### Prefixing npm scripts
When creating "scripts" in `package.json` prefix the command name with the name of the package.

For example:

```
"scripts": {
    "build": "tsup src/index.ts --format cjs,esm --dts --clean",
    "types:build:watch": "tsup src/index.ts --format cjs,esm --dts --watch",
    "types:prepare": "yarn && yarn build"
  },
```

This makes it so that you can run these commands from the root of the repository, rather than having to move into the directory. It's a convenience and makes for some cleaner root-level scripts like:

```
"prepare": "yarn connectors:prepare && yarn types:prepare",
```

### `src` directory
This is a pretty standard convention in packages anyway, but in our case also makes our typescript paths config a little cleaner, ie:

```
"paths": {
    ...
    "@gcforms/*": ["packages/*/src"]
}
```

## Workspaces setup
- Custom packages as workspaces in packages/*
- Use workspace:* in package.json so always using local/latest
- When a PR detects changes to packages gets label !!Package Update!!
- Server action monitors packages for changes, and ensures the version is bumped along with any changes
- On merge, server action detects changes in package.json and runs yarn npm publish for each package with updates

## Server action enforcement
A new server action internal_packages has been added that identifies any internal packages with changes, and fails if the version in the package has not been bumped.

Example:
image

## Publishing packages

On merge, the workflow `internal_packages_publish` will run when any changes are detected in `/packages/**/package.json` (any package.json updates in the packages)

It will detect which packages have been updated, then run `yarn publish` using the `NPM_TOKEN` env var.


bryan-robitaille (dismissed):
Awesome work!

bryan-robitaille (approved):

PR #5256, opened 2025-03-20
chore: adds caretaker feature flag

thiessenp-cds:
# Summary | Résumé

Adds care taker feature flag

### DIFF
--- a/app/(gcforms)/[locale]/layout.tsx
+++ b/app/(gcforms)/[locale]/layout.tsx
@@ -12,6 +12,7 @@ export default async function Layout({ children }: { children: React.ReactNode }
     FeatureFlags.saveAndResume,
     FeatureFlags.formTimer,
     FeatureFlags.hCaptcha,
+    FeatureFlags.careTakerPeriod,
   ]);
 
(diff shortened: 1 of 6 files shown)


app/(gcforms)/[locale]/layout.tsx
```diff
@@ -12,6 +12,7 @@ export default async function Layout({ children }: { children: React.ReactNode }
     FeatureFlags.saveAndResume,
     FeatureFlags.formTimer,
     FeatureFlags.hCaptcha,
+    FeatureFlags.caretakerPeriod,
```
Comment (timarney) [inline app/(gcforms)/[locale]/layout.tsx:15]: We should just call this banner or something as it will change depending on what the text is

Comment (thiessenp-cds) [reply app/(gcforms)/[locale]/layout.tsx:15]: IMO: that seems dangerous to me since what banner content would shown is not clear. The content strings can be changed and what banner is shown is unclear IMO.

Comment (thiessenp-cds) [reply app/(gcforms)/[locale]/layout.tsx:15]: For now let's just leave it and come back to it later to get this out the door

PR #5284, opened 2025-03-25
feat: client logger

thiessenp-cds:
# Summary | Résumé

Allows sending a specific error code from the client to a server action to be logged. Currently this is only used in the save-resume feature.

## Test

**Force a client error**
Fill out a form with save-progress enabled. Save the form file progress locally. Modify the file data to have bad data. Resume progress and upload that file.  A toast error will be shown and a log message should have been sent. Check the cloudwatch logs for the error.

Example Log Error Message:
Screenshot 2025-03-25 at 2 32 47 PM

Example Manually triggering an error:

https://github.com/user-attachments/assets/d5687138-a669-45c0-ad49-dcdf46b9dc74


components/clientComponents/forms/ResumeForm/ResumeForm.tsx
```diff
@@ -17,6 +17,7 @@ import { ErrorResuming } from "./ErrorResuming";
 import { ResumeUploadIcon } from "@serverComponents/icons/ResumeUploadIcon";
 import { LightBulbIcon } from "@serverComponents/icons/LightBulbIcon";
 import { GcdsH1 } from "@serverComponents/globals/GcdsH1";
+import { useLogMessage } from "@lib/hooks/LogMessage/useLogMessage";
```
timarney, line 20: Might want to call this useLogClientMessage ? or something so we can tell the difference as to when / where to use.

thiessenp-cds: I originally called it `useLogMessageClient` but thought it was too long. But I think you're right, especially when trying to find the right import, being specific is better.

lib/hooks/LogClientMessage/action.ts
```diff
@@ -0,0 +1,33 @@
...
+
+// Logs an error message from the client.
+// e.g. log message:
+// {"level":"error","time":1742927527001,"msg":"Client Error: FR05-1742927526697 - formID: cm70y5cl70000yo69tgxe9j25"}
+export const logErrorMessage = async (
+  code: string,
+  formId: string,
+  timestamp: number
+): Promise<boolean> => {
+  if (!code || !formId || !timestamp) {
```
thiessenp-cds, line 18: TODO anyway to check the formId and timestamp - valid, length,...?

timarney (approved):

PR #5309, opened 2025-03-26
chore: Move flag initialization to internal nextjs startup

bryan-robitaille:
# Summary | Résumé
Moves the feature flag initialization step into the NextJS instrumentation function that runs at server start up. 

With the migration to Rainbow Deployments `yarn` scripts cannot be run as part of the server startup routine.


instrumentation.ts
```diff
@@ -10,28 +10,11 @@ export async function register() {
       console.log("No REDIS_URL environment variable found, skipping privilege cache flush");
```
craigzour, line 10:
```suggestion
      console.log("No REDIS_URL environment variable found, skipping flags initialization and privilege cache flush");
```

lib/flags/initialization.ts
```diff
@@ -0,0 +1,87 @@
...
+    .exec();
+};
+export const initiateFlags = async () => {
+  logMessage.info("Running flag initiation");
+  if (!cacheAvailable) {
+    process.env.APP_ENV !== "test" &&
+      logMessage.info("Application is TEST mode, skipping flag initiation");
+    Boolean(process.env.REDIS_URL) ??
+      logMessage.info("No REDIS_URL environment variable found, skipping flag initiation");
+  }
```
craigzour, line 48: We probably don't need this check here since it is already done in the `instrumentation.ts` code.

However, if we do need it, we are missing a `return;` at the end to actually skip the operation.

craigzour (changes requested):
Suggested small tweaks/fix.

lib/flags/initialization.ts
bryan-robitaille: Thank you for the catch about the missing `return`.  My brain is turning to mush.

craigzour (approved):
Nice one :)

.devcontainer/Dockerfile
```diff
@@ -4,18 +4,6 @@ ARG USERNAME=node
 ARG USER_UID=1000
```
timarney, line 4: Does anyone use the dev container --- might be a good time to just remove :)

PR #5336, opened 2025-04-02
feat: add Rainbow deployment cleanup and application deployment rollback scripts

craigzour:
# Summary | Résumé

- TBA


scripts/rollback-latest-deployment.sh
```diff
@@ -0,0 +1,79 @@
...
+    --output text
+)
+
+echo "Create deployment using previous revision ($previous_revision_sha256)"
+
+new_deployment_identifier=$(
+  aws deploy create-deployment \
+    --application-name AppECS-Forms-form-viewer \
+    --deployment-group-name DgpECS-Forms-form-viewer \
+    --revision revisionType=AppSpecContent,appSpecContent={sha256=a72b5abc6eb1a424b958104085eec8a6fa0bce28ee32d307204dd3c125a8ce8b} \
```
bryan-robitaille, line 38: should the `appSpecContent` here be referencing the `$previous_revision_sha256` instead of being hardcoded?

bryan-robitaille (changes requested):
Quick change on a possibly hardcoded value.

scripts/rollback-latest-deployment.sh
craigzour: Good catch!

bryan-robitaille (approved):
Nice Work!

PR #5339, opened 2025-04-02
fix: adds a check to ignore deleted forms for many template operations

thiessenp-cds:
# Summary | Résumé

Adds a prisma extension to do a ttl check on most template queries. 
This extension approach was preferred over adding the ttl check to each prisma query. This way there is no chance of a copy+paste error when copying past prisma queries. It's also more performant than a few cases that queried the DB and then did a ttl check.

## Test

Test the happy paths with form CRUD (create, view, updated, delete). 
Also try:
1. create a form
2. in a separate tab go that form's settings page
3. from the "my forms" page delete that form
4. in the tab with the settings page, try reloading and the page should now git a not found

### DIFF
--- a/lib/integration/prismaConnector.ts
+++ b/lib/integration/prismaConnector.ts
@@ -18,6 +18,27 @@ const prismaClientSingleton = () => {
         },
       },
     },
+    query: {
+      template: {
+        $allOperations({ operation, args, query }) {
+          // Add the ttl check to these operations
+          const allowedOperations = [
+            "findFirst",
+            "findUnique",
+            "findFirstOrThrow",
+            "findMany",
+            "findUniqueOrThrow",
+            "update",
+            "updateMany",
+          ];
+          if (allowedOperations.includes(operation)) {
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            (args as any).where = { ...(args as any).where, ttl: null };
+          }
+          return query(args);
+        },
+      },
+    },
   });
 };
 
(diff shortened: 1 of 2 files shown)


lib/integration/prismaConnector.ts
```diff
@@ -18,6 +18,27 @@ const prismaClientSingleton = () => {
...
+            "findFirst",
+            "findUnique",
+            "findFirstOrThrow",
+            "findMany",
+            "findUniqueOrThrow",
+            "update",
+            "updateMany",
+          ];
+          if (allowedOperations.includes(operation)) {
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
thiessenp-cds, line 35: TODO

lib/templates.ts
```diff
@@ -1227,8 +1225,10 @@ export async function deleteTemplate(formID: string): Promise<FormRecord | null>
   const numOfUnprocessedSubmissions = await unprocessedSubmissions(formID, true);
   if (numOfUnprocessedSubmissions) throw new TemplateHasUnprocessedSubmissions();
 
-  const dateIn30Days = new Date(Date.now() + 2592000000); // 30 days = 60 (seconds) * 60 (minutes) * 24 (hours) * 30 (days) * 1000 (to ms)
+  // Check and delete any API keys from IDP
```
thiessenp-cds, line 1228: Moved up to avoid a race condition when deleting a form

bryan-robitaille (approved):

anikbrazeau (approved):
Tested a bunch of paths and always got a 404 for the deleted form ID:
- draft + published forms
- EN + FR URLs
- edit / preview / settings / publish / responses

timarney (approved):

timarney:
@thiessenp-cds will get this merged for you :)

PR #5346, opened 2025-04-03
chore: update styles for resume page

timarney:
# Summary | Résumé

Updates styles to better align when in french.


samsadasivan (approved):
Looks good!!

PR #5385, opened 2025-04-08
fix: remove Redis disconnect function call (tentative fix)

craigzour:
# Summary | Résumé

- (tentative) Fixes production `Connection is closed` errors
  - Removes the "disconnect from Redis" function call
  
# Note

Not yet sure why we have not seen it before in Staging (probably because the traffic was not high enough) but since it happens in both the ECS task and Rainbow lambdas it must have something to do with the NextJS server being built in standalone mode + what we changed to be able to achieve this (aka flag initialization).

I was able to confirm that the error we keep getting in Production is thrown by the `ioredis` library that we use to interact with Redis.

Here is the specific code that produces those errors:

```

export declare const CONNECTION_CLOSED_ERROR_MSG = "Connection is closed.";

===

function close() {
        self.setStatus("end");
        self.flushQueue(new Error(utils_1.CONNECTION_CLOSED_ERROR_MSG));
    }

===

if (self.manuallyClosing) {
            self.manuallyClosing = false;
            debug("skip reconnecting since the connection is manually closed.");
            return close();
        }
```

### DIFF
--- a/instrumentation.ts
+++ b/instrumentation.ts
@@ -2,13 +2,17 @@ export async function register() {
   if (process.env.NEXT_RUNTIME === "nodejs") {
     if (!process.env.REDIS_URL) {
       // eslint-disable-next-line no-console
-      console.log("No REDIS_URL environment variable found, skipping flags initialization and privilege cache flush");
+      console.log(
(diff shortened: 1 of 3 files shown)


bryan-robitaille (approved):
LGTM

PR #5427, opened 2025-04-15
chore: update share modal

timarney:
# Summary | Résumé

- Updates route to pull template direct from DB vs passing from the UI
- Validates email addresses
- Shortens filename as needed.


craigzour (approved):
LGTM! Tested with both the UI and a direct cURL request.

PR #5432, opened 2025-04-16
chore: Add return for middleware check

timarney:
# Summary | Résumé

- Updates middleware to return early on payload check
- Updates session storage error handling

Screenshot 2025-04-16 at 11 38 40 AM


craigzour (approved):
LGTM!

Just a reminder for when you have time. You can make a PR to check body size instead of relying on headers.

PR #5435, opened 2025-04-16
chore: remove feature flag for save and resume

timarney:
# Summary | Résumé

Clean-up PR to remove save and resume feature flag now that the feature is on by default.


thiessenp-cds (approved):
Tested the save and resume feature and checked that the feature flag is gone from admin. LGTM :)

PR #5461, opened 2025-04-24
feat: add info log when someone signs in from outside of Canada

craigzour:
# Summary | Résumé

- Adds info log for when someone signs in from outside of Canada. It leverages https://github.com/cds-snc/forms-terraform/pull/996


bryan-robitaille (approved):
LGTM

Issue #5441, opened 2025-04-17
Rule-controlled question becomes invisible once the controlling question is deleted


jprince-cds:
## Description
When a question's visibility is controlled by a rule from another question, if that other question is deleted, the question is no longer visible on the form (although it remains visible in edit mode). 

## Steps to reproduce
1. Create a question with an "Other" option. 
2. Test the form, the Other question is visible when "Other" is selected. All good. 
3. Delete the question created in step 1. 
4. The "Other" question becomes invisible even though there is no rule on it anymore. 

## Expected behaviour
Since the question's visibility is no longer controlled by a condition, it should now be always visible. 

## Screenshots or videos
[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM2E1Q1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--43b91d1ff588fce1785a62303ed9599fb047845d/test-bug-2025-04-17.mov)

## Additional context
Reported in [Freshdesk ticket 22204.](https://cds-snc.freshdesk.com/a/tickets/22204)

jprince-cds:
[test-bug-2025-04-17 - Before Deletion.json.txt](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM3U1Q1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f59bde899e1741bf020543898828ed4d407f273c/test-bug-2025-04-17%20-%20Before%20Deletion.json.txt)

[test-bug-2025-04-17 - After Deletion.json.txt](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMzI1Q1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f4f517be80abf20af2e1bc2b1a283975aed702eb/test-bug-2025-04-17%20-%20After%20Deletion.json.txt)

PR #5462, opened 2025-04-24
bug: add clean rules for save action

timarney:
# Summary | Résumé

This PR addresses a bug in the save action by cleaning up element rules (ensuring parent elements exist) before returning the schema and when importing templates.

- Updated save action and template import to invoke rules cleanup
- Added a new helper (cleanElementRules) and tests to validate the cleanup functionality
- Modified store schema generation and updated UI components to use the cleaned schema

Fixes: https://github.com/cds-snc/platform-forms-client/issues/5441

See issue for files to test with.

https://github.com/user-attachments/assets/020ce654-4c9f-4d32-8dce-82a4a51c6be0


craigzour (approved):
Reviewed and tested. Bug is definitely fixed. Will let you decide if it is worth having a second review from either Dave or Pete on this.

Issue #5470, opened 2025-04-28
552 Error from Opening a Form -> Settings


Abi-Nada:
## Description
When opening a form template (lands on Preview) to then navigating to Settings (without making any changes), the user encounters a 552 error.

(Labelled this as Severe, because the flow is pretty confusing. The user would need to go back and toggle around to the other pages, like Edit, to make a Save and then be able to go to Settings - but they wouldn't know to do that intrinsically).

## Steps to reproduce
1. Open a form file (lands on Test/Preview)
2. Go to Settings (don't make any changes to the form)
3. See error

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Prod

## Expected behaviour
If a user opens a form file, they should be able to navigate to any of the other pages (like Settings) without encountering a 552 error.

## Screenshots or videos

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL1BVQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--39cf5059e8926de0c352b2094746ad4955d1c882/Screen%20Recording%202025-04-28%20at%2010.37.34%E2%80%AFAM.mov)

## Additional context

From conversation with Tim, it looks like the code checks to see if any changes were made before prompting a Save. But for this flow of opening a form -> land on Preview, there's no changes. We can potentially implement code to disregard this and save anyway.

timarney:
Digging deeper into 

With the current setup the save template call is getting made but it doesn't complete until after the settings page has loaded i.e. when the settings page loads no id is available yet,

https://github.com/user-attachments/assets/35dcfeec-5a00-4943-9ba0-96981e60306a

Abi-Nada:
Fix implemented in #5478

PR #5471, opened 2025-04-28
chore: update autosave

timarney:
# Summary | Résumé

Updates auto-save fixes https://github.com/cds-snc/platform-forms-client/issues/5470


Abi-Nada:
Tested and LGTM!

Issue #4103, opened 2024-07-30
Sign out page time changes when language is toggled


anikbrazeau:
## Description
Signing out of GC Forms gives a timestamp of when signing out happened.
If you toggle between English and French it keeps updating the time

The size of the font also does not seem to match other "paragraph/body" styles used throughout the app. It seems smaller than expected.

## Steps to reproduce
1. Go to Staging
2. Sign in to your account
3. Sign out
4. Notice the time
5. Switch to the other language
6. Notice the time

## Details
- **Browser**: Edge and Chrome
- **Operating system**: Windows and Mac
- **Environment**: Staging and Prod
- **Language**: En / Fr
- **Viewport**: Tablet

## Expected behaviour
Static timestamp?

## Screenshots or videos

https://github.com/user-attachments/assets/a8aa771b-75e0-4671-8e7b-cd476bd6f6c5

Abi-Nada:
I believe that's expected functionality - looks like it's just the seconds that are updating as we toggle, but the time is correct:

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOHhSQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--21a82895aaa08971d31f589eb306db65c178561d/Screen%20Recording%202024-10-16%20at%208.38.09%E2%80%AFPM.mov)

PR #5500, opened 2025-05-06
fix:  Sign out page time changes

timarney:
# Summary | Résumé

Fixes https://github.com/cds-snc/platform-forms-client/issues/4103

- Saves logout time to a session var so it doesn't change on every load.


thiessenp-cds (approved):
The session time doesn't change on each page reload. LGTM :)

PR #5517, opened 2025-05-08
chore: check value before reset for classification type

timarney:
# Summary | Résumé

- Update to keep value as API when Protected B is selected.


anikbrazeau (approved):
LGTM! Tested and does not seem to revert to vault.

craigzour (approved):
LGTM!

PR #5519, opened 2025-05-08
fix: ensure classification is saved when creating an API key

timarney:
# Summary | Résumé

Ensure save classification happens as needed when creating an API key.


anikbrazeau (approved):
Tested in UI and LGTM!

craigzour (approved):
LGTM

PR #5548, opened 2025-05-13
fix: RichText renderer inserting HTML entity for space char

### LINKED ISSUE #5534: Rich Text Editor Adds Special Characters  to bolded text
Rich Text Editor Adds Special Characters to bolded text only when the trailing space is also bolded. The colon seems to have no impact.

Attached is a video showing how to reproduce the error

https://github.com/user-attachments/assets/44c01917-8ac7-4495-8376-f51310942458

dsamojlenko:
# Summary | Résumé

Fixes #5534 

This appears to be a mismatch between the `markdown-to-jsx` parser and Lexical markdown converter in how they render non-breaking spaces - Lexical inserts `&#32;` for non-breaking spaces, but `markdown-to-jsx` prefers `&nbsp;`. 

You can prove this:
- create a form that demonstrates the problem
- download the form file
- replace `&#32;` with `&nbsp;`
- re-import the form
- note the spaces render fine

This simple fix adds a renderRule to search for and replace the HTML entity when rendering the markdown.

A longer-term fix would be to align our markdown parsing/conversions by using a single rendering engine for both front-end forms and the RichText editor (likely replacing `markdown-to-jsx` with the Lexical Markdown renderer on the front-end)


thiessenp-cds (approved):
Impressive debugging. LGTM :)

thiessenp-cds (approved):
LGTM :)

PR #5560, opened 2025-05-15
fix: fixes notifications sending when delivery is email

thiessenp-cds:
# Summary | Résumé

Fixes an issue where notifications were still being sent when the email delivery options was selected.

### DIFF
--- a/app/(gcforms)/[locale]/(form filler)/id/[...props]/actions.ts
+++ b/app/(gcforms)/[locale]/(form filler)/id/[...props]/actions.ts
@@ -89,11 +89,7 @@ export async function submitForm(
 
     const submissionId = await processFormData(data.fields, data.files, language);
 
-    // Avoid spamming users with emails by only sending one type of email, either an email
-    // delivery or a notification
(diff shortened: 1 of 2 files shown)


timarney (approved):

Abi-Nada:
Tested the email flow and LGTM :) I'm going to do a bit more testing all around to be extra extra sure, but looks good so far!

Abi-Nada:
Tested the following 4 flows and all LGTM!!!! :

✅  Created a PB form to download response from the vault -> validated that the toggle is ON by default
	- Submitted 3 responses, received 2 emails
✅  Created a PA form to download responses from the vault -> validated that the toggle is ON by default
	- Submitted 1 response, then turned OFF the toggle, then submitted more responses
	- Validated that I only received that first email notification
✅  Created a PA form to download responses from the vault -> decided to turn the toggle OFF
	- Submitted 3 responses, received zero notifications
✅  Created an email delivery form -> validated that the toggle is OFF by default
	- Submitted 3 responses, received zero notifications
	- Tried to go back to toggle the feature, still cannot

PR #5574, opened 2025-05-22
chore: Updates notifications to be off for new forms

thiessenp-cds:
# Summary | Résumé

Updates notifications to be off for new forms.

## Test

Create a new form then go in Settings > Form Management. The Notifications setting should be turned OFF by default. Publish the form then go back to the Form Management page and the Notifications settings should remain OFF.

Create a new form then go in Settings > Form Management. The Notifications setting should be turned OFF by default. Turn the setting ON. Publish the form then go back to the Form Management page and the Notifications settings should remain ON.

You could also try creating a few submissions for each form for the above two cases and see if Notifications are sent or not depending on the setting. But this may not be necessary since the update doesn't touch the related code.


thiessenp-cds:
Will be going with a simpler approach

thiessenp-cds:
Re-closing this PR and going with this one instead #5605

PR #5577, opened 2025-05-23
Update Terms of Use to emphasize public service 

anikbrazeau:
Clarifying that this is not intended to be used for political purposes


anikbrazeau:
Closing as policy agrees it was clearer previously. 
Consultants with government email addresses can use our products.

PR #5580, opened 2025-05-23
fix: Rainbow deployment add new lb rule for non actions

bryan-robitaille:
# Summary | Résumé
Screenshot 2025-05-23 at 10 35 32 AM

Ensures that all application assets that do not have a `x-deployment-id` header also get routed to the same legacy instance that the resources can be found on.  This was not done originally as the assumption was that the local client copy of JS package would detect the drift and cause a full refresh on next navigation.  Unfortunately this was found not to be the case.  As we continue to explore ways to provide a more meaningful user experience after version launches this PR should ensure a user never gets a broken app.

Need to find a way to let the user know that a new version is available and replace the JS bundle.

# Pull Request Checklist

### DIFF
--- a/.github/workflows/rainbow-deployment/action.yml
+++ b/.github/workflows/rainbow-deployment/action.yml
@@ -104,3 +104,10 @@ runs:
           --priority 1 \
           --actions Type=forward,TargetGroupArn=$target_group_arn \
           --tags Key=Name,Value=rainbow-${{ inputs.deployment-identifier }} > /dev/null 2>&1
+
+        aws elbv2 create-rule \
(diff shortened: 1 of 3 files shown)


craigzour (changes requested):
With this change we also need to increase the number of maximum rules (one rule was equal to one rainbow lambda) we want to keep when the auto cleanup script runs.

For `staging-rainbow-cleanup` we can set `MAXIMUM_NUMBER_OF_RAINBOW_DEPLOYMENTS_TO_KEEP_ALIVE` to 4

For `prod-rainbow-cleanup` we can set `MAXIMUM_NUMBER_OF_RAINBOW_DEPLOYMENTS_TO_KEEP_ALIVE` to 12

PR #5593, opened 2025-05-26
Fix : Set lastLogin to be saved to DB.

### LINKED ISSUE #5586: Lastlogin and createdat are being set to the same time.
Issue was found when evaluating the number of active users - many returned lastlogins having the same createdat date when they had very likely logged in. 

Issue may be at any point in the transition of information from the system (eg: It not saving lastlogin right) to the storage in the data lake.

ShadeWyrm:
Closes #5586 

Adds a call to save the lastLogin to the current date when the user succesfully logs in.


bryan-robitaille:
Looking at this quickly it seems the `lastLogin` is currently holding the true value for the account creation date. I'm wondering if we want to somehow correct the `createdAt` date with this data before it's lost/overwritten by this PR change?

bryan-robitaille (approved):
Very nice and clean.  Works like a charm.

PR #5633, opened 2025-05-30
[experimental] notifications v2-1 UI (based off branch feat/email-notifications-v2-1)

thiessenp-cds:
# Summary | Résumé

TODO

- update to not store the users / notificationUsers in the template store. Use a server action or prop drilling instead
- update sending notifications to use the notificationsUsers list
- add setter for updating notificationsUsers


thiessenp-cds:
Going with a different approach. Will use a different PR

Issue #5621, opened 2025-05-29
File upload:  Cancel + select new file says no file selected


timarney:
## Description

Re-selecting a file doesn't show the newly selected file

## Steps to reproduce
1. Upload a file
2. Cancel selected file
3. Select a new file
4. See new file doesn't show as selected

## Expected behaviour
Newly selected item should show

## Screenshots or videos

https://github.com/user-attachments/assets/fbaa966a-377e-4b8b-ab81-0748d23e9837

PR #5642, opened 2025-06-02
fix: Cancel + re-select file says no file selected

timarney:
# Summary | Résumé

Fixes issue where uploading a file -> pressing cancel + re-uploading the same file showed "no file selected".

- Update to clear the hidden / real input value when using cancel (reset)
- Updates the cancel button to be a real button for better a11y.
- Removes unused export `ALLOWED_FILE_TYPES` --- as it's now just a duplicate of the same in fileValidationClientSide

Fixes: #5621

https://github.com/user-attachments/assets/2ceb3f63-db46-4c2a-ae24-8438369eb349


lib/fileInput/constants.ts
```diff
@@ -9,19 +9,3 @@ export const FILE_GROUPS = {
     types: ["xls", "xlsx", "csv", "numbers"],
   },
 };
-
-export const ALLOWED_FILE_TYPES = [
```
timarney, line 13: Removes unused export from 

https://github.com/cds-snc/platform-forms-client/pull/5609

dsamojlenko (approved):

PR #5647, opened 2025-06-03
chore: update file size limits

timarney:
# Summary | Résumé

- [x] Update to allow large payloads for file uploads.
- [x] Updates code to make it easier to adjust file size limits

Note: Preview env doesn't support larger uploads.

> "Request must be smaller than 6291456 bytes for the InvokeFunction operation"


i18n/translations/en/form-builder.json
```diff
@@ -379,7 +379,7 @@
     "number": "Enter a number",
     "phone": "For example: 111-222-3333",
     "combobox": "Start typing to narrow down the list.",
-    "fileInput": "Files must not exceed 3.5 MB in total."
+    "fileInput": ""
```
timarney, line 382: Now empty so we don't show by default

dsamojlenko (approved):

PR #5656, opened 2025-06-05
fix: FormattedDate component: set upper limit for Year

dsamojlenko:
# Summary | Résumé

Whoever's around for the Y10K rollover can bump it next.


timarney:
Just some lint issues to look at

PR #5678, opened 2025-06-10
fix: Check page visibility while checking rules for element visibility/validation

dsamojlenko:
# Summary | Résumé

While [recursively checking conditional rules for element visibility](https://github.com/cds-snc/platform-forms-client/pull/5559), also check for page visibility.

Addresses an [issue raised in the conversation](https://github.com/cds-snc/platform-forms-client/pull/5559#issuecomment-2904197788) below that previous PR.

While recursively checking the visibility of an element based on Conditional Rules, each element in that recursive chain appears on a page that may or may not itself be visible. This PR uses the groupHistory array on the values object to determine visibility of each page checked.


timarney:
Tested using template in the comment works great .

PR #5680, opened 2025-06-11
chore: adds prisma data migration for draft forms with delivery option email to be vault

thiessenp-cds:
# Summary | Résumé

Adds a prisma data migration for draft forms with delivery option email to be updated to the vault delivery method.

## Testing

Sign in to AWS and run a query to get a count of all the forms that are in draft and have the email delivery method.

After starting up the app, in the console output there should me a message with a count of the number of records updated. Like this, where count is the number of records updated
```
{count} were migrated for move email Delivery Option to Vault migration
```

Compare the counts and they should match.

Run the same AWS query and the count of all the forms that are in draft and have the email delivery method should be 0.


timarney:
Tested locally 

Can see the migration runs 

Screenshot 2025-06-11 at 8 59 01 AM

Worked to move my test email draft forms over to vault
Screenshot 2025-06-11 at 9 01 23 AM

Kept publish email form intact.
Screenshot 2025-06-11 at 9 01 33 AM

Issue #5630, opened 2025-05-29
File Upload: User-based attribute in the DB to provide access file upload component in the form builder


srtalbot:
### Rationale
We are starting off with an API only cohort that can use file attachments. Teams will need a developer, and only a few users will be able to add the file upload components to their forms.

The user model can be updated to add a "beta" user attribute. or the like; but essentially it helps us release beta features to a sub-set of our clients so they can test on production. This feature would have been useful in the past, and as features continue to get more complex we will need testing cohorts like these more and more. 

**This is not:** a start on Team Management. This is a feature that helps us test beta features that will have bugs with smaller cohorts that understand the risks that come with early access. 

**Requirements definition to do**
- [ ] Connect with support on their needs for adding / removing the beta_user attribute. @Abi-Nada 
- [ ] Do they want full name or partial matching? Live partial matching?

### Done when
- [ ] We can select specific clients on GC Forms that are given access to "beta" features
- [ ] Ensure the "add a file upload" component checks for the beta_user attribute so they can see the option
- [ ] A way for SuperAdmins to view the users with beta access
- [ ] The for SuperAdmins ability to revoke beta feature access

### Out of scope
- Automatic notifications to the client letting them know their beta_user access has been added, or revoked.

ShadeWyrm:
100% would load on login and then store in the session to avoid DB calls. :)

srtalbot:
One other thing to flag, is @Abi-Nada will be reaching out to Support to get a better idea of what they need to enhance their Request to Publish flow. It's their highest frequency ticket. 

We'll get some requirements this week and can discuss whether it should be in scope or not when completing this feature.

Abi-Nada:
Next steps: create a design doc/options analysis. Include pros and cons for each option, to get alignment on a path forward

timarney:
Thanks for the write-up given it sounds like we can latch onto some existing functionality (modified for this use case) the many to many sounds like the way to go.

PR #5682, opened 2025-06-11
feat: Add User based Feature Toggle

ShadeWyrm:
Adds a toggle for feature flags that enables a feature flag if a user has an associated entry in the UserFeature model. Part of #5630 

To test:
-> Ensure addressComplete's feature flag is off.
-> Add your user the "addressComplete" feature with yarn prisma:studio
-> Open a form, and notice it loads the CanadaPost API version of the Address component.

Hooray. 🎉

This PR is Step 1.

Step 2 will be a later PR for adding a more useful admin panel function for managing these flags.


timarney:
Nice work --- will just wait for some others to review as well but good from what I see.

PR #5686, opened 2025-06-11
chore: add new notifications users field to template schema

thiessenp-cds:
# Summary | Résumé

Adds a new column `notificationsUsers` to the Template table.  This will be used by email notifications v2 to keep track of which users on a form have email notifications enabled.


timarney (approved):

timarney:
Verified this outside of Prisma studio  --- looks like the relation gets set up using a view 

Screenshot 2025-06-13 at 10 13 07 AM

Issue #5664, opened 2025-06-06
Publishing a form that is already published throws an error that should be properly communicated to the user


craigzour:
## Description
When someone tries to publish a form that was already published they get a vague error saying that an internal error happened. Also it does generate an error log that we receive in Slack (see context link).

Context: https://gcdigital.slack.com/archives/C05G766KW49/p1749213296541499

## Steps to reproduce
1. Open a browser tab
2. Create a form and set it up so that it is ready to be published. Stay on the publish page.
3. Open a second browser tab and navigate to the publish page of that same form you created before
4. Publish the form using the first tab. It should work as expected
5. Try to publish the same form using the second tab (make sure not to reload the page). It should fail and display a vague error.

## Solution
Check whether the form is already published before trying to publish it.

## Expected behaviour
An error message should clearly tell the user that the form has already been published.

timarney:
Logging the errors 

[14:13:59.576] ERROR: Failed to delete form responses from the Vault during publishing for form cmbzcioxt0000nbwff322exqe.


[14:13:59.576] ERROR: Form is published. Cannot delete draft form responses.

There's also a few other cases  ...

PR #5716, opened 2025-06-16
fix: publish step - catch errors when deleting a draft

timarney:
# Summary | Résumé

Fixes: https://github.com/cds-snc/platform-forms-client/issues/5664


lib/templates.ts
```diff
@@ -604,27 +604,38 @@ export async function updateIsPublishedForTemplate(
   publishFormType: string,
   publishDescription: string
 ): Promise<FormRecord | null> {
+  // Alias the isPublished value to newPublishStatus for clarity within the function
```
timarney, line 607: Didn't want to change the function signature for now but "isPublished" is confusing here as it's the incoming status and not the DB status for the template.

lib/templates.ts
```diff
@@ -604,27 +604,38 @@ export async function updateIsPublishedForTemplate(
...
+      if (e instanceof TemplateAlreadyPublishedError) {
+        // Already published, so we can just return the full template
+        return getFullTemplateByID(formID);
+      }
+
+      throw e;
+    }
   }
 
-  // We use a where unique input to ensure we are only updating the form if it is not published
```
timarney, line 617: Removed this 

> ensure we are only updating the form if it is not published

False can be passed in as the param which would also be an update

lib/vault.ts
```diff
@@ -582,10 +582,18 @@ export async function deleteDraftFormResponses(formID: string) {
       responsesDeleted: accumulatedResponses.length,
     };
   } catch (error) {
-    logMessage.error(
-      `Failed to delete form responses from the Vault during publishing for form ${formID}.`
-    );
-    logMessage.error((error as Error).message);
+    if (error instanceof TemplateAlreadyPublishedError) {
+      // Log a warning if the form is published
+      logMessage.warn(
```
timarney, line 587: Updated to just warn we don't need to error out here --- just let UI move along to the published page.

PR #5723, opened 2025-06-18
feat: Add basic Admin Panel feature flag management

### LINKED ISSUE #5630: File Upload: User-based attribute in the DB to provide access file upload component in the form builder

ShadeWyrm:
Closes #5630 

Adds a new list of users with feature flags to the Feature Flag Admin Panel, with the ability to remove feature flags. A "Add" button is available which routes the user to the User Admin page.

The User Management page leads to a new section under "More" menu that leads to the Users individual feature flag management.


timarney (approved):
Nice work.

PR #5741, opened 2025-06-20
* fix: Pull user Features from DB/Cache call instead of Cache only on User Feature Management

ShadeWyrm:
See title.


timarney (approved):

lib/cache/userFeatureFlagsCache.ts
```diff
@@ -9,6 +9,8 @@ const cacheAvailable: boolean = process.env.APP_ENV !== "test" && Boolean(proces
 const randomCacheExpiry = () => Math.floor(Math.random() * 300 + 300);
 
 export const featureFlagsCheck = async (userID: string): Promise<string[] | null> => {
+  //const canManage = await authorization.canManageUser(userID);
+
```
craigzour, line 13:
```suggestion
```

PR #5755, opened 2025-06-23
chore: enable hCaptcha in blocking mode

thiessenp-cds:
# Summary | Résumé

This PR updates hCaptcha to run in blockable mode behind a feature flag. Blockable means that when a users is detected as suspicious by hCaptcha, the form submission will be block.

Note: this PR re-purposes the existing hCaptcha feature flag
- On: hCaptcha runs in blockable mode
- Off: hCaptcha runs in the background and will track scores but with not block form submissions

## Test

Turn the hCaptcha flag On and forms should submit normally in Preview and Published modes.
Do the same but turn the hCaptcha flag Off.

## Future Work

Come up with a way to automate testing using a tool like Cypress. The automation is important so that hCaptcha will detect a bot is submitting a form and block the form submission. This work should be done in a separate repo, something like `forms-smoke-tests`. The reason for the test files isolation is to prevent accidentally making an unintended change to the system e.g. like accidentally updating a data base.


components/clientComponents/globals/FormCaptcha/FormCaptcha.tsx
```diff
@@ -1,50 +1,42 @@
...
 
-  if (captchaEnabled && !hCaptchaSiteKey) {
-    logMessage.error("hCaptcha: hCaptchaSiteKey is missing");
-    return null;
+  if (
+    process.env.NODE_ENV === "development" &&
+    hCaptcha &&
+    !process.env.NEXT_PUBLIC_HCAPTCHA_SITE_KEY
+  ) {
+    logMessage.info(`hCaptcha: flag is enabled but hCaptchaSiteKey is missing. This will cause
```
bryan-robitaille, line 37: We might want to change this to a `warn` level so it's easier to see among all the other info and debug outputs in the console.

components/clientComponents/globals/FormCaptcha/actions.ts
```diff
@@ -61,15 +61,18 @@ export const verifyHCaptchaToken = async (token: string): Promise<boolean> => {
   return true;
 };
 
-// Looks at the success and score to determine a pass or fail. We can tweak the score over time.
-// See https://docs.hcaptcha.com/enterprise#handling-siteverify-responses
+/**
```
bryan-robitaille, line 64: This whole actions file should probably be removed and added instead to `@libs/validation/hcaptcha`.  This file is not a server action in that it is never called directly from any client or server component and is only called from a Form Filler action.

bryan-robitaille (changes requested):
A few quick small changes and we should be good to merge.

Issue #5750, opened 2025-06-23
Server-side validation enhancement


srtalbot:
Definitions: 
Valid data: data that our system should accept, but not necessarily real data that is legitimate from an end-user

### Acceptance criteria:
- [x] Identify how to map an end-user's path through the form to know which questions are actually required.
- [x] Move required fields and question validation to server-side
- [x] Determine if we need additional errors that we show end-users (Error: 422, a list of errors on the form, exactly as you would see in the browser)
- [x] Significant testing ... 
- [ ] Release in passive mode and implement logs that we are looking for in prod
- [ ] Set a threshold for implementing in production.

PR #5779, opened 2025-06-25
fix: Page visibility and server validation

dsamojlenko:
# Summary | Résumé

Re: #5750 

This PR adds server side validation for form submissions. It uses the same `validateOnSubmit` function that fires for client side validation, but acts on the entire values object on the server on submit.

### validateOnSubmit refactor
This PR also refactors the validateOnSubmit function which previously would only validate data for inputs by looping over the submitted values and validating each one against the FormElements. The problem is, you could bypass validation completely by passing an empty values object or an object not containing any valid form element IDs. Now we loop over the FormElements and validate any provided values.

### repeatingSet validation
Also included is a fix for a pre-existing bug related to required inputs in a repeating set. Repeating sets had the same problem described above- since we were looping over only provided values and checking the related formElement to determine validation, simply not providing an element in values bypasses its validation. This code has been modified similarly to above, where we now loop over the SubElements and validate any provided (or not) values.

### checkPageVisibility
Additionally, checkPageVisibility has been refactored to use submitted values and the form template to determine whether an input should be validated, rather than the client-provided groupHistory.

To accomplish this, a new recursive function `getVisibleGroupsBasedOnValuesRecursive` is introduced that: 
- loops over the form template groups object
- inspects the submitted values to determine the path taken through the form 
- returns an array of `visibleGroups` to replace `groupHistory`

Other new supporting functions:
- `getValuesWithMatchedIds` - returns a values object with matchedIds for radio/checkbox answers for easier nextAction parsing
- `findGroupByElementId` - returns the group that contains a given element

### Review page
Review page has been updated to use the new getVisibleGroupsBasedOnValues for rendering the page instead of groupHistory.

@TODO (future refactorings/cleanup):
- Since `getVisibleGroupsBasedOnValuesRecursive` is now being used for the Review page, can do a cleanup of all the old `groupHistory` code
- The `getValuesWithMatchedIds` can probably replace `mapIdsToValues` but will require further refactoring


lib/formContext.ts
```diff
@@ -147,40 +147,163 @@ export const getElementById = (elements: FormElement[], id: string) => {
...
+ * @param values
+ * @returns {FormValues} - Returns a new FormValues object with matched ids for checkboxes and radios.
+ */
+export const getValuesWithMatchedIds = (formElements: FormElement[], values: FormValues) => {
+  const newValues = { ...values };
+  Object.entries(values).forEach(([key, value]) => {
+    if (key !== "currentGroup" && key !== "groupHistory" && key !== "matchedIds") {
+      const el = getElementById(formElements, key);
+
+      if (value && el && (el.type === "checkbox" || el.type === "radio")) {
```
timarney, line 162: We have a function called 

```
export const findChoiceByValue = (
```

We don't have to use it but might make sense --- just pointing it out.

lib/formContext.ts
```diff
@@ -147,40 +147,163 @@ export const getElementById = (elements: FormElement[], id: string) => {
...
-    : [values.groupHistory];
-  return !!groupHistory.find((visitedGroupId: string | undefined) => visitedGroupId === groupId);
+  // Get an array of values with matched ids instead of raw values
+  const valuesWithMatchedIds = getValuesWithMatchedIds(formRecord.form.elements, values);
+
+  // Recursively build up the groupHistory array based on values
+  const visibleGroups = getVisibleGroupsBasedOnValuesRecursive(
+    formRecord,
+    valuesWithMatchedIds,
+    "start"
```
timarney, line 273: Believe we have a type we can use here vs the sting

PR #5702, opened 2025-06-13
feat: email notifications v2

thiessenp-cds:
# Summary | Résumé

Updates Notifications to be more fine grained and allow turning on/off notifications per user on a form. Note that a user should only be able to change their settings and not anyone else's notification setting on the form.

The update adds a new filed `notificationsUsers`. Users that are in this list have notifications turned on.
This field works with `users` to take a diff and create a list of users that that have the setting enabled/disabled. The `notificationsInterval` is no longer used as a flag and instead as a setting. So a DB migration is needed to retain the state (https://github.com/cds-snc/platform-forms-client/pull/5783).

## Testing

1. Create a new form.
2. Notifications should be off by default for new forms.
3. Add a few users as collaborators to the a form. Ideally some of them could be your test account users so you can receive emails from them.
4. In Preview or Published mode, do a form submission.
5. You shouldn't get any email notifications.
6. Go back to the form settings and enable notifications for the current user.
7. In an incognito tab sign in with one of the test accounts you used in the form.
8. Turn notifications on for this user also.
9. Fill out a form submission from either session (main account or test account)
10. Both users should get an email notification.
11. Try turning notifications off for both users.
12. Do another form submission
13. Neither user should get a form submission

PR #5783, opened 2025-06-27
chore: notification v2 data migration - part 1

thiessenp-cds:
# Summary | Résumé

DB data migration for #5702

This PR ensures that the notifications state (on/off) is retained for older forms. This is done by finding each form where `notificationsInterval` is on and then copying any form owners into a new field `notificationsUsers`. `notificationsInterval` is no longer being used as an on/off flag and instead as it was intended, an interval/frequency. `notificationsUsers` is now the flag for sending notifications - any user in this list will receive notifications.


prisma/seeds/seed.ts
```diff
@@ -143,6 +143,49 @@ async function deliveryOptionMigration() {
...
+      })
+    )
+  );
+
+  console.log(
+    `${templates.length} were migrated for update email notifications users that have a non null notificationsInterval`
+  );
+}
+
+// Part 2: Notifications v2 migration
```
thiessenp-cds, line 175: The notificationsInterval setting could instead be removed but if so this should also be removed in the UI and a hard coded value used consistently.

Issue #5771, opened 2025-06-25
Can still publish a Vault form when file upload is in a Repeating Set


Abi-Nada:
## Description
When I add a Repeating Set with file attachment, and then switch the form back to the "Vault", I'm still able to Publish the form.

## Steps to reproduce
1. Create a new form. Set the delivery to API.
2. Add a Repeating Set with a file attachment.
3. Go back to Settings, delete the API Key.
4. Go back to Edit, see that the element isn't in red.
5. Go to Publish, see that you can publish the form
6. See error

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Staging
- **Language**: EN

## Expected behaviour
You shouldn't be able to publish a form that has file attachments, that is set to "Download responses from GC Forms" (API only).

## Screenshots or videos

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK3VBQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e01bbfa90ac489739bd2e4e69b18ab9bd869b9f6/Screen%20Recording%202025-06-25%20at%208.42.53%E2%80%AFAM.mov)

PR #5795, opened 2025-06-30
chore: add published check for file input within a repeating set

timarney:
# Summary | Résumé

Adds check for sub-elements + file inputs 

Fixes: #5771


thiessenp-cds (approved):
LGTM :)

Issue #5760, opened 2025-06-24
File upload: Manual accessibility testing for end-user screens


Abi-Nada:
### Description:

Ensure every end-user screen and interaction created or affected by the new File Uttachment feature meets WCAG 2.1 Level AA when using the screen readers and browser combinations most commonly used by the public, so we catch real-world barriers before release.

### Acceptance Criteria:
- [ ] All discovered issues are logged as Zenhub bugs using the bug template (with screenshots, reproduction steps, severity label)

thiessenp-cds:
I found only minor errors in my initial tests. Nice work :)

1 Forms > Settings > Response Delivery > Add API Key
Once the key is created the page updates with the key Id. Is it announced? The attached video shows what I hear. The previous attempt though I thought I heard the key number announced? Either way once the key number is announced, it may help to prefix it with “key” or something similar so it’s clear what it is.

2 Forms > Edit Page > Add form Element > File Upload
In the preview window, the Chose File button has a few undefined labels. See image.

3 Forms > Edit Page > Add form Element > File Upload
After adding a file upload element, the UI has what looks like a low contrast issue. See image.

4 Form filler > Choose file > (added a file)
Remove button is missing a label. See image.

5 Form filler > Choose file
Choose file has a null reference in aria-describedby. See image.

Note: video and images are prefixed with the related number and in the attached zip

[a11y-testing.zip](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeDZHQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f9be82d5967460d5f2c869bf144ea9bd22a55848/a11y-testing.zip)

PR #5801, opened 2025-07-03
fix: a11y updates for file input

timarney:
# Summary | Résumé

Partial fixes for https://github.com/cds-snc/platform-forms-client/issues/5760

- 1 Forms > Settings > Response Delivery > Add API Key
Once the key is created the page updates with the key Id. Is it announced? The attached video shows what I hear. The previous attempt though I thought I heard the key number announced? Either way once the key number is announced, it may help to prefix it with “key” or something similar so it’s clear what it is.

- 2 Forms > Edit Page > Add form Element > File Upload
In the preview window, the Chose File button has a few undefined labels. See image.

- 4  Forms > Edit Page > Add form Element > More
File input and label id does not match. See image.

- 5 Forms > Edit Page > Add form Element > More
Hint text input and label missing relationship. See image.

- 6 Form filler > Choose file > (added a file)
Remove button is missing a label. See image.
I wonder if some kind of relationship could be created between the Remove button and the file name?

- 7 Form filler > Choose file
Choose file has a null reference in aria-describedby

### DIFF
--- a/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/components/dialogs/MoreDialog/Description.tsx
+++ b/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/components/dialogs/MoreDialog/Description.tsx
@@ -22,7 +22,7 @@ export const Description = ({
 
   return (
     <div className="mb-2">
-      <Label>{t("inputDescription")}</Label>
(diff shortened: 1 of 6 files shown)


thiessenp-cds (approved):
LGTM :)

Thanks for making these updates!

PR #5877, opened 2025-07-14
fix: update multi-page forms to update their title for sub-pages

thiessenp-cds:
# Summary | Résumé

This PR gives our forms-form pages a more unique title to help meet the below success criteria. This is more of a subjective change since the page titles were arguable already unique enough.
https://www.w3.org/WAI/WCAG22/Understanding/page-titled.html

For single-page forms only update the confirmation page title.
For multi-page forms update, on sub-pages update the page title and also the Review page and confirmation page.

# Testing 

Look at the browser title and verify the page title is updated based on the above.
Using a screen reader won't really help since a page's heading is focussed on dynamic update and read (interrupting the page title update announcement).


lib/hooks/useUpdateHeadTitle.tsx
```diff
@@ -0,0 +1,13 @@
+import { useEffect } from "react";
+
```
thiessenp-cds, line 2: @timarney do you think `<title>mytitle</title>` could work? I tried this and ran into a variety of issues

timarney: Tried and well and no doesn't work --- the title gets over-written

PR #5889, opened 2025-07-15
fix: update formatted date to define a custom widget

thiessenp-cds:
# Summary | Résumé

Update formatted date to define a custom widget. A few updates:

aria-roledescription to define the custom widget
aria-labeledby to describe the custom widget
Would give the widget an accessible name (or similar)  "date formatted date"

tabindex=-1
Adds the widget to an AT form-controls outline to allow it to be focussed. Without this it probably won't be in there. A 0 value could cause the AT to announce the contents twice. Also a user probably expects to tab right into the first form control vs. the widget label.


components/clientComponents/forms/FormattedDate/FormattedDate.tsx
```diff
@@ -113,11 +113,11 @@ export const FormattedDate = (props: FormattedDateProps): React.ReactElement =>
 
   return (
     <fieldset
-      role="group"
-      aria-describedby={description ? `desc-${id}` : undefined}
+      aria-roledescription="formatted date"
```
thiessenp-cds, line 116: add a translated string

i18n/translations/fr/common.json
```diff
@@ -194,7 +194,8 @@
     },
     "YYYY-MM-DD": "AAAA-MM-JJ",
     "DD-MM-YYYY": "JJ-MM-AAAA",
-    "MM-DD-YYYY": "MM-JJ-AAAA"
+    "MM-DD-YYYY": "MM-JJ-AAAA",
+    "roleDescription": "[FR]formatted date"
```
anikbrazeau, line 198:
```suggestion
    "roleDescription": "date mise en forme"
```

timarney (approved):

PR #5892, opened 2025-07-15
fix: update downshift combobox to use valid Ids

thiessenp-cds:
# Summary | Résumé

Update the combobox that uses downshift to use valid Ids. Previously downshift would generate Id's like `downshift-«r0»-label` that did not exist.


components/clientComponents/forms/Combobox/Combobox.tsx
```diff
@@ -20,6 +20,8 @@ export const Combobox = (props: ComboboxProps): React.ReactElement => {
   const [items, setItems] = React.useState(choices);
   const { isOpen, getMenuProps, getInputProps, highlightedIndex, getItemProps, selectedItem } =
     useCombobox({
+      // Stops downshift from generating a garbled Id. See #5834
+      id: `downshift-${id}`,
```
thiessenp-cds, line 24: ditch this

PR #5899, opened 2025-07-17
fix: update validation error list to be removed from the tab order

thiessenp-cds:
# Summary | Résumé

For forms-form error validation and similar pages like contact and support, this PR updates the validation error list to be removed from the tab order. The error list was previously in the tab order and could be confusing as a user tabs and lands on the error container (that is not a form control, link etc.).

The update changes the` tabindex=0` to `tabindex=-1` so the list can still be focussed and announced on an error but not be in the tab order.


thiessenp-cds:
@thiessenp-cds double check the GCDS error list and match that and or map it back to the GCDS

PR #5926, opened 2025-07-24
chore: add sanitize pii

timarney:
# Summary | Résumé

Adds `sanitize-pii` to contact form

Removes personally identifiable information (PII) from text.

See: https://www.npmjs.com/package/@cdssnc/sanitize-pii

current patterns https://github.com/cds-snc/sanitize-pii/blob/main/src/patterns.ts

Also fixes small issue with heading spacing.
Screenshot 2025-07-25 at 2 03 50 PM


wmoussa-gc:
nice, can we see a live test?

timarney:
> nice, can we see a live test?

Ya will test soon.

timarney:
Tested Pat is going to make a small update and I'll bump once more on this end.

PR #5965, opened 2025-08-06
fix: add keyboard control for tree view delete button

thiessenp-cds:
# Summary | Résumé

Add keyboard control for tree view delete button.

Also updates the react-complex-tree library to latest.


app/(gcforms)/[locale]/(form administration)/form-builder/components/shared/right-panel/treeview/Item.tsx
```diff
@@ -95,6 +97,14 @@ export const Item = ({
         (children as ReactElement) && context.isExpanded && "bg-slate-50",
         context.isDraggingOver && "!border-dashed !border-1 !border-blue-focus"
       )}
+      onKeyDown={(e) => {
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/components/shared/right-panel/treeview/Item.tsx:100]: Thinking we'll need to limit this to just `delete`

If you 2x click an item you go into `rename` mode --- pressing backspace to empty out the item name now pops the delete.

Issue #5964, opened 2025-08-06
Add default max limit to repeating sets (25 maximum)


anikbrazeau:
## Context

**Meeting on size limits and coverage of submission error states**
_To align/decide on a global cap for file size/number_
https://docs.google.com/document/d/1X20UXkRzD5M0GWW_DzHukedTETzHpR5ksX-iw8F8B3o/edit?tab=t.0

## Description

In an effort to limit the number of files that can be added to a form submission, we've decided to start with a 25-instance limit for repeating sets to help cap the amount of data and number of files (10MB max per file) that can be added to a form submission -- for the first version of file attachments. 

**This will add a default value to the "Maximum number of answers" for "Repeating sets" in the "More" modal**

We can re-calibrate these maximums if we learn more from clients or from our costing estimate calculations.

## Screenshot

**New default value set as maximum number of answers**

<img width="495" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN2YzQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--922abb1ecff36095801ae75fa4abaeca25ff0cd3/Screenshot%202025-08-06%20at%2011.36.43%E2%80%AFAM.png" alt="Screenshot 2025-08-06 at 11.36.43 AM.png" />

jprince-cds:
@anikbrazeau We have received the following feedback from a client [(Freshdesk ticket 24088)](https://cds-snc.freshdesk.com/a/tickets/24088):

> Our department’s use of the repeating set often requires the ability for clients to add more than 25 entries, so this has greatly affected our ability to continue using GCForms as a solution.
>  
> Could the development team consider limiting repeating sets only when they have files, or setting a global file count limit, rather than limit repeating sets that only contain normal questions?

PR #5966, opened 2025-08-07
chore: add max row count for repeating sets

timarney:
# Summary | Résumé

Closes: https://github.com/cds-snc/platform-forms-client/issues/5964

- Updates the form builder to ensure a max row count is set when adding a repeating set.
- Defaults max row count for repeating sets to 25

Notes: 

In a future PR we might want to update front facing form to ensure a max rows property is set regardless of the template value.  This would handle cases where the property isn't set and also legacy templates.

This PR doesn't alter existing templates i.e. if a form already has created a repeating set.

https://github.com/user-attachments/assets/12ab84df-8af1-4697-a68d-984b452ea75f

### DIFF
--- a/i18n/translations/en/form-builder.json
+++ b/i18n/translations/en/form-builder.json
@@ -580,6 +580,9 @@
       "title": "Additional tags",
       "description": "Add flexible labels to include metadata for form elements so that related questions can be categorized or so that data can be transformed.",
       "errorTagTooLong": "Tag cannot exceed 50 characters"
+    },
+    "dynamicRow": {
+      "errorMaxRows": "Maximum row limit cannot exceed {{maxRows}}."
     }
   },
   "moreOptions": "More options",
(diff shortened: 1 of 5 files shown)


i18n/translations/en/form-builder.json
```diff
@@ -580,6 +580,9 @@
       "title": "Additional tags",
       "description": "Add flexible labels to include metadata for form elements so that related questions can be categorized or so that data can be transformed.",
       "errorTagTooLong": "Tag cannot exceed 50 characters"
+    },
+    "dynamicRow": {
+      "errorMaxRows": "Maximum row limit cannot exceed {{maxRows}}."
```
anikbrazeau, line 585:
```suggestion
    },
    "dynamicRow": {
      "errorMaxRows": "The maximum number of answers cannot exceed {{maxRows}}."
```

ShadeWyrm (approved):
LGTM, works as expected.

PR #6011, opened 2025-08-14
fix: add more semantics for when a file input is invalid in the form-builder

thiessenp-cds:
# Summary | Résumé

Previously in the form-builder when a file input did not have an API key it was flagged as invalid using some styled text. This PR adds a strong element around the style text and an aria-invalid="true" in the related form input to help AT users know there is a problem with the file input.


thiessenp-cds:
> Will approve as this is all working
> 
> This line isn't quite sitting right sitting where it is but it's fine --- wondering if it should be higher up or something -- fine for now.
> 
> // const isInvalid = isFileUpload && !hasApiKeyId;

Sure could move it up to the ElementPanel.tsx file or did you have another place/method in mind?

PR #6018, opened 2025-08-15
feature: beta permissions

bryan-robitaille:
# Summary | Résumé
1. Provides a method to ensure users can not mistakenly use beta components in their forms.
2. Updates the FeatureFlagsProvider with an update option so that mutations on the flags can be propagated through the react tree.  With the provider was only getting it's values at rendering from the server.
3. Removed authorization check from `checkAll` flags function.  The data is publicaly available in the page data and there's nothing secret there anyway.

Note that users who have required flags or once the flag is set globally do not encounter these errors

Authenticated User trying to import a JSON that has beta components
Screenshot 2025-08-15 at 3 34 09 PM

Unauthenticated User trying to import a JSON that has beta components
Screenshot 2025-08-15 at 3 34 24 PM

Authenticated user clicking on edit form from Forms page.
Screenshot 2025-08-15 at 3 35 33 PM

Authenticated User:
- Give yourself user level flag rights for beta component
- Create a form with beta components
- Save form.
- Remove user level flag rights for beta component
- return to forms page and attempt to open form

Unauthenticated User:
- Attempt to import form JSON containing beta components from Start page

Authenticated user with out of sync client / server
- Give yourself user level flag rights for beta component
- Create a form with beta components
- Save form.
- Open second Tab
- Remove user level flag rights for beta component
- Return to first tab
- Add or remove an element in the form
- Save Form
- Should show an error that the form could not be saved and to contact support.

links in this section

# Pull Request Checklist


lib/templates.ts
```diff
@@ -1424,3 +1432,45 @@ export const checkIfClosed = async (formId: string) => {
     return null;
   }
 };
+const checkForBetaComponents = async (elements: FormElement[]) => {
```
timarney, line 1435: Can we use the ` checkForBetaComponents ` from the store level here?  

Looking to have 1 "checker" function vs the current 2

timarney:
Wondering about the something went wrong flow ... 

It fully takes you out of edit mode and provides no opportunity to remove the component --- if someone has worked on or is editing a large form and we basically kick them out how can they correct the situation --- they can't provide a form file to support etc...

Should we keep them in the editor but fail the save?

I haven't thought about all the possible scenarios but flagging as something that should be considered.

lib/templates.ts
bryan-robitaille: We could.  I was just trying to separate the concerns between client and sever since the Store is completely client based.  What I could do is pull the betaCheck out of the Store folder structure and find a safe place for lib's that support both client and server.

bryan-robitaille:
> It fully takes you out of edit mode and provides no opportunity to remove the component --- if someone has worked on or is editing a large form and we basically kick them out how can they correct the situation --- they can't provide a form file to support etc...

The `Something went wrong` flow should only be triggered by clicking on the 'edit form' link from the forms page.  So they're not quite in editing mode yet so there's no work to save.  If they are editing a form and the flag changes they do get the `Something went wrong` but instead the error that the form could not be saved.  There's currently no reference as to why it couldn't be saved as the error that fails on the server isn't passed through as Client Readable..

If you triggered the `Something went wrong` from the edit page please let me know because that shouldn't happen.

timarney (dismissed):
Approving --- as is and this is fully functional and tests are all passing now.

lib/templates.ts
```diff
@@ -148,7 +155,7 @@ export async function createTemplate(command: CreateTemplateCommand): Promise<Fo
     throw e;
   });
 
-  await checkForBetaComponents(command.formConfig.elements);
+  checkForBetaComponentsAsync(command.formConfig.elements, checkFlag);
```
timarney, line 158: Are we good skipping the await here ... and expecting the throw to manage ?

timarney (approved):

timarney (approved):

PR #6060, opened 2025-08-26
chore: update max row count

timarney:
# Summary | Résumé

Adjusts the max row count for repeating sets


anikbrazeau (approved):
LGTM, thanks Tim!

PR #6073, opened 2025-08-27
chore: Revert formattedDate deserialization

dsamojlenko:
# Summary | Résumé

Reverting date deserialization. Leaving some code and placeholders in place for future.


app/(gcforms)/[locale]/(form filler)/id/[...props]/lib/server/normalizeFormResponses.ts
```diff
@@ -125,7 +125,7 @@ export const fillData = (
         return fileInputFiller(value as Response);
       case FormElementTypes.formattedDate:
         if (typeof value === "string") {
-          return deserializeDateObject(value);
+          // return deserializeDateObject(value);
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form filler)/id/[...props]/lib/server/normalizeFormResponses.ts:128]: Maybe just add a comment here saying this will be uncommented in a future breaking change or something.

PR #6076, opened 2025-08-27
feat: Negative Numbers Functionality

### LINKED ISSUE #5881: Ability to enter a negative number
## Description

Currently, the numeric field / "number" form element does not allow an entry that is not a number (ex: letters). However, it also does not allow for a negative number to be entered.

Support for negative characters has also been flagged by Growth as a need for a client at TC working on a G&C application. 

## Expected

A negative number should be an acceptable valid number. Use cases where this could apply include a reimbursement, tax form, or other monetary input where the value can be positive or negative.

ShadeWyrm:
Closes #5881

Adds the ability to toggle negative numbers on Form Numeric fields, with appropriate checks if on or off.


i18n/translations/fr/form-builder.json
```diff
@@ -256,7 +256,8 @@
       "description": "Un champ qui n’accepte qu’une valeur numérique.",
       "title": "Chiffre",
       "amount": "Montant",
-      "enterOnlyNumbers": "Saisissez un chiffre."
+      "enterOnlyNumbers": "Saisissez un chiffre.",
+      "allowNegative": "Autoriser les nombres négatifs"
```
anikbrazeau, line 260:
```suggestion
      "allowNegative": "permettre les valeurs numériques négatives"
```

anikbrazeau (approved):
Tested the preview env. and looks great from experience perspective

Issue #6041, opened 2025-08-22
"Skip to pages set-up" links at the top and bottom of the page do not work


lauren-baroud:
## Description
When editing the form, the "Skip to pages set-up" links at the top and at the bottom of the page do not work as expected

## Steps to reproduce
1. Go to any page editor
2. Begin tabbing through the page using a keyboard until the "Skip to page set-up" link appears
3. Hit enter
4. Link does not work
5. Continue to tab to the bottom to get to the bottom "Skip to page set-up" link
6. Hit enter
7. Link does not work

## Details
- **Browser**: Any
- **Operating system**: Any
- **Environment**: Production
- **Language**: English, but could be both

## Expected behaviour
The link should take the user to the Page set-up menu

## Screenshots or videos
<img width="856" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBLzRtQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2027f2975bdba069bb6cf066cf1558645e991aac/Screenshot%202025-08-22%20at%202.04.02%E2%80%AFPM.png" alt="Screenshot 2025-08-22 at 2.04.02 PM.png" />

anikbrazeau:
https://gcdigital.slack.com/archives/C01GFD150MC/p1755800732807669

PR #6089, opened 2025-08-29
fix: skip to right panel

timarney:
# Summary | Résumé

Update to ensure id exists when panel is in an open or a closed state.

Fixes: #6041

https://github.com/user-attachments/assets/0e598809-512e-418f-8255-fa5d2a8727c7


thiessenp-cds (approved):
Tested in VO, Jaws and NVDA

LGTM :)

PR #6123, opened 2025-09-16
chore: wrap waf check

timarney:
# Summary | Résumé

Adds a try / catch to waf check to avoid built time errors when headers are not available

```
N: fetch failed. Read more at https://errors.authjs.dev#autherror
    at P (.next/server/chunks/280.js:1:55849)
    at async $ (.next/server/chunks/280.js:1:57437) {
  type: 'AuthError',
  kind: 'error'
```

### DIFF
--- a/lib/auth/nextAuth.ts
+++ b/lib/auth/nextAuth.ts
@@ -174,12 +174,17 @@ const {
         return;
       }
 
-      const requestHeaders = await headers();
+      try {
+        const requestHeaders = await headers();
 
-      if (requestHeaders.get("x-amzn-waf-cognito-login-outside-of-canada")) {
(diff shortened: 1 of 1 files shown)


timarney:
> While generating static pages, Next.js will throw a DynamicServerError if it detects usage of a dynamic function, and catch it to automatically opt the page into dynamic rendering. However, when it's uncaught, it will result in this build-time error.


Pushed another commit which I was debating including in this PR but let's have the discussion :) 

I think given the use of headers the static pages are not really static so we might as well just use 

```
export const dynamic = "force-dynamic";
```

Thoughts --- with that the built warnings go away making it easier in my opinion to spot if we actually do have a build error

PR #6125, opened 2025-09-18
chore: Update i18n package

timarney:
# Summary | Résumé

Bumps i18n packages to latest


bryan-robitaille (approved):
Checked the migration guide and we should be good to jump to this major version.
We're already using the I18Next JSON v4 format.
[I18Next Upgrade Guide](https://www.i18next.com/misc/migration-guide)

PR #6164, opened 2025-10-01
fix : Accessibility fix for en/fr Toggle for Screen Readers

### LINKED ISSUE #3097: Language toggle link using the lang attribute has an issue
**HC Assessment Label: Serious**

Pages:  Advancing Accessibility Standards Research Program Expression of Interest,
Feedback on our Accessibility,
Thank you for submitting an expression of interest,
Thank you for your feedback
URL: https://forms-staging.cdssandbox.xyz/en/id/134
https://forms-formulaires.alpha.canada.ca/en/id/clq1bz2xl005ax382blj94w82
https://forms-staging.cdssandbox.xyz/en/id/134/confirmation
https://forms-formulaires.alpha.canada.ca/id/clq1bz2xl005ax382blj94w82/confirmation
User agent:  Chrome (v119.0.6045.200), JAWS2020, NVDA 2023.1
Language: English and French

Issue Description: 
The `"<a>"` element associated with the language toggle link contains the "aria-label" attribute, In this case, the value in the "aria-label" attribute becomes the accessible name for the link and screen readers will announce that as link description instead of the visual text for the link. However, the "aria-label" attribute value on the English page contains both English and French text and vice versa on the French page. In this case, the "lang" attribute doesn't correctly identify the change in language in the document with the value "fr" and "en" used in the `"<a>" `element. As a result, it causes the screen readers to announce the English text as French text and vice versa. 

**English:**
```
aria-label="Language selection: Français" lang="fr"
<div class="inline-flex gap-4"><a href="/fr/id/134" class="text-base text-right " aria-label="Language selection: Français" lang="fr">Français</a></div>
```
**French:**
```
aria-label="Sélection de la langue: English" lang="en"
<div class="inline-flex gap-4"><a href="/id/134" class="text-base text-right " aria-label="Sélection de la langue: English" lang="en">English</a></div>
```

Steps to reproduce:

1. Run the NVDA screen reader
2. Open the above URL in the Chrome browser
3. Navigate using the "Tab" key
4. Listen to what the screen reader reads when the language toggle link receives focus

Expected result: In this case, the `"aria-label" `attribute should be eliminated in the `"<a>" `element to avoid the above issue. The text "language selection" should be coded as a heading similar to how it is implemented in the Canada.ca template to provide more context to screen reader users. For example, https://wet-boew.github.io/GCWeb/sites/header/header-docs-en.html

ShadeWyrm:
Closes #3097

### DIFF
--- a/cypress/e2e/your_account_user_profile.cy.ts
+++ b/cypress/e2e/your_account_user_profile.cy.ts
@@ -10,8 +10,8 @@ describe("User profile", () => {
       cy.get("[data-testid='yourAccountDropdownContent']").contains("Profile").click();
       cy.url().should("include", "/profile");
       cy.get("h1").contains("Profile");
-      cy.get("h2").first().contains("Account information");
-      cy.get("h2").eq(1).contains("Security questions");
+      cy.get("h2").eq(1).contains("Account information");
+      cy.get("h2").eq(2).contains("Security questions");
     });
     it("Redirects to user profile page if security questions are already completed", () => {
       cy.visitPage("/en/auth/setup-security-questions");
@@ -41,8 +41,8 @@ describe("User profile", () => {
       cy.url().should("include", "/profile");
 
       cy.get("h1").contains("Profile");
-      cy.get("h2").first().contains("Account information");
-      cy.get("h2").eq(1).contains("Security questions");
+      cy.get("h2").eq(1).contains("Account information");
+      cy.get("h2").eq(2).contains("Security questions");
     });
     it("Renders the My Account dropdown as admin", () => {
       cy.visitPage("/en/forms");


components/clientComponents/globals/Header/LanguageToggle.tsx
```diff
@@ -28,15 +28,20 @@ const LanguageToggle = () => {
...
-      {currentLang === "en" ? "Français" : "English"}
-    </a>
+    <div className="gc-lang-toggle-link text-right text-base">
+      <span className="sr-only" lang={currentLang}>
+        {t("lang-toggle")}:{" "}
+      </span>
+      <a
+        href={href}
+        className="text-right text-base"
+        aria-label={currentLang == "en" ? "Français" : "English"}
```
thiessenp-cds, line 38: I think this aria-label is now redundant so it can be removed.

thiessenp-cds (approved):
LGTM :)

PR #6168, opened 2025-10-02
Chore: Replace formBuilder.tsx with Component Model

ShadeWyrm:
A refactor of the formBuilder.tsx to adopt a component model with the final intent being components (eg: TextField, Checkbox, AddressComplete, FileInput, etc...) are fully self contained, and logic is not spread across the application.

This PR is an early DRAFT for review to adopt any style changes the team is looking for before pushing a full change.

### DIFF
--- a/components/clientComponents/forms/TextInput/TextInputPlugin.tsx
+++ b/components/clientComponents/forms/TextInput/TextInputPlugin.tsx
@@ -1,0 +1,82 @@
+// Update the path below to the correct location of ComponentPlugin
+import { ComponentPlugin } from "@lib/components/ComponentPlugin";
+import { TextInput } from "./TextInput";
+import { FormElementTypes } from "@lib/types";
+import { getLocalizedProperty } from "@lib/utils";
+import { Label, Description } from "@clientComponents/forms";
+
+export const TextInputPlugin: ComponentPlugin = {
+  meta: {
+    type: FormElementTypes.textField,
+  },
+  render({ element, lang }) {
+    const id = element.id;
+    const isRequired = !!element.properties.required;
+
+    const descriptionPerLocale = element.properties[getLocalizedProperty("description", lang)];
+    const description = descriptionPerLocale ? descriptionPerLocale.toString() : "";
+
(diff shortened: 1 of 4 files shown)


components/clientComponents/forms/TextInput/TextInputPlugin.tsx
```diff
@@ -0,0 +1,84 @@
...
+  meta: {
+    type: FormElementTypes.textField,
+  },
+  render({ element, lang }) {
+    const id = element.subId ?? element.id;
+    const isRequired: boolean = element.properties.validation
+      ? element.properties.validation.required
+      : false;
+
+    const descriptionPerLocale = element.properties[getLocalizedProperty("description", lang)];
```
timarney, line 18: Maybe we should have helpers for pulling descriptions titles etc...  ?

Just thinking about how repetitive this is going to get.

ShadeWyrm: Same with the Label, I've just been thinking through how I want to do it in case we need to break a label out of an element in other ways.

components/clientComponents/forms/TextInput/TextInputPlugin.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { TextInput } from "./TextInput";
+import { FormElementTypes } from "@lib/types";
+import { getLocalizedProperty } from "@lib/utils";
+import { Label, Description } from "@clientComponents/forms";
+
+export const TextInputPlugin: ComponentPlugin = {
+  meta: {
+    type: FormElementTypes.textField,
+  },
+  render({ element, lang }) {
```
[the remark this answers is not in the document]
timarney: We can save it for the next one :)

components/clientComponents/forms/TextInput/TextInputPlugin.tsx
ShadeWyrm: I went with just making a linked component, but it feels a bit ugly. I'll keep thinking, and if anyone has a better idea.

ShadeWyrm:
Component changes have made the pattern have some issues with how Formik is being used, would want to re-apply the pattern on a clean branch.

PR #6183, opened 2025-10-08
chore: reduce db lookups done by notifications

thiessenp-cds:
# Summary | Résumé

Previously Notifications did a db lookup for delivery options to determine whether or not to send a notification. This was done to keep the code as isolated as possible. The downside though is that deliveryOption is only used by legacy forms so the db call is not needed for most forms.

The update alters the forms template call to include deliveryOption in the "public properties" and passes this down to notifications.

## Testing

Try sending a few form submissions on a form. You should receive notifications.

Create a new form and set the form to receive email delivery (manually). Send a few form submissions. You should not receive notifications.

This is probably easiest to test locally by updating a template through prisma. For example.
Screenshot 2025-10-08 at 9 42 37 AM

To know whether you successfully added a deliveryOption you should see an email delivery option in the form setting.
Screenshot 2025-10-08 at 9 44 06 AM

### DIFF
--- a/app/(gcforms)/[locale]/(form filler)/id/[...props]/actions.ts
+++ b/app/(gcforms)/[locale]/(form filler)/id/[...props]/actions.ts
@@ -116,7 +116,12 @@ export async function submitForm(
       language,
     });
 
-    sendNotifications(formId, template.form.titleEn, template.form.titleFr);
+    sendNotifications(
+      formId,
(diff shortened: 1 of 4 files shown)


thiessenp-cds:
deliveryOption should not be made public so an extra db lookup is required (existing code). Perhaps this can be solved with a form version number in the future

PR #6229, opened 2025-10-15
feat: Ability to Archive Templates

### LINKED ISSUE #6223: Key actions: Delete -> "Archive"
As a form owner, how might I know when a form was deleted and by whom?

For the user story above, we need the form to be accessible in the UI in order for anyone to grab the logs. The following flow introduces the idea that forms could be archived then deleted after 30 days, so that users can grab the form activity. 

PROPOSED: New "Archive" function.

Challenge: If we want to see who deleted the form, the form still has to exist in the interface. I propose removing "delete" and adding "archive" function:

- User archives a form
- An "archive" folder appears in their /Forms page
- If the form is in the archive folder, they have 30 days to retrieve it before it is fully deleted
- If they select the form in the archive folder, it opens it again in "Draft" mode and interface
- BONUS: they can republish it with the caveat that they need to add a new API key and add any collaborators post publishing

If we don't want to add the ability to make a change on a published form, the link would just open the "Published" interface we have today.

ShadeWyrm:
Repurposes the soft-delete functionality as archive functionality.

Closes #6223 

Deferred issues:
https://github.com/cds-snc/platform-forms-client/issues/6289
https://github.com/cds-snc/platform-forms-client/issues/6290
https://github.com/cds-snc/platform-forms-client/issues/6291
https://github.com/cds-snc/platform-forms-client/issues/6292

### DIFF
--- a/lib/tests/templates.test.ts
+++ b/lib/tests/templates.test.ts
@@ -248,6 +248,7 @@ describe("Template CRUD functions", () => {
         expect.objectContaining({
           where: {
             id: "formTestID",
+            ttl: null,
           },
         })
       );
@@ -324,7 +325,7 @@ describe("Template CRUD functions", () => {
         expect.objectContaining({
           where: {
             id: "test1",
-            isPublished: false,
+            isPublished: false
           },
           data: {
             jsonConfig: updatedFormConfig as unknown as Prisma.JsonObject,


lib/integration/prismaConnector.ts
```diff
@@ -22,7 +22,7 @@ const prismaClientSingleton = () => {
       template: {
         $allOperations({ args, query }) {
           if ("where" in args) {
-            args.where = { ...args.where, ttl: null };
+            args.where = { ...args.where };
```
timarney: @thiessenp-cds 

Looks like you have the most context on the `ttl` here ...

https://github.com/cds-snc/platform-forms-client/pull/5339/files

Concerns with removing?

app/(gcforms)/[locale]/(form administration)/forms/components/client/Unarchive.tsx
```diff
@@ -0,0 +1,52 @@
+"use client";
+import { useTranslation } from "@i18n/client";
+import { clearTemplateStorage } from "@lib/store/utils";
+import { cloneForm, restoreForm } from "../../actions";
+import { toast, ToastContainer } from "@formBuilder/components/shared/Toast";
+
+export const Unarchive = ({
```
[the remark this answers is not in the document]
Comment (ShadeWyrm) [reply app/(gcforms)/[locale]/(form administration)/forms/components/client/Unarchive.tsx]: I'd agree, can add as another option in the menu. @samsadasivan - think that makes sense to do?

lib/templates.ts
```diff
@@ -1395,6 +1398,71 @@ export async function deleteTemplate(formID: string): Promise<FormRecord | null>
...
+        securityAttribute: true,
+        formPurpose: true,
+        publishReason: true,
+        publishFormType: true,
+        publishDesc: true,
+        saveAndResume: true,
+        notificationsInterval: true,
+      },
+    })
+    .catch((e) => prismaErrors(e, null));
```
thiessenp-cds: I wonder if we could combine the two prisma queries. So remove the first query and add the isPublished check to the above query. Though I guess the first query is more specific, just finding the template vs. the second that finds that template with no ttl. More specific errors. Just thinking out loud :) Probably good as is.

i18n/translations/fr/my-forms.json
```diff
@@ -3,11 +3,14 @@
   "nav": {
     "drafts": "Ébauches",
     "published": "Publiés",
```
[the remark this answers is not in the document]
ShadeWyrm: Sorry, not sure I follow - where should that string go in the UI?

i18n/translations/fr/my-forms.json
```diff
@@ -3,11 +3,14 @@
...
-    "all": "Tous"
+    "all": "Tous",
+    "archived": "Archivé"
```
anikbrazeau:
```suggestion
    "archived": "Archivés"
```

i18n/translations/fr/my-forms.json
```diff
@@ -3,11 +3,14 @@
...
   },
   "actions": {
     "resumeForm": "Reprendre la rédaction de formulaire en cours",
-    "createNewForm": "Créer un nouveau formulaire"
+    "createNewForm": "Créer un nouveau formulaire",
+    "unarchiveForm": "Restaurer le formulaire",
```
anikbrazeau:
```suggestion
    "unarchiveForm": "Désarchiver le formulaire",
```

i18n/translations/en/my-forms.json
```diff
@@ -3,11 +3,14 @@
...
-    "all": "All"
+    "all": "All",
+    "archived": "Archived"
   },
   "actions": {
     "resumeForm": "Resume editing form",
-    "createNewForm": "Create new form"
+    "createNewForm": "Create new form",
+    "unarchiveForm": "Restore form",
+    "duplicatePublishedForm": "Duplicate published form"
```
anikbrazeau: Question: Why is the action here to duplicate (archived, published) rather than restore?

i18n/translations/fr/form-builder.json
```diff
@@ -477,16 +477,17 @@
...
+      "message1": "La suppression d'un formulaire ne peut être annulée. Votre formulaire brouillon sera définitivement supprimé de Formulaires GC.",
+      "message2": "Téléchargez une copie du fichier de formulaire pour vos archives. Vous pourrez ensuite le téléverser à nouveau dans Formulaires GC pour recréer un nouveau formulaire avec de nouveaux liens URL.",
+      "message3": "Avant d'archiver votre formulaire, assurez-vous de supprimer tous les liens publics menant à ce formulaire. Cela permet d'éviter d'envoyer les gens dans une impasse. Pour arrêter les soumissions, vous pouvez plutôt considérer la fermeture du formulaire dans les paramètres."
     },
-    "okay": "Supprimer le formulaire",
+    "okay": "Archiver le formulaire",
     "published": {
       "message1": "La suppression d'un formulaire ne peut être annulée. Votre formulaire publié sera définitivement supprimé de l'Internet, ce qui risque de créer des liens brisés.",
       "message2": "Téléchargez une copie du fichier de formulaire pour vos archives. Vous pourrez ensuite le téléverser à nouveau dans Formulaires GC pour recréer un nouveau formulaire avec de nouveaux liens URL.",
```
anikbrazeau:
```suggestion
      "message1": "L'archivage de ce formulaire publié ne peut être annulée. Votre formulaire publié sera définitivement enlevé de l'Internet, ce qui risque de créer des liens brisés.",
      "message2": "Téléchargez une copie du fichier de formulaire pour vos registres. Vous pourrez ensuite le téléverser à nouveau dans Formulaires GC pour recréer un nouveau formulaire avec de nouveaux liens URL.",
```

i18n/translations/en/my-forms.json
ShadeWyrm: Initial iteration was that they'd duplicate a published form instead of bringing it back. eg: Fresh and clean copy. Could be changed to whatever implementation we'd like.

i18n/translations/en/form-builder.json
```diff
@@ -477,16 +477,17 @@
...
+      "message3": "Download a copy of the form file for your records. You can upload it again in GC Forms to re-create the form later."
     },
-    "okay": "Delete form",
+    "okay": "Archive form",
     "published": {
-      "message1": "Deleting a form cannot be undone. Your published form will be removed from the Internet, potentially creating broken links.",
-      "message2": "Download a copy of the form file for your records. You can then upload it in GC Forms to re-create a new form with new URL links.",
-      "message3": "Before deleting your form, make sure you remove any public links leading to this form. This avoids sending people to a dead end. To stop submissions, you might instead consider closing the form in Settings."
+      "message1": "Archiving will move this published form to the “Archived” section of your forms.",
+      "message2": "Archive and edit your form, to update an existing form instead. The older version will temporarily remain available in the archive folder.",
```
anikbrazeau:
```suggestion
      "message2": "Archive and duplicate your form, to create a new draft from the archived form that can be edited as a new version. The older version will temporarily remain available in the archive folder.",
```

i18n/translations/fr/my-forms.json
```diff
@@ -3,11 +3,14 @@
...
-    "all": "Tous"
+    "all": "Tous",
+    "archived": "Archivés"
   },
   "actions": {
     "resumeForm": "Reprendre la rédaction de formulaire en cours",
-    "createNewForm": "Créer un nouveau formulaire"
+    "createNewForm": "Créer un nouveau formulaire",
+    "unarchiveForm": "Désarchiver le formulaire",
+    "duplicatePublishedForm": "Dupliquer le formulaire publié"
```
anikbrazeau: Where does this get duplicated?
 - in what section/folder --> draft or archived? 
 - in what state --> draft or published?

timarney:
@ShadeWyrm you mentioned a few times --- believe it was for this PR that there we're some comments / asks that can be broken into another PR --- if so what should be broken out (or was that re: the log PR?)

ShadeWyrm:
> @ShadeWyrm you mentioned a few times --- believe it was for this PR that there we're some comments / asks that can be broken into another PR --- if so what should be broken out (or was that re: the log PR?)

That's this one, I'll go through and make them.

i18n/translations/fr/my-forms.json
```diff
@@ -3,11 +3,14 @@
...
     "published": "Publiés",
-    "all": "Tous"
+    "all": "Tous",
+    "archived": "Archivés"
   },
   "actions": {
     "resumeForm": "Reprendre la rédaction de formulaire en cours",
-    "createNewForm": "Créer un nouveau formulaire"
+    "createNewForm": "Créer un nouveau formulaire",
+    "unarchiveForm": "Désarchiver le formulaire",
```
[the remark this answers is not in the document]
ShadeWyrm: Design deferred to another card : https://github.com/cds-snc/platform-forms-client/issues/6289

i18n/translations/en/my-forms.json
```diff
@@ -96,6 +104,9 @@
   "navLabel": "Forms navigation",
   "errors": {
     "formDownloadNotExist": "The form you are trying to download does not exist.",
-    "formDownloadFailed": "Failed to download the form."
-  }
+    "formDownloadFailed": "Failed to download the form.",
+    "formUnarchiveFailed": "Failed to restore the form."
+  },
+  "archivedNotice": "You can access your archived forms for 30 days.",
```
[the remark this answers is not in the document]
ShadeWyrm: Deferred to : https://github.com/cds-snc/platform-forms-client/issues/6292

lib/templates.ts
```diff
@@ -1395,6 +1398,71 @@ export async function deleteTemplate(formID: string): Promise<FormRecord | null>
   return _parseTemplate(templateMarkedAsDeleted);
 }
 
+/**
+ * Restores a form template from the archived state.
+ * @param formID ID of the form template
+ * @returns A boolean status if operation is sucessful
+ */
+export async function restoreTemplate(formID: string): Promise<FormRecord | null> {
```
craigzour: Now that we have `restoreTemplate` we should probably rename `deleteForm` (with associated logs/errors...) to `archiveForm`.

lib/templates.ts
```diff
@@ -1395,6 +1398,71 @@ export async function deleteTemplate(formID: string): Promise<FormRecord | null>
...
+    logEvent(
+      e.user.id,
+      { type: "Form", id: formID },
+      "AccessDenied",
+      "Attempted to unarchive Form"
+    );
+    throw e;
+  });
+
+  // Check if the form is draft or not.
```
craigzour: I may be wrong but I feel like this comment does not reflect what is actually being achieved below. Especially considering the thrown error is `TemplateNotFoundError`.

lib/templates.ts
```diff
@@ -1395,6 +1398,71 @@ export async function deleteTemplate(formID: string): Promise<FormRecord | null>
...
+  const template = await prisma.template.findFirstOrThrow({
+    where: {
+      id: formID,
```
craigzour: If only archived forms can be restored I think we should add a condition on `ttl` not being null to throw an early error.

lib/integration/prismaConnector.ts
```diff
@@ -22,7 +22,7 @@ const prismaClientSingleton = () => {
       template: {
         $allOperations({ args, query }) {
           if ("where" in args) {
-            args.where = { ...args.where, ttl: null };
+            args.where = { ttl: null, ...args.where };
```
craigzour: With this change, some of the permission checks (related to templates) that are executed on archived forms are throwing errors because Prisma cannot find them since they have a non null `ttl` value.

This is due to the code that retrieves the template in the permission check logic not having `ttl` in the `where` clause:

```typescript
case "FormRecord":
      return prisma.template.findUniqueOrThrow({
        where: {
          id: subject.id,
        },
        select: {
          id: true,
          created_at: true,
          updated_at: true,
          name: true,
          isPublished: true,
          users: {
            select: {
              id: true,
              name: true,
              email: true,
              active: true,
            },
          },
        },
      });
```

I am guessing we can probably add it in there but we will need to make sure it does not break any other part of the code where permission checks on form subject are used.

craigzour:
@samsadasivan On the "my forms" main page, should not we also display archived forms when "All" is the selected filter at the top of the page (it is not the case at the moment)?

ShadeWyrm:
> @samsadasivan On the "my forms" main page, should not we also display archived forms when "All" is the selected filter at the top of the page (it is not the case at the moment)?

This wouldn't match other apps around the web. (eg: Especially Email)

samsadasivan:
Looks great so far! Two small things: 

- A bug with the deleted date: I think it is a month off. It said I deleted the form on "2025-12-12" when I think it meant 11 for the month. Perhaps we would also change the "delete" to "Archived on" :) 



- I archived a published form on October 19th (in the screenshot it says november but I think its october) and its throwing an error when I wanted to duplicate it. I initially thought it was because it was past the 30 days, but I think it may be something else

ShadeWyrm:
> Looks great so far! Two small things:
> 
> * A bug with the deleted date: I think it is a month off. It said I deleted the form on "2025-12-12" when I think it meant 11 for the month. Perhaps we would also change the "delete" to "Archived on" :)

Is the date it will be deleted, not the date it was archived. :) 

> 
> * I archived a published form on October 19th (in the screenshot it says november but I think its october) and its throwing an error when I wanted to duplicate it. I initially thought it was because it was past the 30 days, but I think it may be something else


Fixed!~

samsadasivan:
My bad on the date haha 

Awesome! Ok ill re-test

craigzour (approved):
LGTM!

PR #6233, opened 2025-10-17
chore: Package Prisma into own NPM package

bryan-robitaille:
# Summary | Résumé

Context: https://github.com/cds-snc/platform-forms-client/issues/6731

- Rename `.env` to `.env.yarn`.  Yarn v4 will automatically export the values into environment variables so we no longer need to rely on `dot-env` package.  This is required now that Prisma no longer leverages the dot-env package and we need to get the database_url into process.env
- Move and isolate the Prisma package from the application code into it's own NPM package.  This will allow for Prisma to be used across out platform in the different services (Lambda, API Server) and ensure easy access to Models and their types.

Should be no change to application functionality in using the Database

# Note

Inform the development team about the need to rename `.env` to `.env.yarn` and also to make sure they update the `DATABASE_URL` to have `?connect_timeout=30&pool_timeout=30&sslmode=prefer&uselibpqcompat=true` at the end of the string


craigzour (dismissed):
LGTM! Nice work 🙂

craigzour (approved):

PR #6270, opened 2025-10-28
chore: Move values-to-types to core

timarney:
# Summary | Résumé

- Moves `valueMatchesType` to core package
- Removes `validateResponses` as a standalone function -- checks now happen within the validate submit action.
- Moves `isValidEmail` to core package


lib/tests/validation/validation.test.js
```diff
@@ -126,7 +126,7 @@ const testCases = [
   },
   {
     fieldType: "radio",
-    passConditions: [{ 1: ["myChoice"] }],
+    passConditions: [{ 1: "myChoice" }],
```
timarney: Radios are strings

PR #6272, opened 2025-10-28
chore: update the settings page to allow branding changes on a published form

thiessenp-cds:
# Summary | Résumé

Updates the Settings page to allow branding changes on a published form. 

Previously the template update call would only allow published forms to be updated (query restriction). This PR adds the ability to update the branding on a published form by adding a new get and set jsonConfig method and server action.

Related Freshdesk ticket: https://cds-snc.freshdesk.com/a/tickets/25159

### DIFF
--- a/lib/templates.ts
+++ b/lib/templates.ts
@@ -1666,7 +1666,7 @@ export const updateFormJsonConfig = async (formId: string, jsonConfig: FormPrope
 
   if (formCache.cacheAvailable) formCache.invalidate(formId);
 
-  logEvent(user.id, { type: "Form", id: formId }, "UpdateFormJSONConfig");
+  logEvent(user.id, { type: "Form", id: formId }, "UpdateFormJsonConfig");
 
   return _parseTemplate(updatedTemplate);
 };
--- a/lib/auditLogs.ts
+++ b/lib/auditLogs.ts
@@ -18,7 +18,7 @@ export enum AuditLogEvent {
   RevokeFormAccess = "RevokeFormAccess",
   UpdateNotificationsInterval = "UpdateNotificationsInterval",
   UpdateNotificationsUserSetting = "UpdateNotificationsUserSetting",
-  UpdateFormJSONConfig = "updateFormJSONConfig",
+  UpdateFormJsonConfig = "updateFormJsonConfig",
   // Invitations
   InvitationCreated = "InvitationCreated",
   InvitationAccepted = "InvitationAccepted",


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/components/branding/Branding.tsx
```diff
@@ -68,6 +66,7 @@ export const Branding = ({ hasBrandingRequestForm }: { hasBrandingRequestForm: b
     const operationResult = await updateTemplate({
       id,
       formConfig,
+      allowPublishedUpdate: true,
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/components/branding/Branding.tsx]: Thoughts on making a more granular action for updateBranding  where we get the current val from the db and take in a param for just the branding --- merge that into the exiting template.

Thinking something like that would offer more protection as to only updating part of a published template

app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/components/branding/Branding.tsx
```diff
@@ -108,7 +106,7 @@ export const Branding = ({ hasBrandingRequestForm }: { hasBrandingRequestForm: b
         <Label htmlFor="branding-select">{t("branding.select")}</Label>
         <BrandingSelect
           className="mb-5 mt-2 max-w-[450px] truncate bg-gray-soft p-1 pr-10"
-          disabled={isPublished as boolean}
+          disabled={false}
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/components/branding/Branding.tsx]: We can remove this `attr`

lib/templates.ts
```diff
@@ -1562,3 +1563,86 @@ export const checkIfClosed = async (formId: string) => {
...
+      e.user.id,
+      { type: "Form", id: formID },
+      "AccessDenied",
+      "Attempted to update security attribute"
+    );
+    throw e;
+  });
+
+  // Ensure it is a pure JSON value
+  const sanitized = JSON.parse(JSON.stringify(jsonConfig)) as Prisma.JsonObject;
```
craigzour: As mentioned in Slack we should probably validate the JSON object with the template schema.

Here is an existing function that does it https://github.com/cds-snc/platform-forms-client/blob/a9de7c5e68a51d826d6d36c66499ef2d223c3cdb/lib/templates.ts#L503

lib/templates.ts
```diff
@@ -1562,3 +1562,111 @@ export const checkIfClosed = async (formId: string) => {
...
+  return jsonConfig;
+};
+
+/**
+ * WARNING:
+ * Avoid using this function for any update that would modify the structure of the form
+ * e.g. groups, layouts, elements, etc.
+ * Doing so would cause an error in the infra pipeline when processing submissions.
+ */
+export const updateFormJSONConfig = async (formID: string, jsonConfig: FormProperties) => {
```
timarney:
```suggestion
export const updateFormJSONConfig = async (formId: string, jsonConfig: FormProperties) => {
```

PR #6276, opened 2025-10-29
chore: update content for accuracy

anikbrazeau:
## Description

Previously we said "branding" was locked upon publishing. 

Fix this on the publishing page for accuracy.


thiessenp-cds (approved):
LGTM :)

PR #6288, opened 2025-10-30
chore: Small fix for wrong text in FR from Settings update

anikbrazeau:
Notifications found in new section in Settings:
- Was updated in English but not in French

Gestion de formulaire --> `Options liées aux réponses`


timarney:
@anikbrazeau 

You might want to start a fresh PR from main --- looks like this one ends up including other commits 

i.e. deleting the heritage logo
Screenshot 2025-10-31 at 7 53 59 AM

PR #6414, opened 2025-11-20
fix: show all user feature flags

thiessenp-cds:
# Summary | Résumé

This PR updates the Admin page - User Feature Flag section to show all users feature flags (system wide). Previously the user feature flags that were not pre-loaded into Redis. This meant that only accessed feature flags would be in the cache and not the complete list.


app/(gcforms)/[locale]/(app administration)/admin/(with nav)/flags/page.tsx
```diff
@@ -23,6 +25,10 @@ export async function generateMetadata(props: {
 export default AuthenticatedPage([authorization.canAccessFlags], async () => {
   const { t } = await serverTranslation("admin-flags");
 
+  // Get and sync user feature flags for all users
+  const usersWithFeatures = await getAllUsersWithFeatures();
+  await syncUserFeatureFlagsToRedis(usersWithFeatures);
```
Comment (dsamojlenko) [inline app/(gcforms)/[locale]/(app administration)/admin/(with nav)/flags/page.tsx]: why sync to redis here?

lib/userFeatureFlags.ts
```diff
@@ -110,3 +111,39 @@ export const featureFlagAllowedForUser = async (
...
+    const users = await getUsers({ id: { in: userIds } });
+
+    // Map user details back to the features
+    return usersWithFlags.map((uf) => {
+      const user = users.find((u) => u.id === uf.userId);
+      return {
+        ...uf,
+        userText: user ? `${user.name} (${user.email})` : uf.userId,
+      };
+    });
```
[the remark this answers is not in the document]
thiessenp-cds: I originally opted to reuse getUsers() but on second thought I prefer the more efficient query. Will do

app/(gcforms)/[locale]/(app administration)/admin/(with nav)/flags/page.tsx
Comment (dsamojlenko) [reply app/(gcforms)/[locale]/(app administration)/admin/(with nav)/flags/page.tsx]: Redis doesn't need a complete list of user features, user permissions get synced to Redis when they login, as needed.

Comment (thiessenp-cds) [reply app/(gcforms)/[locale]/(app administration)/admin/(with nav)/flags/page.tsx]: Right but to flip the question, is there there any reason not to populate the Redis case with all user features?

I'm not tied to this though. I see it as more of a nice to have.

Comment (thiessenp-cds) [reply app/(gcforms)/[locale]/(app administration)/admin/(with nav)/flags/page.tsx]: Actually just to get this done, I'll remove that part of the code. Easy to add it back later if we decide to

PR #6495, opened 2025-12-03
chore: Responses pilot - disable pause when navigating 

timarney:
# Summary | Résumé

Adds code to disable pause / pausing button

Removes conditional show hide in favour of disable the button

Screenshot 2025-12-03 at 1 33 09 PM


timarney:
Test
https://github.com/user-attachments/assets/f905487a-0b4e-4436-8bd3-c5f723ec655b

PR #6534, opened 2025-12-09
feat: add notification and AWS utils to connector package

thiessenp-cds:
# Summary | Résumé

Add notification and AWS utils to the connector package for reuse in both the app and infra code.

Note: updating the app code to use the utils will be done in a separate PR.

### DIFF
--- a/packages/connectors/src/index.ts
+++ b/packages/connectors/src/index.ts
@@ -1,4 +1,3 @@
 export { GCNotifyConnector, type Personalisation } from "./gc-notify-connector";
 export { PostgresConnector } from "./postgres-connector";
 export { notification } from "./notification";
-export { getAwsSecret, getAwsSQSQueueURL } from "./utils";


packages/connectors/src/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+import { DynamoDBDocumentClient } from "@aws-sdk/lib-dynamodb";
+import { LambdaClient } from "@aws-sdk/client-lambda";
+import { SQSClient, GetQueueUrlCommand } from "@aws-sdk/client-sqs";
+import { S3Client } from "@aws-sdk/client-s3";
+
+const globalConfig = {
+  region: process.env.AWS_REGION ?? "ca-central-1",
+};
+
+export const cognitoIdentityProviderClient = new CognitoIdentityProviderClient({
```
thiessenp-cds: This is currently only in the app but seemed like a good location to put it.

packages/connectors/src/notification.ts
```diff
@@ -0,0 +1,115 @@
...
+  } catch (error) {
+    throw new Error(`Could not create record: ${JSON.stringify(error)}`);
+  }
+};
+
+const _enqueueDeferredNotification = async (notificationId: string): Promise<void> => {
+  try {
+    const queueUrl = await getAwsSQSQueueURL("NOTIFICATION_QUEUE_URL", "notification_queue");
+    if (!queueUrl) {
+      throw new Error("Notification Queue not connected");
```
timarney: We should add the `cause` here in case we need the stack

i.e. 

throw new Error("Connecting to database failed.", { cause: err });

packages/connectors/src/notification.ts
```diff
@@ -0,0 +1,115 @@
...
+  emails: string[];
+  subject: string;
+  body: string;
+}): Promise<void> => {
+  const id = notificationId ?? randomUUID();
+  try {
+    await _createNotificationRecord({ notificationId: id, emails, subject, body });
+    await _enqueueDeferredNotification(id!);
+  } catch (error) {
+    throw `Error creating immediate notification id ${id} - ${(error as Error).message}`;
```
timarney: We should add the `cause` here in case we need the stack

i.e. 

throw new Error("Connecting to database failed.", { cause: err });

Also looks like your throwing a string?

Issue #3965, opened 2024-07-08
Remove /api/templates endpoint


bryan-robitaille:

timarney:
~Our cypress tests.~

https://github.com/cds-snc/platform-forms-client/pull/6545

ShadeWyrm:
Generate Responses Util needs a think through to identify a potential route.

Previous itterations scraped the app, another option is to simply request the form JSON, but it would increase the overhead for the tool.

anikbrazeau:
@ShadeWyrm is this issue still relevant or does the PR linked above cover it off?

ShadeWyrm:
It doesn't, generate_responses still needs to be remade.

anikbrazeau:
Only used to generate form test responses script, pre-server actions (before endpoints)

PR #6553, opened 2025-12-11
chore: Remove api/templates from response generator.

ShadeWyrm:
Makes the util script no longer use the \api\templates endpoint.

Part of #3965


timarney (approved):
Approving so you can test on staging

PR #6581, opened 2025-12-16
fix: Visibility infinite recursion prevention

dsamojlenko:
# Summary | Résumé

Fixes an infinite recursion bug in the element visibility checker.

To recreate the issue, the attached form file has two elements with conditionalRules that reference each other. With the old code, this results in an infinite loop recursion error.

Under the new code, we bail out of the loop and default the visibility of both elements to invisible. This is the safest approach. It does not prevent the user from creating a recursion error in their form, but prevents our system from crashing out and erroring.

Added test coverage for this situation as well.

[test-2025-12-16.json](https://github.com/user-attachments/files/24196253/test-2025-12-16.json)

Additionally, a fix to prevent the creation of this circular logic by preventing the user from selecting a target for a rule that already has a rule pointing back at it.

Screen Recording 2025-12-16 at 12 19 56 PM


timarney (dismissed):

timarney (approved):
QA'd the scenarios I can think of -- should be further QA'd on staging before release.

PR #6616, opened 2025-12-23
chore: update to Node 24

timarney:
# Summary | Résumé

Update to Node 24

### DIFF
--- a/package.json
+++ b/package.json
@@ -166,7 +166,7 @@
     "@types/lodash.merge": "^4",
     "@types/lodash.set": "^4.3.9",
     "@types/lodash.unset": "^4.5.9",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -166,7 +166,7 @@
     "@types/lodash.merge": "^4",
     "@types/lodash.set": "^4.3.9",
     "@types/lodash.unset": "^4.5.9",
-    "@types/node": "^24.7.2",
+    "@types/node": "^25.0.3",
```
bryan-robitaille: Just checking to see if this should remain 24 instead of 25?

bryan-robitaille (approved):

PR #6656, opened 2026-01-07
chore: update conditional indicator to use title (choice) vs question # 

timarney:
# Summary | Résumé

Updates conditional indicator to show question + choice id and not a question #

Screenshot 2026-01-07 at 2 53 36 PM


timarney:
@anikbrazeau 

Pushed a small update to only show the tip if there's text

Issue #6623, opened 2025-12-29
"Confirmation page" shows "Page title" tag or button


anikbrazeau:
## Description 

Unexpected rectangle with the text "Page title" appears on the "Confirmation page" when in "Edit" mode for a draft form:

<img width="828" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN0o4REE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--56808bbdf6dd43b1395050e87d776e39379e982a/Screenshot%202025-12-29%20at%203.18.31%E2%80%AFPM.png" alt="Screenshot 2025-12-29 at 3.18.31 PM.png" />

PR #6671, opened 2026-01-12
fix: confirmation page title

timarney:
# Summary | Résumé

Fixes: #6623


timarney:
@anikbrazeau here are the strings if you want to adjust them

PR #6702, opened 2026-01-14
fix: allow default or custom messages for all element types in error summary

thiessenp-cds:
# Summary | Résumé

This PR applies a previous fix #6682 to allow default or custom messages but now for all element types in the forms-form error summary.

From my testing the impacted form element types that will have a custom error message are:
- Short answer (with max length)
- Long answer (with max length)
- File Input (file over 10MB)
- Attestation

Before
Screenshot 2026-01-14 at 2 56 02 PM

After
Screenshot 2026-01-14 at 2 57 36 PM

### DIFF
--- a/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/components/notifications/Notifications.tsx
+++ b/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/manage/components/notifications/Notifications.tsx
@@ -1,11 +1,11 @@
 "use client";
 
-import { useCallback, useEffect, useState } from "react";
+import React from "react";
 import { useTranslation } from "@i18n/client";
 import { toast } from "@formBuilder/components/shared/Toast";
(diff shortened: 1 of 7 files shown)


thiessenp-cds:
@anikbrazeau I updated the Text input (short answer) to match the Long answer input behavior. Another option would be to update both inputs to only show the message in red text when reaching the max length.

Currently the code does::

**No error message under the max**
Screenshot 2026-01-15 at 11 21 11 AM

**Error message in red approaching max**
Screenshot 2026-01-15 at 11 18 01 AM

Screenshot 2026-01-15 at 11 18 10 AM

**Error message once gone past max**
Screenshot 2026-01-15 at 11 18 17 AM

thiessenp-cds:
@anikbrazeau For singular would "caractère" work for French - so "Il vous reste 1 caractère"? (same for the 0 char message)

Also looks like I found a bug in the French version of the Long answer missing the FR text (not to self)
Screenshot 2026-01-15 at 2 47 38 PM

PR #6707, opened 2026-01-15
chore: update hide modified on submit progress screen

timarney:
# Summary | Résumé

- Updates date modified to hide itself when a submission is in progress 
- Moves the visibility logic to within itself for confirmation page + is past closing date


anikbrazeau:
Tried to test and got stuck on the "submit" action, and didn't get a review page? 

Screenshot 2026-01-16 at 2 34 07 PM

timarney:
@anikbrazeau 

I retested a few times and not seeing the issue you bumped up against.

I also tried by downloading a copy of the form in your screenshot.

My first attempt - didn't submit at all

I tried again with different inputs and it submitted fine.  

Wondering if there's an issue happening with that specific form?  Maybe try that form on staging ? Thinking it's not related to this PR.

dsamojlenko (approved):

PR #6824, opened 2026-03-05
feat: notification v3

thiessenp-cds:
# Summary | Résumé

Adds the new notification pipeline to GC Forms. The big advantage is allowing emails to be sent deferred once submission processing is complete e.g. a submission with a file attachment. The file attachment case previously caused a confirmation email to be sent while the submission was often still being processed and not available - leading to confusion. The deferred option solves that. An immediate notification can also be called.

One of the big "infra" related changes is calling the GC Notify API on the infra side vs. in the App. This will add more opportunities for monitoring and tracking down issues like timeout errors.

*This PR also updates all email sending to use the new notification pipeline by updating sendEmail to use it by default. One exception is the 2FA uses the option to force sending email direct to GC Notify.* Once the notify pipeline is shown to be very stable 2FA can use it as well.

The feature is feature flagged. ON will send emails through the Notification pipeline. OFFwill send direct to GC Notify.

Note: This PR updates the Notify util function to more easily handle sending file attachments. The file size (field size) limit in DynamoDB (according to the docs) is 400 KB. A very large template with a size of about 300 KB was tested and worked well.

Note: This PR updates Notifications to use the new push notification found in @gcforms/connector. See the related infra PR for more info #1137

Note: **Requires https://github.com/cds-snc/forms-terraform/pull/1137**

### Test Legacy

Do the same as below (Test Notification) but ensure the Notification flag is OFF. You should see no CloudWatch log messages about notifications.

### Test Notification

Note: Remember to turn the Notification feature flag ON

#### Sending a Immediate Notification (+file attachment)

1. Open CloudWatch and tail the Notification and Reliability lambdas
2. Open a form in form-builder and go to Settings > Share > Share form file (enter details and send)
3. You should receive a relate email with a file attachment and see a log like the below

Screenshot 2026-07-03 at 8 57 05 AM

#### Sending a Deferred Notification 

1. Open CloudWatch and tail the Notification and Reliability lambdas
2. Open a form in form-builder and go to Settings > Response Options > Receive notifications and toggle your user on (optionally add a few more users with some having the feature on and others off)
3. Preview the form and send a submission
4. You should receive a related email (depending on whether the 2-email notifications threshold has been reached - can reset in Redis) and see a log like the below

Screenshot 2026-07-03 at 9 02 05 AM


packages/connectors/src/notification.ts
```diff
@@ -91,7 +93,7 @@ const _createRecord = async ({
   body: string;
 }): Promise<void> => {
   try {
-    const ttl = Math.floor(Date.now() / 1000) + 86400; // 24 hours from now
+    const ttl = Math.floor(Date.now() / 1000) + 3600; // 1 hour from now
```
thiessenp-cds: I set this originally to 24h for the case of a file attachment taking a very long time to process. Are you confident with the 1 hour?

thiessenp-cds: Makes sense based on infra refactoring.

craigzour (approved):

PR #6843, opened 2026-03-13
feature: Locked draft editing - feature flagged.

timarney:
# Summary | Résumé

This PR is an initial discovery for lightweight locked editing using simple server events and a heartbeat mechanism.

It has been designed to avoid infrastructure updates, preferring a lightweight coordination approach built on the existing application and transient lock state.

- No database updates
- Doesn't using Web Sockets or anything that could be potentially blocked or exploited.

👉  I have intentionally set it up to only perform lock checks for draft forms with 2+ users.  Later this could be adjusted to allow editing of published forms.

👉  **_This should allow us to merge as we see fit as only admins can add other users to a draft_.**

https://github.com/user-attachments/assets/99823b40-2baf-4659-857e-03a311966c76

<hr>

# Edit Lock Flow Summary

## SSE, locking, and heartbeat

The app keeps track of who currently has editing control of a draft form and shares that state across open sessions.

When the app loads an editable draft route, the client initializes `useEditLock` automatically and the lock flow starts for that form and browser session.

Locking checks happen in a few places:
- when an editable draft route first loads
- while the current editor session sends heartbeat updates
- while other sessions poll or receive SSE updates about lock changes
- when save or other protected update actions run on the server

Note:
- The server checks the lock again on protected updates, so the UI is not the only thing preventing conflicting edits.

If the lock is acquired:
- The client becomes the lock owner.
- A heartbeat loop starts.
- The heartbeat sends periodic `heartbeat` actions to keep the lock alive and extend its expiry.

If the lock is owned by someone else:
- The client marks the form as _read-only_.
- The banner is shown.
- The client polls lock status until the lock is released or taken over.

SSE is used to reduce delay between editors.
- The client opens an `EventSource` connection to `/api/templates/[formID]/edit-lock/events`.
- The server pushes `lock-status` events whenever the lock state changes.
- Other sessions can react quickly instead of waiting for the next poll interval.

https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events

⚠️ The heartbeat is not an activity signal.
- It only means the owner session still holds the lock.
- It mainly prevents stale locks from remaining forever.

For this PR the lock has only been setup for:
- Draft forms only.
- Forms with enough assigned users to require locking - avoids lock checking when only 1 user is assigned to the form.

## What happens on takeover

When a user clicks Take over editing:
- The client sends a `takeover` action to the edit-lock API.
- The server replaces the existing lock owner with the requesting user/session.
- A lock update event is published so other sessions are notified.

After the takeover succeeds:
- The client refreshes the latest saved form data from the server.
- Only after that refresh completes does the UI switch local ownership state.
- This avoids dropping the lock banner before the latest saved server version is loaded.

Current page-specific behavior:
- Preview: the refreshed server-backed store state triggers a preview rerender so the latest saved template is shown.
- Settings: after takeover, the page does a full reload so server-rendered settings values come back cleanly.

User-facing feedback during takeover:
- The banner shows a syncing message while the latest saved form is being loaded.
- A success toast appears after the newest saved form has been synced.

See:
- `lib/editLocks.ts`
- `lib/hooks/form-builder/useEditLock.tsx`
- `app/(gcforms)/[locale]/(form administration)/form-builder/[id]/components/EditLockBanner.tsx`
- `app/api/templates/[formID]/edit-lock/events/route.ts`


lib/editLocks.ts
```diff
@@ -0,0 +1,725 @@
...
+
+const getEditLockTakeoverSaveAcks = (): Map<string, number> => {
+  const globalWithStore = globalThis as EditLockGlobal;
+  if (!globalWithStore.__editLockTakeoverSaveAcks) {
+    globalWithStore.__editLockTakeoverSaveAcks = new Map();
+  }
+  return globalWithStore.__editLockTakeoverSaveAcks;
+};
+
+const redisEnabled = () => Boolean(process.env.REDIS_URL);
```
bryan-robitaille: Is there an environment / use case where we don't have Redis available?

bryan-robitaille (commented):
Amazing work!
Did a quick passthrough.  A few questions. 
Definitely a lot of context switching trying to wrap my head around this new functionality but I think I get it.
Side note for me was that the code was a little hard to follow because the functions were often chunked out into very small parts and jumping from block to block felt hard to follow when diving through the logic.

dsamojlenko (approved):

PR #6955, opened 2026-04-01
chore: shared Redis for draft edit

timarney:
# Summary | Résumé

This PR updates locked editing SSE to use shared `globalThis` state see instead of opening a Redis subscriber per request.

see: https://github.com/cds-snc/platform-forms-client/pull/6955/changes#diff-94b997dbaca86e71050a239f94dc8d0099297f47fde2f389b780129435d136baR34

The shared state tracks active streams by `userId:formId` so a new stream replaces any older stream for the same user and form.


dsamojlenko (approved):

timarney:
Will merge and we can iterate further as we go.

PR #7020, opened 2026-04-22
fix: update to use Redis streams and optimize a useEffect

thiessenp-cds:
# Summary | Résumé

The main changes in this PR:

1. Redis Streams instead of Pub/Sub 
Pub/Sub is meant for fire-and-forget that doesn't really do what we want. Instead we want a sequential append-only store that can be shared between 1 or more users. Redis stream fits this prefectly and was implemented.
Or in more detail:  `XADD` writes a unique entry with an incremental Id. `XREAD BLOCK` reads entries in guaranteed order and never misses one. Each template gets its own stream key (`edit-lock-stream:<templateId>`), and each SSE connection gets a dedicated Redis connection so the blocking read doesn't stall other commands. (_performance issue?_)

2. Deliver lock-free events to waiting users
The old SSE handler silently dropped locked false events. When user A released the lock, user B (who was polling and waiting) received the event and did nothing and they had to wait up to 60 seconds for their next poll tick. The fix makes both the SSE handler and the polling loop call `setTakeoverFallbackState() `immediately when the lock disappears, so the "Take over" button is shown right away.

3. Stop the acquire > release > acquire loop of madness!
This was the source of the "flashing" UI and the many unnecessary server requests. Every call to updateStore() wrote to Zustand > React re-rendered > a callback in the effect's dependency array etc.
The fix uses a new(?) "latest-ref" pattern. This groups all callbacks into a single `cbRef` object that is updated on every render, but because useRef is not reactive, the two main effects only re-run when enabled or status (e.g. authentication) changes and never because a callback was recreated.


timarney (approved):
Nice work @thiessenp-cds 

Tested locally --- seems to be working smooth and consistently --- let's merge and run a  "real" test on staging.

PR #7063, opened 2026-04-30
feat: Ability to Download Filtered Audit Logs / UX Improvement

### LINKED ISSUE #7011: Ability to request specific logs
### User Story:

As a GC Forms client,
I want to request specific logs/key actions,
So that I receive only the information that is valuable to me.

*Note: this is valuable from a performance perspective as well.

### Description:

With the current Key Actions log feature, clients can download a log of their last 100 items. This enhancement will allow them to specify which logs they would like to retrieve, so we're not pulling in unnecessary information.

- We've had client requests/needs -- can reach out to Jose for more insight.

ShadeWyrm:
Adds a dialog prior to clicking download audit logs that then provides the ability to select all events or a subset of events based on major categories.

Closes : #7011


timarney:
Found an issue while testing this ... 

Believe these would have been missed in the previous PR. 

Seeing the `placeholder` value for user id

Screenshot 2026-05-05 at 1 30 27 PM

To stay consistent I would suggest we use a toast message in place of this alert 

Screenshot 2026-05-05 at 1 29 05 PM

Let me know if you want to fix them in this PR or another one.

timarney:
Looks like there's a margin that should be removed on the buttons

Screenshot 2026-05-05 at 1 36 56 PM

timarney:
@samsadasivan @ShadeWyrm To keep consistent with the more dialog do we want to use the GCDS style checkboxes for this dialog?  


**Current checkboxes**

Screenshot 2026-05-05 at 1 42 06 PM

<hr>

**GCDS style checkboxes** --- more dialog

Screenshot 2026-05-05 at 1 38 47 PM

Screenshot 2026-05-05 at 1 41 02 PM

timarney (approved):

PR #7084, opened 2026-05-04
chore: added missing global-bundle.pem file output folder in @gcforms/database package

craigzour:
# Summary | Résumé

- Copies `global-bundle.pem` file to output folder (`dist`) in `@gcforms/database` package

Note: this is a requirement for when the package is used outside of the web app project (Lambda/API).


bryan-robitaille (approved):
Nice, I wasn't sure if it was or not.

PR #7190, opened 2026-05-20
chore: update right panel to use html native popover

timarney:
# Summary | Résumé

Updates right nav to use native html pop-over.

Demo current staging vs this PR

https://github.com/user-attachments/assets/c2bf5cb4-276a-4763-91e4-17f811b518f1


timarney:
closing for now will pick this back up later

PR #7193, opened 2026-05-20
fix: Fix auto flow start order

### LINKED ISSUE #7163: Creating a form without branching does not connect new pages added to each other
## Description

- Branching isn't automatically in a linear view (Start page -> Exit Page)

## Steps to recreate

1. Create a form
2. Add title description, etc. to start page
3. Add confirmation message to last page
4. Add page 1, page 2, page 3, page 4, page 5
5. Go to Preview, see how you go from Start to End without hitting the pages created in between
6. Go to Branching visualization, see how it does not match the expected linear flow

## Expected

Without any additional branching applied, for simple linear, one-path forms, clients expect pages to connect in the order within tree view. 

Once flow is reset to linear, additional pages added get inserted and moved, they are in the right / expected spot in the visualization and the preview.

timarney:
# Summary | Résumé

Fixes: https://github.com/cds-snc/platform-forms-client/issues/7163

Fixes the add-page auto-flow bug where `start` was not connecting properly after creating a new page.

The issue was that page insertion was using raw `form.groups` object order, while the actual tree order is tracked in `form.groupsLayout`.

### DIFF
--- a/lib/groups/__tests__/useGroupStore.test.tsx
+++ b/lib/groups/__tests__/useGroupStore.test.tsx
@@ -81,7 +81,11 @@ describe("useGroupStore addGroup", () => {
     });
 
     // Start should reconnect to the first page in the current layout order.
-    expect(result.current.template.groupsLayout).toEqual([secondPageId, firstPageId, expect.any(String)]);
+    expect(result.current.template.groupsLayout).toEqual([
+      secondPageId,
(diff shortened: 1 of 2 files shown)


anikbrazeau (approved):
Tested and LGTM!

dsamojlenko (approved):

Issue #7237, opened 2026-05-22
[Spike] Explore whether saving form progress can happen on the client-side


anikbrazeau:
## Context 

Better support and resilience for users in remote or rural communities with limited bandwidth, or intermittent connectivity. 

> Currently a user cannot save their answers while offline to continue the form later if the browser tab is closed. This is possible for forms that do not have file uploads. This feature does require an active internet connection.

https://gcdigital.slack.com/archives/C0117FN8G9K/p1779214894686169

## Description

Currently, if they don't close the tab the answers will remain. However to save the file onto the device means generating the saved data on the server side --- we can likely make that happen client side but would need to dig into it to see level of effort.

anikbrazeau:
This was completed in #7255

PR #7255, opened 2026-05-26
chore: Generate saved progress client side instead of on the server

dsamojlenko:
# Summary | Résumé

#7237 

Save/Resume and Save Response now generate the downloadable HTML on the client, and cleaned up some of the dialog wiring.

- Added a client-side HTML generator that uses React’s browser server renderer.
- Switched the confirmation dialog to use the new client generator, simplified close handler typing, and only close after a successful save.
- Switched mobile drawer save flow to client generation as well, with matching error handling/toast behavior.
- Simplified dialog mounting in two places (always mounted, controlled by open prop).
- Removed the now-unused server action file.


timarney (dismissed):
Tested locally --- with / without Internet connection.

🥇

timarney (approved):

PR #7270, opened 2026-05-29
chore: cache tree view

timarney:
# Summary | Résumé

This PR improves the responsiveness of the form builder tree view when working with very large templates.

- Share one cached tree-data snapshot across headless-tree item and children lookups during the same task, avoiding repeated full tree rebuilds while expanding sections.
- Use the shared data loader in the headless tree view.
- Optimize `groupsToTreeData` by indexing form elements with a `Map` instead of scanning the full elements array for each child item.

## Context

Loading a large template into the form builder made the tree view feel unresponsive. Opening a section could appear to do nothing for a noticeable amount of time before child items appeared.

The slow path was concentrated around the synchronous headless-tree data loader. Each item and children lookup called `getTreeData()`, which rebuilt the full tree. On large forms, a single expand action could trigger that work many times before the UI updated.


dsamojlenko (approved):

timarney:
@Abi-Nada I did some further testing ... going to merge so this can get tested on staging while folks are testing other PRs.

PR #7290, opened 2026-06-02
fix : Add missing Audit Log for Branding

### LINKED ISSUE #7288: Audit Log Events : Branding Changes
Missing Audit Log Entry:
> Branding change missing event/details

ShadeWyrm:
Closes #7288

### DIFF
--- a/lib/templates.ts
+++ b/lib/templates.ts
@@ -1798,7 +1798,7 @@ export const getFormJSONConfig = async (formId: string) => {
  * e.g. groups, layouts, elements, etc.
  * Doing so would cause an error in the infra pipeline when processing submissions.
  */
-export const updateFormJsonConfig = async (formId: string, jsonConfig: FormProperties) => {
+export const updateFormBranding = async (formId: string, jsonConfig: FormProperties) => {
   const { user } = await authorization.canEditForm(formId).catch((e) => {
     logEvent(
       e.user.id,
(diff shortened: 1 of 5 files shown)


lib/templates.ts
```diff
@@ -1860,7 +1860,14 @@ export const updateFormJsonConfig = async (formId: string, jsonConfig: FormPrope
...
 
-  logEvent(user.id, { type: "Form", id: formId }, "UpdateFormJsonConfig");
+  const brandName = jsonConfig.brand?.name ?? "N/A";
+  logEvent(
+    user.id,
+    { type: "Form", id: formId },
+    AuditLogEvent.UpdateFormBranding,
+    AuditLogDetails.UpdateFormBranding,
+    { brand: brandName }
+  );
```
craigzour: If we connect this audit log here, it means that no matter what we change in the form JSON config we are always going to produce a branding change audit log? Am I missing something?

ShadeWyrm: The function seems to only ever called for branding changes, so I think it may be worth renaming the functions involved? Not sure.

craigzour: Oh ok! Well I think we should definitely rename it to something that refers to branding update only.

Abi-Nada:
Tested and LGTM! 
Screenshot 2026-06-02 at 4 35 52 PM

Abi-Nada:
It looks like the naming for "Government of Canada" branding appears as "N&#x2F;A"
Screenshot 2026-06-03 at 9 11 22 AM

craigzour (approved):

PR #7294, opened 2026-06-03
chore: Preparatory work for Template Versioning

dsamojlenko:
# Summary | Résumé

This PR contains a bunch of ui changes related to Template Versioning.
Excludes changes to templates.ts
None of the code in this PR will execute as the Feature Flag is turned off.


timarney (approved):
Let's get this in so we can create more focussed PRs around the template queries.

PR #7299, opened 2026-06-04
chore: form generator 

thiessenp-cds:
# Summary | Résumé

Adds a script to generate forms for testing. See readme for brief instructions.

Note: this code is AI generated. AI because otherwise I wouldn't feel it was worth developing.

Quick Demo:

https://github.com/user-attachments/assets/c8149f61-5c79-4e1a-9529-e4f43ef02ae1


bryan-robitaille (commented):
A few things:
- The db seed script already creates templates in the db and can associate them to the [local dev](https://github.com/cds-snc/platform-forms-client/blob/c4c5002de272208d51a37412186133f70e1f4086/.env.yarn.example#L39)
- The script is using pure SQL and pg instead of prisma which can cause things to get out of sync as the product continues to evolve, especially around the templates model.
- We can add more fixtures to the dev db seed for different types of forms etc without needing an additional script.
- The comments are out of sync because .env doesn't exist anymore.  It's `.env.yarn`

I'm happy you didn't spend time writing this code before checking in if there were existing ways to accomplish what you're trying to do but this AI code isn't helpful long term.

thiessenp-cds:
Thanks for taking a look and the helpful comments! If I ever need to generate a bunch of templates for testing, I'll come back to this PR. More as a prototype to scrap and take ideas from than working code to build off.

PR #7300, opened 2026-06-04
chore: update share menu

timarney:
# Summary | Résumé

We've have some support tickets regarding the removal of the share json option for authenticated users.

This update adds a share menu dropdown when authenticated to allow folks to pick between share access and share file.

## Testing 

- test unauthenticated flows

- test  authenticated when no form name is filled --- should show as a button 

https://github.com/user-attachments/assets/0c7d921b-a56b-490d-a088-e2818acfac73

- test authenticated with form name filled

https://github.com/user-attachments/assets/699912b0-6582-4dec-8015-93bfce9c6b31


dsamojlenko (dismissed):

anikbrazeau (approved):
LGTM! :D

PR #7302, opened 2026-06-04
chore: Refactor Templates functions

dsamojlenko:
# Summary | Résumé

Split the large monolithic lib/templates.ts (~1800+ lines) into a modular structure organized by semantic responsibility. All external imports remain unchanged via a public barrel file, maintaining API compatibility (with a few exceptions outlined below). 

## Motivation
- Navigability: Single 1800+ line file is difficult to scan; individual operations are now in focused files.
- Feature isolation: Upcoming template versioning work touches many functions; modular structure reduces merge conflicts and isolates changes.

## Renamed functions/API changes

Old Name | New Name | Reasoning
-- | -- | --
checkUserHasTemplateOwnership | assertCanEditTemplate | Clarifies this is a guard/assertion, not a boolean query
checkIfClosed | getTemplateClosureState | Returns state object, not boolean; name reflects actual return type
assignUserToTemplate | addAssignedUserToTemplate | Pairs clearly with removeAssignedUserFromTemplate; explicit add/remove symmetry
onlyIncludePublicProperties | mapTemplateToPublicFormRecord | Explicit mapper naming; aligns with pure transform intent
getTemplateWithAssociatedUsers | getTemplateWithAssignedUsers | Matches domain language used elsewhere in codebase
updateAssignedUsersForTemplate | syncAssignedUsersForTemplate | Clarifies operation computes deltas and syncs, not a generic partial update

## File structure
```
lib/templates/
├── index.ts                          # Public API barrel
├── internal/
│   ├── index.ts                      # parseTemplate, checkFlag
│   ├── errors.ts                     # Custom error classes
│   └── notifications.ts              # Email template dispatch
├── mappers/
│   └── mapTemplateToPublicFormRecord.ts
├── mutations/                         # State-changing operations
│   ├── createTemplate.ts
│   ├── updateTemplate.ts
│   ├── updateIsPublishedForTemplate.ts
│   ├── ...
├── queries/                          # Read-only operations
│   ├── getPublicTemplateByID.ts
│   ├── getFullTemplateByID.ts
│   ├── ...
└── guards/
    └── assertCanEditTemplate.ts
```


lib/templates/guards/assertCanEditTemplate.ts
```diff
@@ -0,0 +1,6 @@
+import { authorization } from "@lib/privileges";
+
+// Remove and replace this utility with new authorization object in code
+export const assertCanEditTemplate = async (formID: string) => {
```
bryan-robitaille: Since this utility function is only used once in the code base and there is a note `Remove and replace this utility with new authorization object in code` could we just do the quick refactor in this PR?

lib/templates/internal/index.ts
```diff
@@ -0,0 +1,74 @@
+import { Prisma } from "@gcforms/database";
+import { authorization } from "@lib/privileges";
+import { checkOne } from "@lib/cache/flags";
+import { FormRecord, FormProperties, SecurityAttribute, ClosedDetails } from "@lib/types";
+import { NotificationsInterval } from "@gcforms/types";
+
+export const checkFlag = async (flag: string) => {
+  return (await Promise.all([checkOne(flag), authorization.canAccessBetaComponents(flag)])).reduce(
```
bryan-robitaille, line 14: While we have this open and are reorganizing I wonder if we can rename this `canAccessBetaComponents` to something like `checkUserFlag` as it is no longer just for access to beta components.
We could also flag as a todo to merge with the `checkOne` somewhere so the func is more global and can live outside of the Templates module

lib/templates/internal/notifications.ts
```diff
@@ -0,0 +1,81 @@
...
+import { sendEmail } from "@lib/integration/notifyConnector";
+
+/**
+ * Notify owners of ownership changes (owner removed)
+ *
+ * @param userToRemove User to be removed
+ * @param form Form properties object
+ * @param users Current owners
+ */
+export const notifyOwnersOwnerRemoved = async (
```
bryan-robitaille: Since we're refactoring names I wonder if this would be easier to read at `notifyOwnerRemoved`

lib/templates/internal/notifications.ts
```diff
@@ -0,0 +1,81 @@
...
+};
+
+/**
+ * Notify all owners when ownership changes (owner added)
+ *
+ * @param user New owner
+ * @param form Form properties object
+ * @param users Current owners
+ */
+export const notifyOwnersOwnerAdded = async (
```
bryan-robitaille: Since we're refactoring names I wonder if this would be easier to read at `notifyOwnerAddedd

lib/templates/mappers/mapTemplateToPublicFormRecord.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/*
+ * Extract only the public properties from a form record.
+ * The public properties are the ones that are needed to display the form
+ * to unauthenticated users. (e.g. when filling out a form)
+ * Also sets some of the default values for properties that are not set.
+ * @param template A Form Record, containing all the properties
+ * @returns a Public Form Record, with only the public properties
+ */
+export const mapTemplateToPublicFormRecord = (template: FormRecord): PublicFormRecord => {
```
bryan-robitaille: Semanticly wouldn't this be the same as `parseTemplate` found under internal/index

bryan-robitaille (commented):
There's a strong push to move away from large barrel files because of the performance impact it has and the complexity it can bring when writing test mocks.

Instead of having a single function per file could we not merge the `mutations` into a single module and the `queries` into a single module?

lib/templates/guards/assertCanEditTemplate.ts
dsamojlenko: We could. I was trying to avoid any functionality or api changes, but this seems reasonable.

Simple as this? https://github.com/cds-snc/platform-forms-client/pull/7302/changes/b9a66810b91087ef60519aa094c24e6c8a5a2e6e

lib/templates/mappers/mapTemplateToPublicFormRecord.ts
dsamojlenko: makes sense https://github.com/cds-snc/platform-forms-client/pull/7302/commits/f562b5d2ca19fd81bf44cffb86d5f55b3e460aff

timarney:
**Note:**
I think all the above kind of just comes down to user preference so I'm fine with wherever we land. 

> There's a strong push to move away from large barrel files because of the performance impact it has and the complexity it can bring when writing test mocks.

Believe the primary issue we've hit is mixing and matching server and client code but I think we could remove the root barrel file and prefer direct import.


> Instead of having a single function per file could we not merge the mutations into a single module and the queries into a single module?

Perhaps the the mutations dir and queries dir each get a index file?  That would keep the functions standalone --- allow import exporting as "singles" and or use the index if you need multiple.    I feel like if we put all the functions in one file we will end up back in the same kind of situation where the single file keeps growing and gets harder to figure out what's in there.  

<hr>

lib/templates/guards/assertCanEditTemplate.ts
bryan-robitaille: yup, that should work.

bryan-robitaille:
> Believe the primary issue we've hit is mixing and matching server and client code but I think we could remove the root barrel file and prefer direct import.

Originally yes, but I've also run into multiple issues when creating tests.  Some of our modules are self instantiating with Immediately Invoked Function Expressions (IIFE).  If one of those modules are a dependency of one of the functions in a barrel file, the module is instantiated even if you haven't imported that function into the module you're working on.  When building out mocks for tests I've found that we often need to mock lib modules that have nothing to do with the function we're building a test for.
There's also the issue that barrel files can impact tree shaking and bloat bundle size.  NextJS has a way around this but it's still very much a manual process of specifically identifying the modules: https://vercel.com/blog/how-we-optimized-package-imports-in-next-js

>  I feel like if we put all the functions in one file we will end up back in the same kind of situation where the single file keeps growing and gets harder to figure out what's in there.

I fully understand and support that if the files keep growing in size with additional functions that we'll get back to the same place we are now.  My question is: Do we really need all these specialized query functions?  Looking quickly at the mutations folder there is a  generalized `updateTemplate` file/function that seems to cover off a few of the use cases for other mutation files listed in the same folder.  Instead of recreating new functions for when we need to do a mutation should we not be reusing existing functions where we can?


I can definitely compromise on an index file in each of the query / mutation folders but I'd like to have the whole template module flagged as needing some love and refactoring from a tech debt perspective.

timarney:
> when we need to do a mutation should we not be reusing existing functions where we can?

100%

Maybe we settle on  index file in each of the query / mutation folders and make real effort in the next sprint to see what we can re-purpose / remove etc...  We'll want to do that work as we start to build out the versioned equivalents.

dsamojlenko:
The reason I used the barrel file was to try to maintain a consistent API and limit the blast radius of this PR. Changing to direct/leaf imports or sub-barrels means touching every import across a lot of files, which makes this change much harder to review safely.

If the preference is away from the barrel file, then I'd say we should just go straight to direct/leaf imports over sub-barrels. There is no reason to maintain sub module barrel files over a main index. You'd still have to update all imports.

My preference would be to merge this structural refactor first, then do a focused follow-up PR (or series of PRs) for import migration, keeping them scoped and easier to review/test.

Just a small note too that tests can (and arguably should) only be using direct leaf imports over any barrel file for your exact reasoning, but again I was trying to keep the blast radius small.

lib/templates/internal/index.ts
dsamojlenko: Maybe this could be done in a small follow up PR

bryan-robitaille (approved):
As discussed with @dsamojlenko we'll merge this in for now and then follow up in the next couple of days with more PRs to implement the changes we talked about in this PR Review.

lib/templates/internal/index.ts
dsamojlenko: https://github.com/cds-snc/platform-forms-client/pull/7338

PR #7337, opened 2026-06-10
chore: update credentials check to allow for both archived and non-archived forms

thiessenp-cds:
# Summary | Résumé

The new edit-lock route used for polling was occasionally getting 403 errors. The reason became more obvious when comparing the initial template DB lookup on page load vs. the polling that has an authorization check. The difference is that the authorization check  was filtering templates by archive status (`ttl` field), but it was implicitly defaulting to "non-archived only". When users viewed archived templates, polling would fail with 403 errors because the authorization couldn't find those templates.

The fix modifies the authentication (`_getSubject()`) to keep the existing behaviour (backwards compatible)  but adds an additional state to check for both archived and non-archived (vs. one or the other):
- **`undefined/false` (default)**: non-archived only — preserves existing behaviour
- **`true`**: archived only — for restore operations  
- **`null`**: **no filter** — checks both archived and non-archived

### DIFF
--- a/app/api/forms/edit-locks/route.ts
+++ b/app/api/forms/edit-locks/route.ts
@@ -62,6 +62,7 @@ export const POST = middleware([sessionExists()], async (_req: NextRequest, prop
     }
 
     // Verify user has access to all requested templates using single batch DB query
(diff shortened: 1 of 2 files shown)


thiessenp-cds:
Will try another approach fist.

PR #7339, opened 2026-06-10
feat: Enable Start Date for Client Side Audit Logs

ShadeWyrm:
See title.


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/actions.ts
```diff
@@ -116,15 +116,23 @@ export const getFormEvents = AuthenticatedAction(
...
-          formId: event.subject.split("#")[1],
-          userId: event.userId,
-          event: event.event,
-          timestamp: new Date(event.timestamp).toISOString(),
-          description: event.description,
-        };
-      });
+      return events
+        .map((event) => {
+          const dateOfEvent = new Date(event.timestamp);
```
Comment (bryan-robitaille) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/actions.ts:121]: I can approve as is for now seeing as this is a release blocker but ideally we would be sending the query with the timestamp key greater than 'June 15' so we don't run a query only to discard all the responses.

bryan-robitaille (approved):

PR #7355, opened 2026-06-11
chore: update app to use new lastEditedBy field

thiessenp-cds:
# Summary | Résumé

Updates the client code to use the new lastEditedBy field.

lastEditedBy includes only the Name and NOT the email for security reasons.

**Actions updated: create, update, publish, archive**

My Page Draft and Published Cards now show the Name of the last person who did an update if available, otherwise nothing.
_Note: legacy forms will not have a lastEditedBy value set so these will be null until someone does an edit._

Additional Misc changes:
- add new form button back to recently edited
- updated a published card to have recently edited info


app/(gcforms)/[locale]/(form administration)/forms/components/server/Card.tsx
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(form administration)/forms/components/server/Card.tsx:1]: Moved to client

timarney (approved):

timarney:
Tested locally and in prev env with Pete

PR #7356, opened 2026-06-11
chore: Unbundle Log Action for getAllTemplatesForUser

ShadeWyrm:
Ensures a subject ID is available for the log actions that occur.

### DIFF
--- a/lib/templates/queries/getAllTemplatesForUser.ts
+++ b/lib/templates/queries/getAllTemplatesForUser.ts
@@ -77,9 +77,9 @@ export async function getAllTemplatesForUser(
           ability.user.id,
           { type: "Form", id: template.id },
           AuditLogEvent.ReadForm,
-          AuditLogDetails.AccessedForms,
+          AuditLogDetails.AccessedForm,
           {
-            formList: template.id,
+            formId: template.id,
           }
         );
       }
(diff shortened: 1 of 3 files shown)


lib/templates/queries/getAllTemplatesForUser.ts
```diff
@@ -71,10 +71,19 @@ export async function getAllTemplatesForUser(
...
-      logEvent(ability.user.id, { type: "Form" }, "ReadForm", AuditLogDetails.AccessedForms, {
-        formList: templates.map((template) => template.id).toString(),
-      });
+    if (templates.length > 0) {
+      for (const template of templates) {
+        logEvent(
+          ability.user.id,
+          { type: "Form", id: template.id },
+          AuditLogEvent.ReadForm,
+          AuditLogDetails.AccessedForms,
```
bryan-robitaille: Do we still need the form list?  Or could we swap it to something like:
`logEvent(ability.user.id, { type: "Form", id: template.id }, AuditLogEvent.ReadForm);`

ShadeWyrm: Swapped.

bryan-robitaille (dismissed):

bryan-robitaille (approved):

PR #7398, opened 2026-06-18
chore: Remove unused api/templates routes

dsamojlenko:
# Summary | Résumé

Re: #3965 

Removes unused /api/templates routes.


bryan-robitaille (approved):
Love it

PR #7431, opened 2026-06-24
fix: useEffect loop with GCFormsContext and Formik

bryan-robitaille:
# Summary | Résumé

This PR addresses an issue where the `useValueChanged` hook was creating multiple renders by modifying one of it's dependencies.

GCFormsContext holds `matchedIds` in state
Formik holds `values` in state

1st useEffect trigger after initialization:
- `values` change after user interaction with form element
- calls `updateValues` which runs logic at GCFormsContext level and updates matchedIds
- set properties on Formik `values` for ref and state values held at GCFormsContext level

2nd trigger
- `machedIds` and `updateValues` changed because `matchedIds` was recalculated and updated the state of GCFormsContext on previous trigger
- set properties on Formik `values` with new `matchedIds`

3rd trigger
- `machedIds` on Formik values has changed due to previous trigger
- call `updateValues` on GCFormsContext level but no state change needed to auto-triggering stops here
- same values are set again on Formik 'values' but 

**Solution**
- Change `matchedIds` from a state to a ref
- Remove the groups, matchedIds, and history from the Formik form state
- Fetch the latest values for matchedIDs, group, and history from the GCForms context at submission and validation
- Pass ref's or getter fn's to `Form` as props so they are available during submission and validation

**Before:**

https://github.com/user-attachments/assets/67b8651e-6377-4a6f-8225-7233e9fa5719

**After:**

https://github.com/user-attachments/assets/bd7b8ca3-ff94-4fbb-8ec3-ab04f4c79aae 

Ensure forms can still be submit and validated

There are still many performance enhancements that can be made in terms of state management but I'll keep those out of scope for now until we replace Formik.

# Pull Request Checklist


packages/core/src/process.ts
```diff
@@ -9,6 +9,13 @@ import {
   hasValue,
 } from "./validation/valueMatchesType";
 
+export type ResponseValidationValues = {
```
thiessenp-cds, line 12: You may want to add this to the export list in index.ts

thiessenp-cds (dismissed):
Code looks clean and after some quick testing of conditional logic, repeating set, and show-hide, the changes do not appear to break any existing functionality. LGTM :)

thiessenp-cds (approved):

PR #7483, opened 2026-07-10
bug: ensure empty values for keys that do not have values

### LINKED ISSUE #7482: form validation and show-hide appears to not always behave as expected
See [slack support thread](https://gcdigital.slack.com/archives/C03T4FAE9FV/p1783618374210319) for more info.

In some cases this form will miss validation. 
E.g. load answers and follow the flow to a page with an email input. Do not interact with the form inputs and try to hit continue. The validation picks up the email missing but misses the previous required inputs on the page.

In some cases the form also seems to fail to initially display the show-hide fields.
E.g. load answers and follow the flow until it appears you have filled in the required fields but are blocked. Navigate back and then continue. When the same screen loads it now shows the show-hide fields.

timarney:
# Summary | Résumé

fixes: https://github.com/cds-snc/platform-forms-client/issues/7482

Fixes issue where when a form is duplicated and answers are loaded from the previous form new fields with required values end up having the validation skipped --- unless the values have been interacted with.

This updates ensure when re-loading values we load load the initial values but also ensure "new fields" start as empty. 

The code will now ensure all keys have at least "" 

And keys that don't exist are removed.


timarney:
Testing 

Create a form a form with a few pages 

- Form 1
- Publish the form and save the confirmed answers

[saved.html](https://github.com/user-attachments/files/29898189/saved.html)

[form-1.json](https://github.com/user-attachments/files/29898198/form-1.json)


Duplicate form 1 --- load the saved answers.

[form-2.json](https://github.com/user-attachments/files/29898197/form-2.json)

Prior to this PR the required fields on the new page are skipped.

thiessenp-cds (approved):

PR #7492, opened 2026-07-14
chore: show+hide update

### LINKED ISSUE #7499: Show+Hide architecture update
## Context 

**Goal:** Make code/content more maintainable

**What's involved**

Update the show+hide architecture to be more performant and maintainable. The motivation for this work is from [this support ticket](https://app.zenhub.com/workspaces/gcforms-60cb8929764d71000e481cab/issues/gh/cds-snc/platform-forms-client/7485).

Some requirements ideally:

- event-driven approach
- separates concerns cleanly and makes the visibility logic testable in isolation
- visibility logic has unit tests or at least the newly added/touched code does

thiessenp-cds:
# Summary | Résumé

Closes #7499 

This PR updates show+hide to use an event-driven approach that is more performant and maintainable. It separates concerns cleanly and makes the visibility logic testable in isolation. Additionally the unit tests and playwright tests were updated to include cases for hiding elements and navigation.

**Summary of how the new implementation works:**
1. On form mount, dependency graph (dependent elements) is built showing which elements depend on which
2. Initial visibility is computed for all elements and cached in a Map
3. When form values change:
   - System detects which choice-based elements (radio/checkbox/select) changed
   - Filters to only those with dependents (using dependency graph)
   - Recomputes visibility only for affected elements
   - Fires custom event `form-values-changed` if any elements have changed
4. Components check the cached visibility map for what to show+hide

Additionally `visibilityMap` has been moved out of `GCFormsContext` into its own dedicated `VisibilityContext`. This scopes re-renders from visibility changes to only components that consume `useVisibilityContext`. Previously, any change to `visibilityMap` would trigger a re-render in every component subscribed to `useGCFormsContext`.

Note:
A follow up PR is needed to add (back?) accessibility semantics (aria-controls + aria-expanded) for show+hide. These semantics are just WCAG related they don't actually do much to help screen reader users. An announcement should also be used, or a similar strategy.


components/clientComponents/forms/ConditionalWrapper/ConditionalWrapper.tsx
```diff
@@ -26,19 +26,14 @@ export const ConditionalWrapper = ({
     currentGroup &&
     groups &&
     Object.keys(groups).length >= 1 &&
-    groups &&
```
timarney, line 29: We'll need to test this for older forms pre groups.

lib/hooks/useCustomEvent.tsx
```diff
@@ -26,8 +26,14 @@ export const EventKeys = {
   submitProgress: "submit-progress",
   openUnconfirmedApiKeyDialog: "open-unconfirmed-api-key-dialog",
   openCreateDraftConfirmDialog: "open-create-draft-confirm-dialog",
+  formValuesChanged: "form-values-changed",
 } as const;
 
+// Per-callback, per-event wrapper registry so the same callback can be
+// registered/removed for multiple event names without interference.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const listenerMap = new WeakMap<(detail: any) => void, Map<string, (event: Event) => void>>();
```
timarney, line 43: Let's try to fix the `any` here

lib/hooks/useGCFormContext.tsx
```diff
@@ -330,3 +417,12 @@ export const useGCFormsContext = () => {
   }
   return formsContext;
 };
+
+export const useVisibilityContext = () => {
+  const ctx = useContext(VisibilityContext);
+  // Outside the provider (e.g. tests, Storybook) treat every element as visible
```
timarney: Let's remove `Storybook` here -- we don't use it ... add other examples if they exist else let's just remove the comment

packages/core/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## [2.2.15] - 2026-07-14
```
timarney: I think this would be a major correct -- we're adding params ... what happens if you don't pass the params ... that would break correct?

timarney:
@thiessenp-cds nice work --- I added a couple of small comments --- 

Given this update has the potential to break any live form this will need a good amount of QA time - we're getting a little backlogged on that so we'll need to add this to the QA queue - maybe later in the week but likely more towards next week.

If there's anything you can add for test that would be great.

lib/hooks/useCustomEvent.tsx
thiessenp-cds: I left you a story in the comments. I agree it's not ideal and I'm open to other ideas

components/clientComponents/forms/Form/Form.tsx
```diff
@@ -1,5 +1,5 @@
 "use client";
```
thiessenp-cds, line 1: Before I could merge in changes I had to appease the React linter by not using sync code in a useEffect

timarney:
Noting https://github.com/cds-snc/platform-forms-client/pull/7587 should be merged when this gets merged.

lib/hooks/useVisibilityContext.tsx
```diff
@@ -0,0 +1,131 @@
+"use client";
```
thiessenp-cds, line 1: Separating this out from the GCFormContext gave little perf benefit - only a few less re-renders. So the value is just separation of concerns/readability.

thiessenp-cds:
We're going to try a simpler approach: https://github.com/cds-snc/platform-forms-client/pull/7781

This PR can be used as a future reference if we ever want to do some fancy element diffing :)

PR #7496, opened 2026-07-14
chore: update publish dialog

timarney:
# Summary | Résumé

Screenshot 2026-07-14 at 1 06 21 PM


thiessenp-cds (approved):
For future thought: I wonder if highlighting the bottom dialog text in some way would help users pay more attention to it.

PR #7519, opened 2026-07-16
chore: avoid deleting API key during template deletion when none exists

craigzour:
# Summary | Résumé

Context: while investigating on a different issue I noticed that, for a form created with no API key, we always had a `DeleteAPIKey` audit log created upon form deletion

- Adds condition to avoid deleting API key during template deletion when none exists


timarney (approved):
Nice  catch

PR #7521, opened 2026-07-16
chore: prevent edit published for forms using email delivery

timarney:
# Summary | Résumé

Adds client side checks to avoid email delivery forms from creating a new draft.

https://github.com/user-attachments/assets/975d3645-eee8-405a-a1bc-734fa45a26e4

I have avoided doing server checks here on purpose to avoid a wasteful db queries given email has been retired for a while now --- didn't want to litter the code with a bunch of checks for an edge case.


anikbrazeau (approved):
Tested in preview environment and looks good to me!

PR #7590, opened 2026-07-27
chore: added formPurpose migration process for legacy forms

### LINKED ISSUE #7569: Edge case: Editing an older published form
## Description

For forms that were published **_before_** we introduced "Intended Use" --> there's no way to republish them, given you cannot update these Settings post-publishing.

There are [1247 forms in Prod](https://docs.google.com/spreadsheets/d/14W-Y17dowNOvBgkyM9iir6tZ8a71scn9uSAsGPKDeIM/edit?gid=0#gid=0) that don't have an Intended Use selected.

## What's involved

Add `N/A` to old forms that don't have an intended purpose (administrative / non-administrative) to allow republishing of old forms (otherwise published forms would be blocked from future edits by not having "Settings" requirements)

## Steps to reproduce
1. Go to a published form created before we introduced "Intended Use"
2. Edit the published form
3. Try to re-publish it
4. See error -> you need to update Intended Use, but the option in Settings is greyed out.

## Details
- **Browser**: Chrome
- **Operating system**: Mac
- **Environment**: Staging
- **Language**: EN

## Expected behaviour

I'm not quite sure what the expected behaviour here should be. Ideally they could update that intended use section, so they can re-publish the form? Are there impacts of allowing that field to be updated post-publishing?

## Screenshots or videos

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMDZPRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--688c7ac65a514f79618957fabf886003d97d84f6/Screen%20Recording%202026-07-21%20at%201.12.18%E2%80%AFPM.mov)

craigzour:
# Summary | Résumé

closes #7569

- Adds migration code to make sure legacy forms get a `formPurpose` equal to `"n/a"` instead of `""`

Note: we will have to monitor deployment process in both Staging and Production to make sure migration was successful.

### DIFF
--- a/packages/database/CHANGELOG.md
+++ b/packages/database/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.1.6] - 2026-07-27
+
+- Added migration process for legacy forms with no set `formPurpose`
+
 ## [1.1.5] - 2026-07-22
 
 - Republish package following failed NPM publication
--- a/packages/database/package.json
+++ b/packages/database/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gcforms/database",
-  "version": "1.1.5",
+  "version": "1.1.6",
   "author": "Canadian Digital Service",
   "license": "MIT",
   "publishConfig": {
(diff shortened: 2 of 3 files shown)


thiessenp-cds (approved):
I did a local test with 4 published forms and 1 draft form. 2 of the published forms had a formPurpose and 2 other published forms had an empty formPurpose. Running the seed updated only the 2 published forms with an empty formPurpose. LGTM :)

PR #7591, opened 2026-07-27
feat: AddressComplete Server Side

ShadeWyrm:
Switches from Client Side to Server Side Implementation for AddressComplete


components/clientComponents/forms/AddressComplete/AddressComplete.test.tsx
timarney, line 83: Let's see if we can fix the any here

components/clientComponents/forms/AddressComplete/actions.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  const responseData = await response.json(); //Todo #4341  - Error Handling
+
+  return responseData.Items as AddressCompleteChoice[];
+};
+
+// Helper function combines API component results into single address object.
+export const getAddressComponents = async (
+  addressCompleteResult: AddressCompleteResult[],
+  language: string
```
timarney: Minor but let's use out language type here 

```
import { type Language } from "@lib/types/form-builder-types";
```

timarney (approved):

PR #7594, opened 2026-07-28
fix: make sure user is defined before accessing user attributes

craigzour:
# Summary | Résumé

Context: https://gcdigital.slack.com/archives/C05G766KW49/p1785227425200069

- Modifies if statement to make sure session's user is defined before accessing its properties


timarney (approved):
Nice

PR #7595, opened 2026-07-28
fix: Allow draft forms to be submitted in Preview

### LINKED ISSUE #7585: Using the "draft" preview does not allow "submit" on a versioned form to be republished
## Steps to replicate

1. Go to a published form
2. Create a new editable draft version of the published form
3. Make changes in the "Edit" section -- make sure form is multiples pages
4. Go to "Preview/Test"
5. Fill out the form all the way to the review page
6. Try to submit
7. Get blocked

## Details

- cmi3n6dqz0003x901npyyt74u
- Staging
- English
- Macbook
- Chrome

## Screen capture

https://github.com/user-attachments/assets/e61b7043-ce0d-4f7b-8134-03676258add3

### LINKED ISSUE #7597: Submissions on draft forms are not saved with the right form template version
## Description
Submissions on Draft forms (via Preview page) are not saved to Vault with the right form template version. Form must have two versions (v1 published and v2 in draft)

## Steps to reproduce
1. Create a form and publish it (version 1)
2. Edit previously published form (will create version 2) and add a new element to the form
3. Go to Preview page and submit a response (while being in draft mode)
4. Check database. "Draft" submission was saved with `Version` attribute value being `1` instead of `2`

## Expected behaviour
Submission should have had latest form template version assigned to it

dsamojlenko:
# Summary | Résumé

Fixes #7585 
When determining whether to run hCaptcha on a form, we're just checking if the app is not in either "development" or "test" mode, and if the template isPublished.

The problem is with Draft forms, the main Template record is published, and so hCaptcha runs, even if we're in the Preview screen.

This PR adds a specific check for if we're on the Preview screen to the hCaptcha check.

Also fixes #7597 
Similarly when in Preview on a form submission, we retrieve the template to run validation on the submission and to attach the version number to the submission record. The problem is, we were always retrieving the Published version if it exists (ignoring any drafts).

This PR adds a property to `getPublicTemplateByID` to distinguish between when we want to retrieve the published version, or draft. Most of the time we want published so that's the default, but in the submitForm action, we use the `isPreview` property added in the above fix to indicate that we want to retrieve the draft version.

### A note about Form.tsx
When committing these changes, I got several blocking eslint errors related to Form.tsx, but not related to code I had modified, so these are pre-existing errors:

```
✖ eslint --config eslint.config.mjs:

/Users/dave.samojlenko/Code/gcdigital/forms-client/components/clientComponents/forms/Form/Form.tsx
  108:9  error  Error: Calling setState synchronously within an effect can trigger cascading renders

Effects are intended to synchronize state between React and external systems such as manually updating the DOM, state management libraries, or other platform APIs. In general, the body of an effect should do one or both of the following:
* Update external systems with the latest state from React.
* Subscribe for updates from some external system, calling setState in a callback function when external state changes.

Calling setState synchronously within an effect body causes cascading renders that can hurt performance, and is not recommended. (https://react.dev/learn/you-might-not-need-an-effect).

/Code/forms-client/components/clientComponents/forms/Form/Form.tsx:108:9
  106 |     if (!props.isValid && !canFocusOnError) {
  107 |       if (props.submitCount > lastSubmitCount) {
> 108 |         setCanFocusOnError(true);
      |         ^^^^^^^^^^^^^^^^^^ Avoid calling setState() directly within an effect
  109 |         setLastSubmitCount(props.submitCount);
  110 |       }
  111 |     } else if (!props.isValid) {  react-hooks/set-state-in-effect
  120:7  error  Error: Calling setState synchronously within an effect can trigger cascading renders

Effects are intended to synchronize state between React and external systems such as manually updating the DOM, state management libraries, or other platform APIs. In general, the body of an effect should do one or both of the following:
* Update external systems with the latest state from React.
* Subscribe for updates from some external system, calling setState in a callback function when external state changes.

Calling setState synchronously within an effect body causes cascading renders that can hurt performance, and is not recommended. (https://react.dev/learn/you-might-not-need-an-effect).

/Code/forms-client/components/clientComponents/forms/Form/Form.tsx:120:7
  118 |   useEffect(() => {
  119 |     if (typeof window !== "undefined") {
> 120 |       setIsServer(false);
      |       ^^^^^^^^^^^ Avoid calling setState() directly within an effect
  121 |     }
  122 |   }, [setIsServer]);
  123 |
```


Abi-Nada:
Tested and LGTM!
- I inputted a Draft form response and was successfully able to submit ✅ 
- I was successfully able to download + sign-off the on the Draft response ✅

timarney (approved):

PR #7606, opened 2026-07-30
fix: Adds session check for edit-lock polling

thiessenp-cds:
# Summary | Résumé

Fixes an issue where a user could be signed out and continue edit-lock polling - eventually causing them to be blocked at the firewall level.

The fix is to add an auth check to cancel polling attempts if unauthorized.

See [Slack thread](https://gcdigital.slack.com/archives/C01GFD150MC/p1785411960811629) for more info.

### Test

1. Sign into forms-forms and open the developer tools network panel - note the requests to edit-lock
2. In a separate tab open forms-form and sign out the same user from step 1
3. Go back to the original tab with the open network panel. You should see two edit-lock polling requests fail with a 401 and then no more polling requests (or 1 request on prod)

### DIFF
--- a/app/(gcforms)/[locale]/(form administration)/forms/components/hooks/useEditLockPolling.ts
+++ b/app/(gcforms)/[locale]/(form administration)/forms/components/hooks/useEditLockPolling.ts
@@ -1,6 +1,5 @@
 import { useCallback, useEffect, useRef } from "react";
 import { FormsTemplateWithLockInfo, EditLocksResponse } from "../types";
-import { logMessage } from "@root/lib/logger";
 import { POLLING_TEMPLATE_CAP } from "../constants";
 
 const PROGRESSIVE_BACKOFF = {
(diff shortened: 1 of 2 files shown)


app/(gcforms)/[locale]/(form administration)/forms/components/hooks/useEditLockPolling.ts
```diff
@@ -202,6 +211,8 @@ export function useEditLockPolling({
     }
   }, []);
 
+  const scheduleNextPollRef = useRef<() => void>(() => undefined);
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(form administration)/forms/components/hooks/useEditLockPolling.ts:188]: These other updates are just to make the new React linting rules happy

timarney:
Tested on preview env...

Looks like it's working as expected

Screenshot 2026-07-30 at 9 43 11 AM

timarney (dismissed):

timarney (approved):

PR #7622, opened 2026-07-31
[experimental] feat: client logging

thiessenp-cds:
# Summary | Résumé

Adds the ability for browser-side code to send log messages to the server, where they are forwarded to CloudWatch via the existing pino logger.

Why?
Server logs give us no visibility into what happens inside the browser. This feature lets us capture client errors, warnings, and diagnostic info in the same CloudWatch log stream as the rest of the application, with a shared session ID for correlation.

### Test

```
// Direct singleton
import { clientLogger } from "@lib/client/clientLogger";
clientLogger.warn("Something unexpected", { formId: "abc" });

// Hook with default context
const { info, error } = useClientLogger({ formId: form.id });
info("Step completed", { step: 2 });
error("Submission failed");
```


thiessenp-cds:
Included other changes and looks messy - will recreate on a new branch

PR #7623, opened 2026-07-31
[experimental] feat: client logging

thiessenp-cds:
# Summary | Résumé

Adds the ability for browser-side code to send log messages to the server, where they are forwarded to CloudWatch via the existing pino logger.

**Why?**
Server logs give us no visibility into what happens inside the browser. This feature lets us capture client errors, warnings, and diagnostic info in the same CloudWatch log stream as the rest of the application, with a shared session ID for correlation.

**Features:**
- Buffered batching with multiple flush triggers (timer, buffer full, immediate on error)
- sendBeacon on page hide - ensures buffered logs aren't lost on tab close
- Per-session ID for CloudWatch log correlation
- Consecutive duplicate suppression with count

**Security:**
- Input validation and sanitization (length caps, control char stripping, timestamp skew check, allowlists)
- HMAC page token - requires a valid page load before logs are accepted
- Multi-layer rate limiting (firewall > Redis per-IP > batch size cap)
- Type-safe interface with primitive-only context values (prevents object injection)
- Timing-safe token comparison (prevents timing attacks on token validation)

**Any Other Options?**
An OpenTelemetry browser SDK implementation. This would be a replacement of the transport/format layer. It would offer a single connected timeline: browser click > API route > database query all correlated by one trace ID. Or simply put, more detailed tracing. But it would require and OTel-compatible backend (AWS Application Signals, Datadog, Honeycomb, other stuff?). So more infra work and of course then monitoring it.
It would not replace the HMAC token or Redis rate limiter or deduplication logic. So there would not really be any security gain.
IMO OpenTelemetry would be "nice" but not required since we just want to see error/info log messages. Plus all the infra work, little to no security gain, and added complexity, it doesn't seem worth it to me. 

**_TODO BEFORE RELEASING_**
Add `process.env.TOKEN_SECRET`. Actually probably move to a Github secret.

### Test

```
// Direct singleton
import { clientLogger } from "@lib/client/clientLogger";
clientLogger.warn("Something unexpected", { formId: "abc" });

// Hook with default context
const { info, error } = useClientLogger({ formId: form.id });
info("Step completed", { step: 2 });
error("Submission failed");
```


thiessenp-cds:
This was a fun experiment. Perhaps we can come back to it one day.

PR #7624, opened 2026-08-04
[WIP] feat: element plugin architecture

thiessenp-cds:
# Summary | Résumé

The motivation for this PR is to reduce the tedious task of adding a new form element. Adding a new element requires modifying 20+ files, not complex but it's a lot to remember and easy to make a mistakes. To solve this a plugin architecture replaces the old model.

Approach
- elements are registered in registry.ts
- element logic contained in plugin subfolder

Questions
- should there be separate "base elements" that are shared like input? Or is the duplication OK and keep each plugin self contained
- should there be two versions of elements during development to reduce conflicts / make it easier to copy changes over? (currently feature flagged)


thiessenp-cds:
Closing this for now. We can get back to this for use or reference when we get to the element plugin work.

Note: this work used simple elements as a test. In future work we may want to start with more complicated form elements to make sure the architecture works for all cases.

PR #7633, opened 2026-08-04
chore: pass reason through Error message for Notification module in Connectors package

craigzour:
# Summary | Résumé

Context: https://gcdigital.slack.com/archives/C05G766KW49/p1785858138713009

- Changed the way we pass the reason of a failure in the Notification module to make sure we have the root cause of a problem in our Cloudwatch logs


timarney (approved):

thiessenp-cds (approved):
LGTM

PR #7641, opened 2026-08-05
fix: form-filler sticky bar a11y update

thiessenp-cds:
# Summary | Résumé

Adds an aria role and label to the sticky bar to improve accessibility. This way it will show up in the landmarks list.
Also added a z-index to help make sure the sticky bar remains on top of other content - current and future.

### DIFF
--- a/i18n/translations/fr/form-builder.json
+++ b/i18n/translations/fr/form-builder.json
@@ -1714,5 +1714,6 @@
     "description": "Modifier un formulaire déjà en production",
     "button": "Mettre à jour"
   },
-  "formDownloadFailed": "Le téléchargement du formulaire a échoué."
+  "formDownloadFailed": "Le téléchargement du formulaire a échoué.",
+  "formActions": "[FR]Form actions"
 }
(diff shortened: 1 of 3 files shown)


i18n/translations/fr/form-builder.json
```diff
@@ -1714,5 +1714,6 @@
     "description": "Modifier un formulaire déjà en production",
     "button": "Mettre à jour"
   },
-  "formDownloadFailed": "Le téléchargement du formulaire a échoué."
+  "formDownloadFailed": "Le téléchargement du formulaire a échoué.",
+  "formActions": "[FR]Form actions"
```
thiessenp-cds: @anikbrazeau Quick translation when you have a minute. This is the label used to describe the "sticky" bar region on form filler.

anikbrazeau:
```suggestion
  "formActions": "Actions liées au formulaire"
 "
```

timarney (approved):

Issue #7466, opened 2026-07-07
Limit Address Complete form element to verifying Canadian addresses only -- revert to static fields for other countries [AddressComplete]


anikbrazeau:
## Context

The MOU specifies Canadian address lookups only. International lookups have a higher cost.
Canada-only will be the scope for version 1 and we will reassess the need for global lookups in the future.

## Description

Address component is designed to work for all addresses but Address Complete lookup only applies to Canadian addresses. 

## What's needed

**Alter our implementation to:**

- [x] Make sure the address component accepts all types of addresses with country selection coming first
- [x] Make Canadian addresses the default country selection
- [x] Set expectations with users when adding a form element that it's scoped to Canadian addresses only (content)
- [x] Limit lookup options to Canadian addresses only
- [x] Set up the Address Complete aspect to only validate for Canada
- [x] Re-examine the customizable aspect of Address in the "More menu"
- [ ] Allow clients to opt out of using AddressComplete
- [ ] Track the number of times Address component reverts back to static fields

timarney:
Does defaulting the checkbox to `Canadian addresses only` go far enough?

I'll start a PR to at least cover that piece first

anikbrazeau:
TBD, we'll connect with Ioana and Leanne on this next week, given there's cost implications, but that's a perfect first step, Canada by default for sure.

anikbrazeau:
Additional considerations re: cost / usage here for dicussion: https://docs.google.com/document/d/1QfoU9mfusgnVODghCDYNJw-8rUBKb6xw-LLSeyNy3Ks/edit?tab=t.0#heading=h.pgir47fbizxn

anikbrazeau:
@timarney we've made a scope/cost decision to only allow for Canadian address look ups, I've adjusted the card description above, if we're able to adjust the implementation accordingly. :)

timarney:
@dsamojlenko is going to pick up the "Canadian addresses only" updates as he's working on related pieces already.

YedidaZalik:
Updated content and content card for localization based on the content AC in this card

Abi-Nada:
FOLLOW-UP edge case: Error handling when it comes to rate-limiting:
- Make sure we don't show the error
- Hide the text around "Start typing"

anikbrazeau:
PR: https://github.com/cds-snc/platform-forms-client/pull/7647

**Remaining considerations:** 
 - Allow clients to opt out of using AddressComplete
[...]

PR #7647, opened 2026-08-05
chore: Address complete enabled only for Canadian addresses

dsamojlenko:
# Summary | Résumé

#7466 

Limit AddressComplete to Canadian addresses.

If the FormBuilder selects "Canadian Only":
- do not display a Country selector
- display an AddressComplete input
- show Province/Territory and Postal Code labels

If the FormBuilder selects "Allow International":
- display a Country selector that defaults to Canada
- If Canada is selected: display an AddressComplete input
- If any other country is selected: display a plain text input
- show Province/Territory/State and Postal or Zip labels

### Element Properties Defaults
Introduces a new pattern - a `defaults.ts` file in the AddressComplete component directory contains configuration properties for the component and their defaults. This then gets applied to the element on creation in `itemHelpert.ts`

This had the side effect of greatly simplifying the AddressCompleteOptions component in the MoreDialog.

This will be helpful for FormattedDate component as well and removes the need for hardcoding a default based on the non-existence of the property when a new element is created.

Arguably, we could also use this for the `choices` array for checkboxes/radios/etc as currently all elements get a choices array regardless if they need one because it's in the `defaultField` object in `itemHelper.ts`


timarney:
Did some testing seems to be working quite well.

timarney:
When reloading from a saved .html file the autocomplete doesn't show

**Form**
Screenshot 2026-08-07 at 10 05 12 AM

**Re-loaded Form**
Screenshot 2026-08-07 at 10 05 42 AM

**Re-loaded on current staging**
Screenshot 2026-08-07 at 10 09 07 AM

timarney (approved):

PR #7656, opened 2026-08-06
chore: address complete security

thiessenp-cds:
# Summary | Résumé

This PR focusses on securing the server action and more specifically the requests between the server action and Canada Post API. The main concern is blocking malicious users from spamming the CanadaPost API to keep costs down.

_Note: even with these measures a user can still spam the end point. The additional protection added just makes it much more difficult. (See testing section)_

**_TODO:_**
Add github secret for `ADDRESSCOMPLETE_TOKEN_SECRET`

## How the server action is protected:

1. Request parameters are validated and sanitized - user blocked if fails 
  a. Why? Help protect downstream Canada Post API from oversized or malformed inputs
2. HMAC validation - user blocked if fails  (see below for more info)
  a. Why? Reduce the number of invalid hits hitting the rate limiting logic
4. Rate limiting - user blocked if fails  (see below for more info)
  a. Why? Offer more fine grained rate limiting vs AWS rate limiting rules. This will gives us more control how we limit access to malicious users

Note: Security measures already in place
- Firewall Rate limiting: by client IP
- CSRF protection: automatic through NextJS

### HMAC Validation

FormId + HMAC validation on server action required before reaching the Rate limiter logic. The HMAC work was taken from previous work in https://github.com/cds-snc/platform-forms-client/pull/7623

How does Hash-based Message Authentication Code (HMAC) work?
1. The **page server component** generates a token using formId + the current hourly time window and embeds it in the HTML response as `<meta name="x-address-token">`. (note: token generated here and not server action)
6. `AddressComplete` reads the token from the meta tag on initial render and holds it as a constant for the component's lifetime.
7. `AddressComplete` includes the formId + token in the server action request.
8. The  **server action**  receives formId + token. The server action then recomputes `HMAC(TOKEN_SECRET, "address-complete:{formId}:{window}")` and rejects the request if it doesn't match either the current or previous hour's token. (hmac is part of the crypto lib)

### Rate Limiting

Rate limit all requests to Canada Post API to help reduce costs.

Uses Redis + client IP and a variety of limits set in constants. These values should be first discussed and potentially moved to an app config for easy tuning. Also we may want to add some sort of an allow list or way to disable this (flag?) if to many users are being blocked (then tune the limits and retry). 

Note: Because of shared IPs we need to be cautions not to let a few users spam from one IP and block all users on this IP. HMAC provides the initial gate to make this more difficult.


components/clientComponents/forms/AddressComplete/validation.ts
```diff
@@ -0,0 +1,17 @@
+// Canada Post IDs are alphanumeric with pipe separators e.g. "CA|CP|A|1234567"
```
thiessenp-cds, line 1: TODO: Double check these regex patterns are correct. See https://www.canadapost-postescanada.ca/ac/support/api/addresscomplete-interactive-find/

components/clientComponents/forms/AddressComplete/actions.ts
```diff
@@ -5,13 +5,49 @@ import { AddressCompleteChoice, AddressCompleteResult, AddressElements } from ".
...
 const autoCompleteUrl =
   "https://ws1.postescanada-canadapost.ca/AddressComplete/Interactive/Find/v2.10/json3.ws";
 const retriveAddressUrl =
   "https://ws1.postescanada-canadapost.ca/AddressComplete/Interactive/Retrieve/v2.11/json3.ws";
 const addressCompleteKey = process.env.ADDRESSCOMPLETE_API_KEY || "";
 
+// Note: 60 requests per minute across all three actions combined, since they all consume
+// Canada Post API quota. The requests will bypass Redis if it goes down to not break the form.
+const RATE_LIMIT_KEY_PREFIX = "address-complete:rate-limit";
+const RATE_LIMIT_MAX = 60;
```
thiessenp-cds, line 28: These amounts need more thought

thiessenp-cds:
This PR was broken up into separate PRs to make it easier to review

PR #7661, opened 2026-08-06
fix: add fallback solution for when an email fails to be sent through the Notification pipeline

craigzour:
# Summary | Résumé

Context: https://gcdigital.slack.com/archives/C05G766KW49/p1786033328562449

- Adds fallback solution for when an email fails to be sent through the Notification pipeline


thiessenp-cds (approved):
Works for both cases - LGTM :)

Large Form file shared (> 400 KB)
Image

Small Form file shared (< 400 KB)
Image

PR #7688, opened 2026-08-10
chore: address complete field sanitization

### LINKED ISSUE #7670: Prevent unauthenticated form builder lookups [AddressComplete]
## Context

## What's involved 

## Option analysis

## What's involved

thiessenp-cds:
# Summary | Résumé

This PR focusses on securing the server action and more specifically the requests between the server action and Canada Post API. The main goal is to protect ensure the Canada Post API only receives valid data. Any requests with invalid data should stop at the server action. This will act as the initial gate for securing the endpoint.

### DIFF
--- a/components/clientComponents/forms/AddressComplete/utils.ts
+++ b/components/clientComponents/forms/AddressComplete/utils.ts
@@ -7,6 +7,7 @@ import frResponses from "@i18n/translations/fr/form-builder-responses.json";
 import { countries } from "@lib/managedData/countries";
 import { Language } from "@lib/types/form-builder-types";
 import { AddressElements } from "./types";
+import { sanitizeField, truncateField } from "@gcforms/core";
 
 type AddressFieldKey = keyof AddressValidationError["fields"];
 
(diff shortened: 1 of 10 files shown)


components/clientComponents/forms/AddressComplete/utils.ts
```diff
@@ -187,3 +188,23 @@ export function getCountryCodeFromName(value?: string): string {
 
   return trimmed;
 }
+
+export const MAX_ADDRESS_FIELD_LENGTH = 200;
```
thiessenp-cds, line 192: The address length is used for the fields: city, street, province
Separate const lengths and functions could have been created but I think the address field is flexible enough in length and simplifies the checking into one util function.

timarney: yes that should be flexible enough

packages/core/src/helpers.ts
```diff
@@ -316,3 +316,12 @@ export const getElementsUsingChoiceIdParent = ({
 
   return elements;
 };
+
+export const sanitizeField = (value: string | unknown): string => {
```
timarney: Might want to rename 

`toStringOrEmpty`
`normalizeString`

This function sounds like it does replacements removes HTML or SQL risks etc...

packages/core/src/helpers.ts
```diff
@@ -316,3 +316,12 @@ export const getElementsUsingChoiceIdParent = ({
...
+  return String(value ?? "").trim();
+};
+
+export const truncateField = (value: string | unknown, maxLength: number): string => {
+  const str = String(value ?? "");
```
timarney: This could call the other function `toStringOrEmpty` or  `sanitizeField ` as it's currently called.

timarney:
Just noting if an error does return what I'm seeing is the street address ends up being the "full address" and the individual fields don't populate --- which in turn will show required errors when you try to submit.


Screenshot 2026-08-11 at 8 29 43 AM

or 

Screenshot 2026-08-11 at 8 31 09 AM


Not suggesting we need to adjust the behaviour ... under normal usage this shouldn't happen just posting for awarness


cc: @Abi-Nada @anikbrazeau @samsadasivan

components/clientComponents/forms/AddressComplete/validation.ts
```diff
@@ -0,0 +1,22 @@
...
+  return COUNTRY_CODE_PATTERN.test(code);
+};
+
+const ALLOWED_LANGUAGES = ["en", "fr"] as const;
+export const isValidLanguage = (lang: string): boolean => {
+  return (ALLOWED_LANGUAGES as readonly string[]).includes(lang);
+};
+
+// format used by Prisma-generated form IDs
+const FORM_ID_PATTERN = /^[a-z0-9]{20,40}$/;
```
timarney: Believe this is unused correct --- this remove from this PR

timarney (approved):

PR #7689, opened 2026-08-10
chore: address complete rate limiting 

### LINKED ISSUE #7670: Prevent unauthenticated form builder lookups [AddressComplete]

thiessenp-cds:
# Summary | Résumé

This PR focusses on securing the server action and more specifically the requests between the server action and Canada Post API. This is done by rate limiting requests between the client and server action.  We already have AWS firewall rate limiting but that threshold is way too high for this use case. The server action rate limiter adds a more fine grain rate limit and in a much smaller window.

The implementation uses Redis to temporarily store the IP info of the user. It is removed after the window expires e.g. 1 minute 

Note: because the Redis expire is called on every request, the window is sliding, not fixed. An IP that makes exactly rateLimitMax requests spread out slowly will never get blocked because the TTL keeps getting pushed forward. A static window could be used instead on creation if we feel it fit better with the shared IPs.

## Testing on localhost

1. Update the app setting `addressCompleteRateLimitMax` to something like 2
2. Run the app and open a form with an address complete element. Type really fast in the address input.
3. Look in your browser dev terminal network panel. Look for the related response and you should see

Screenshot 2026-08-10 at 7 04 28 PM

4. You can also check your Redis store and see that your IP is flagged for rate limiting. Note: `address-complete:rate-limit:::1` is expected. `::1` is the IPv6 loopback address, the equivalent of `127.0.0.1`. When you inspect the key you can see the current count.

Screenshot 2026-08-11 at 9 33 27 AM

5. After waiting about a minute your IP should no longer be rate limited


components/clientComponents/forms/AddressComplete/actions.ts
```diff
@@ -261,3 +275,33 @@ const nestedAddressPattern = /\s+-\s+\d+\s+(Addresses|Adresses)$/i;
 export async function matchesAddressPattern(input: string): Promise<boolean> {
   return nestedAddressPattern.test(input);
 }
+
+const RATE_LIMIT_KEY_PREFIX = "address-complete:rate-limit";
+const RATE_LIMIT_MAX = 200; // CanadaPost API quota may be 60? but 200 seems safer given shared IPs
```
thiessenp-cds: We may want to move this to an app configuration variable to give us more real time options for "tuning"

components/clientComponents/forms/AddressComplete/actions.ts
```diff
@@ -261,3 +276,36 @@ const nestedAddressPattern = /\s+-\s+\d+\s+(Addresses|Adresses)$/i;
...
+
+const RATE_LIMIT_KEY_PREFIX = "address-complete:rate-limit";
+
+// Returns true when the IP has exceeded the per-minute request "budget"
+const isRateLimited = async (ip: string): Promise<boolean> => {
+  try {
+    const rateLimitMax = Number(await getAppSetting("addressCompleteRateLimitMax"));
+    const rateLimitWindowSeconds = Number(
+      await getAppSetting("addressCompleteRateLimitWindowSeconds")
+    );
```
thiessenp-cds, line 329: The app settings are server cached so the DB lookup should only happen once.

craigzour: Caching happens in Redis so it will still require a Redis query every time this code runs. Not a big deal but still something to remember in case there is an opportunity to remove those settings in the future once we have finalize tweaking the limits

components/clientComponents/forms/AddressComplete/actions.ts
```diff
@@ -302,3 +317,36 @@ const nestedAddressPattern = /\s+-\s+\d+\s+(Addresses|Adresses)$/i;
 export async function matchesAddressPattern(input: string): Promise<boolean> {
   return nestedAddressPattern.test(input);
 }
+
+const RATE_LIMIT_KEY_PREFIX = "address-complete:rate-limit";
+
+// Returns true when the IP has exceeded the per-minute request "budget"
+const isRateLimited = async (ip: string): Promise<boolean> => {
```
craigzour: My only issue in this PR is the actual function name here because it hides the fact that it increments the counter and sets an expiry date as well. Though I am not sure how to solve this as I understand that is was done to simplify the code and optimize the number of Redis interactions 😅

thiessenp-cds: That was the intention and I agree. What about the very verbose option `incrementAndCheckRateLimiting` - does that seem clearer?

components/clientComponents/forms/AddressComplete/actions.ts
thiessenp-cds: Good to flag that. 
Another option I was thinking about was to get the vars at the page level and pass them to the action. But then I think they could be tampered with in the request.

craigzour: Yep, that is correct!

craigzour (approved):
LGTM! Nice work :)

thiessenp-cds:
@anikbrazeau @Abi-Nada I was thinking through and testing adding a `formId + IP` to the rate limiting key (how we identify a user). I realized that since the formId would be sent from the client to the server action, a malicious user could hit the server action directly and send random formId's - bypassing the rate limiting check with a random formId each time. We could check that the formId is valid but that is a bit "expensive" since it would require a DB lookup. 
The IP works because that is pulled in the server action itself. So I'd recommend not using the `formId` to make more fine grained rate limiting.

PR #7690, opened 2026-08-10
chore: address complete HMAC

### LINKED ISSUE #7670: Prevent unauthenticated form builder lookups [AddressComplete]

thiessenp-cds:
# Summary | Résumé

This PR focusses on securing the server action and more specifically the requests between the server action and Canada Post API. The main concern is blocking automation from spamming the CanadaPost API to help keep costs down.

In retrospect:
**_The HMAC adds code complexity and provides limited protection. It is probably not worth including._**

### HMAC Validation

FormId + HMAC validation on server action required before reaching the Rate limiter logic. The HMAC work was taken from previous work in https://github.com/cds-snc/platform-forms-client/pull/7623

How does Hash-based Message Authentication Code (HMAC) work?
1. The **page server component** generates a token using formId + the current hourly time window and embeds it in the HTML response as `<meta name="x-address-token">`. (note: token generated here and not server action)
6. `AddressComplete` reads the token from the meta tag on initial render and holds it as a constant for the component's lifetime.
7. `AddressComplete` includes the formId + token in the server action request.
8. The  **server action**  receives formId + token. The server action then recomputes `HMAC(TOKEN_SECRET, "address-complete:{formId}:{window}")` and rejects the request if it doesn't match either the current or previous hour's token. (hmac is part of the crypto lib)

TODO: Add github secret for `ADDRESSCOMPLETE_TOKEN_SECRET`


thiessenp-cds:
Closing this branch to clean up the open PRs list.

We can come back to this PR if ever there is interest in it.

PR #7702, opened 2026-08-12
chore: add owner checks

timarney:
# Summary | Résumé

Updates a few server actions to ensure owner checks

### DIFF
--- a/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/api-integration/components/throttlingRate/actions.ts
+++ b/app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/api-integration/components/throttlingRate/actions.ts
@@ -9,13 +9,17 @@ import {
 import { AuthenticatedAction } from "@lib/actions";
 import { ServerActionError } from "@lib/types/form-builder-types";
 import { AuditLogEvent, AuditLogDetails, logEvent } from "@lib/auditLogs";
+import { authorization } from "@lib/privileges";
 
 // Public facing functions - they can be used by anyone who finds the associated server action identifer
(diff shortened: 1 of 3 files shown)


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/api-integration/components/throttlingRate/actions.ts
```diff
@@ -9,13 +9,17 @@ import {
...
 import { ServerActionError } from "@lib/types/form-builder-types";
 import { AuditLogEvent, AuditLogDetails, logEvent } from "@lib/auditLogs";
+import { authorization } from "@lib/privileges";
 
 // Public facing functions - they can be used by anyone who finds the associated server action identifer
 
 export const getCurrentThrottlingRate = AuthenticatedAction(async (_, formId: string) => {
-  return getThrottling(formId).catch(() => {
+  try {
+    await authorization.canViewForm(formId);
```
Comment (craigzour) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/api-integration/components/throttlingRate/actions.ts]: To be consistent with the other `ThrottlingCache` related functions we could move this authorization check in https://github.com/cds-snc/platform-forms-client/blob/5b37489e613c5881fbb91cb2e15a3783166f13c9/lib/cache/throttlingCache.ts#L15

Comment (craigzour) [reply app/(gcforms)/[locale]/(form administration)/form-builder/[id]/settings/api-integration/components/throttlingRate/actions.ts]: Or we move all authorization checks from the file I linked to here. It makes more sense to me but also forces us to make sure that the authorization check should be added everywhere we call the function (something that could happen in the future)

craigzour (approved):

PR #7717, opened 2026-08-12
chore: address complete rate limiter followup

thiessenp-cds:
# Summary | Résumé

This PR mainly includes a change to add a new scope to the rate limiter. Other updates include code cleanup and adding a non-user identifiable log message when a user is rate limited.

The idea with Address Complete "scopes" is to give us more control over how we rate limit. The two scopes are: 
- `find` - rate limit with a separate key for find operations (app config value default 400 requests per minute)
- `retrieve` - rate limit with a separate key for retrieve operations (app config value default 200 requests per minute)

The log message is fired as a warning when a user in either scope is rate limited.

Also to help reduce "request noise" a minimum of two characters is now required before a request can be sent to Canada Post.

### DIFF
--- a/components/clientComponents/forms/AddressComplete/actions.ts
+++ b/components/clientComponents/forms/AddressComplete/actions.ts
@@ -353,14 +353,11 @@ const incrementAndCheckRateLimiting = async (
 ): Promise<boolean> => {
   try {
     const rateLimitMax = Number(await getAppSetting(RATE_LIMIT_MAX_SETTING[scope]));
-
     const rateLimitWindowSeconds = Number(
       await getAppSetting("addressCompleteRateLimitWindowSeconds")
     );
-
     if (!isPositiveSafeInteger(rateLimitMax) || !isPositiveSafeInteger(rateLimitWindowSeconds)) {
-      logMessage.error(new Error("Invalid AddressComplete rate limit configuration"));
-      return false;
+      throw new Error("Invalid configuration");
     }
 
     const redis = await getRedisInstance();
@@ -388,7 +385,10 @@ const incrementAndCheckRateLimiting = async (
       );
     }
     return isRateLimited;
-  } catch {
+  } catch (err) {
+    logMessage.error(
+      `AddressComplete rate limiting failed. Reason: ${err instanceof Error ? err.message : String(err)}`
+    );
     // Most likely case is Redis is down, just pass through to avoid breaking the form UX
     return false;
   }
--- a/components/clientComponents/forms/AddressComplete/actions.ts
+++ b/components/clientComponents/forms/AddressComplete/actions.ts
@@ -83,7 +83,8 @@ export const getAddressCompleteChoices = async (
     return { items: [], error: "API_KEY_MISSING" };
   }
 
-  // Avoid upstream calls for empty or very short queries that are unlikely to produce useful results.
+  // Do not call the API if the query is empty or only whitespace, as it will return a 1001 error (SearchTerm not supplied).
+  // Also avoid upstream calls for short queries that are unlikely to produce useful results.
   const normalizedQuery = normalizeQuery(query);
   if (normalizedQuery.length < MIN_ADDRESS_SEARCH_LENGTH) {
     return { items: [], error: null };
--- a/components/clientComponents/forms/AddressComplete/actions.ts
+++ b/components/clientComponents/forms/AddressComplete/actions.ts
@@ -9,7 +9,13 @@ import { getRedisInstance } from "@root/lib/integration/redisConnector";
 import { getClientIp } from "@root/lib/ip";
 import { getAppSetting } from "@root/lib/appSettings";
 import { isValidCanadaPostId, isValidCountryCode, isValidLanguage } from "./validation";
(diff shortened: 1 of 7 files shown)


components/clientComponents/forms/AddressComplete/actions.ts
```diff
@@ -77,13 +83,10 @@ export const getAddressCompleteChoices = async (
...
 
-  if (await incrementAndCheckRateLimiting(await getClientIp())) {
-    return { items: [], error: "RATE_LIMITED" };
-  }
-
   // Do not call the API if the query is empty or only whitespace, as it will return a 1001 error (SearchTerm not supplied).
+  // Also avoid upstream calls for short queries that are unlikely to produce useful results.
   const normalizedQuery = normalizeQuery(query);
-  if (!normalizedQuery || normalizedQuery === "") {
+  if (normalizedQuery.length < MIN_ADDRESS_SEARCH_LENGTH) {
```
timarney, line 89: I think this makes sense but I feel like it's kind of a UX thing ...

Given many website around the world use this type of component 

And even the example allows single char entry
https://www.canadapost-postescanada.ca/ac/learn-more/try-now

Does typing a single char give an indication to the user that "it's working" ?

I don't have a strong opinion either way 

cc: @samsadasivan

components/clientComponents/forms/AddressComplete/actions.ts
```diff
@@ -97,6 +100,10 @@ export const getAddressCompleteChoices = async (
     return { items: [], error: "INVALID_INPUT" };
   }
 
+  if (await incrementAndCheckRateLimiting(await getClientIp(), RATE_LIMIT_SCOPE.find)) {
+    return { items: [], error: "RATE_LIMITED" };
```
timarney, line 104: Just noting and this has changed as we've build out the feature but 

I don't believe we using error anymore ...

Maybe double check if this is going to log etc... as you like

packages/database/src/fixtures/settings.ts
```diff
@@ -76,25 +76,36 @@ const editLockRedirectIdleMs: Setting = {
...
+  internalId: "addressCompleteFindRateLimitMax",
+  nameEn: "Address Complete Find Rate Limit Max (count)",
+  nameFr: "Limite maximale de recherche d'adresse complète (nombre)",
   descriptionEn:
-    "The maximim number of requests (within the threshold) before the user is rate limited.",
+    "The maximum number of typeahead find requests (within the threshold) before the user is rate limited.",
   descriptionFr:
-    "Le nombre maximal de requêtes (dans la limite du seuil) avant que l'utilisateur ne soit soumis à une limitation de débit.",
+    "Le nombre maximal de requêtes de recherche avec saisie semi-automatique (dans la limite du seuil) avant que l'utilisateur ne soit soumis à une limitation de débit.",
+  value: "400",
```
timarney, line 87: @thiessenp-cds @craigzour 

Will the waf block before we reach this point?

craigzour: WAF limit is 2500 requests per 5 minutes. Is this `400` value the number of requests an individual can make within a certain period?

components/clientComponents/forms/AddressComplete/actions.ts
craigzour: I personally like this (I suggested it on Slack this morning on the caching thread). I think this is an easy way to save on a couple of requests (per use) that would not bring any relevant information. I would be surprised if people complain because nothing happens after they entered a single character in the field. Who does that anyway? 🤔

packages/database/src/fixtures/settings.ts
thiessenp-cds: Yup the window is set to 60 seconds. So 400 requests within 60 seconds. Or within 5 minutes 2000 requests.

components/clientComponents/forms/AddressComplete/actions.ts
samsadasivan: Hi! I am so sorry for the delay here - I think I am leaning towards triggering the search after 2-3 char. That may be better for anyone using a mobile device or magnifying tools. Maybe 2-3 characters and a 0.5 sec delay? If thats too much, just 2 characters - if this is something we can still do.

thiessenp-cds: Thanks for looking into this! I can increase the delay to 3 characters

samsadasivan: Thanks Pete - lets do 2! just so folks don't think its broken

packages/database/src/fixtures/settings.ts
thiessenp-cds: @ShadeWyrm what do you think of the rate limiting default of 400 requests per minute for "find" lookups? And 200 requests per minute for "retrieve" lookups?
Would it make sense to adjust them base on CanadaPost API's rate limiting?

craigzour (approved):
LGTM!

packages/database/src/fixtures/settings.ts
ShadeWyrm: I think it's worth a try at that base, I don't think we have anything but best guesses right now.

As long as we can adjust the number quickly I think it's fine.

PR #7755, opened 2026-08-19
fix: star rating excel issue

thiessenp-cds:
# Summary | Résumé

Excel treats a star rating output value as a date e.g. `4/5` becomes `05-May`

(Current implementation)
One option is to try to force Excel to treat the data "raw" this PR add CSV prefix `'` to avoid Excel treating as date e.g. `'4/5`. The downside to this approach is that any non-Excel spreadsheet or app that reads CSV will probably show the prefix and make the data look a little "messy".

Another option is to add spacing in the output format. Excel will probably then not parse it as a date e.g. instead of `4/5` the output would be `4 / 5`.


thiessenp-cds:
After some discussion we'll be going with a different direction. PR moved to https://github.com/cds-snc/platform-forms-client/pull/7774

PR #7760, opened 2026-08-19
chore: Restrict cross-origin access to server actions

dsamojlenko:
# Summary | Résumé

Adds middleware layer protection to all server actions to prevent XSS cross-origin requests.


proxy.ts
```diff
@@ -110,6 +111,18 @@ export default async function proxy(req: NextRequest, ctx: NextFetchEvent) {
         { status: 400 }
       );
     }
+
+    const origin = req.headers.get("origin");
```
bryan-robitaille: I was thinking about this on the drive home.  With the app behind the load balancer the origin it sees is the ECS task url and not the public one.  We'll need to use the `x-forwarded-host` header instead that that load balancer adds.
We can use the `getOrigin` function that you've already created in the past to get a sanitized version that can be used to compare against.  https://github.com/cds-snc/platform-forms-client/blob/main/lib/origin.ts

bryan-robitaille (changes requested):
We'll just need to look at the `x-forwarded-host` header instead of `origin` due to the way the app is hosted behind loadbalancers.

craigzour:
## :test_tube: Review environment
https://ekhtlycufn4j44iujzox4yjluy0qkzax.lambda-url.ca-central-1.on.aws/

bryan-robitaille (approved):

PR #7768, opened 2026-08-19
chore: address complete rate limiter update

thiessenp-cds:
# Summary | Résumé

Our focus has moved from protecting the entire Address Complete end point from abuse to mainly the retrieve requests. The reason is retrieve requests cost money while find requests are free. Plus if a malicious user decides to spam the endpoint with find requests, the AWS rate limiter will eventually block them.

This PR refactors the rate limiter logic to only rate limit retrieve requests. The only catch is that since a find request cannot be fully completed without a retrieve request (UX reasons), the find requests have a check if the user is rate limited and if so blocks find requests also.

## How to test

1. Update the "Address Complete Retrieve Rate Limit Max (count)" app setting to something easy to test like `2`
2. Open a published form with an Address Complete element
3. Type in and select an address twice
4. Try again and both the typeahead and retrieve results should be disabled (rate limit)
5. Wait a minute and try again and you should no longer be rate limited

### DIFF
--- a/components/clientComponents/forms/AddressComplete/rateLimiter.test.ts
+++ b/components/clientComponents/forms/AddressComplete/rateLimiter.test.ts
@@ -18,7 +18,7 @@ vi.mock("@lib/logger", () => ({
 import { getAppSetting } from "@root/lib/appSettings";
 import { getRedisInstance } from "@root/lib/integration/redisConnector";
 import { logMessage } from "@lib/logger";
-import { checkRateLimited, incrementAndCheckRateLimiting } from "./rateLimiter";
+import { isRateLimited, recordAndCheckRateLimit } from "./rateLimiter";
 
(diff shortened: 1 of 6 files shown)


components/clientComponents/forms/AddressComplete/rateLimiter.test.ts
```diff
@@ -0,0 +1,213 @@
+import { beforeEach, describe, expect, it, Mock, vi } from "vitest";
```
thiessenp-cds, line 1: FYI: AI helped me write these tests

craigzour (approved):
LGTM!

The `addressCompleteFindRateLimitMax` old setting will not be automatically deleted from the database so we will need to manually delete it (Staging + Prod). Also the new value for `addressCompleteRetrieveRateLimitMax` will have to be manually updated (Staging + Prod).

PR #7774, opened 2026-08-20
chore: star rating - store rating as a value 

thiessenp-cds:
# Summary | Résumé

Previously the Star Rating element stored the value as an object to make versioning more consistent and working with it in the output format easier. Unfortunately Excel treated the output value as a date e.g. `4/5` becomes `05-May`. 

After some discussion we decided on placing the rating (max) in the header and the value in the field (cell).  

The reasoning:
- versioning a value is not really a concern / primarily our responsibility, if a user changes e.g. a star rating max it's up to them to handle the output differences
- a traditional header description (max) and cell value would be more intuitive and easier to parse for automation

This PR updates the Star Rating to store in the DB as a value and have the rating max in an output header and the value in a cell.


thiessenp-cds:
Going to try another approach in a new PR - the star rating element will save as an object to help with versioning.

PR #7804, opened 2026-08-25
chore: Refactor responseDownload server action to use mapAnswers

dsamojlenko:
# Summary | Résumé

In https://github.com/cds-snc/platform-forms-client/pull/6204 the response mapping function that runs in the Responses download server action before passing the data off to transformers, was duplicated (and refactored a bit) into a package. It is this package and lib that are now used to map data in the Responses Pilot, but we never went back and refactored the Responses download server action to use it. The effect of which is that we've had to maintain two versions of essentially the same code as we've added features and components.

This PR refactors the Responses server action to use the same mapAnswers functions as Responses Pilot and also centralizes component-specific formatting functions in the component folders.


lib/responses/mapper/utils/toString.ts
```diff
@@ -1,25 +1,10 @@
-import { DateFormat, DateObject } from "@clientComponents/forms/FormattedDate/types";
-import { getFormattedDateFromObject } from "@clientComponents/forms/FormattedDate/utils";
-import { AddressElements } from "@clientComponents/forms/AddressComplete/types";
-import { getAddressAsString } from "@clientComponents/forms/AddressComplete/utils";
+import { DateFormat } from "@clientComponents/forms/FormattedDate/types";
+import { getFormattedDateResponse } from "@clientComponents/forms/FormattedDate/utils";
+import { getAddressCompleteResponse } from "@clientComponents/forms/AddressComplete/utils";
+import { getStarRatingResponse } from "@root/components/clientComponents/forms/StarRating/utils";
```
timarney: @lib :)

timarney (dismissed):

timarney (approved):

PR #7834, opened 2026-08-31
chore: remove hCaptcha feature flag 

thiessenp-cds:
# Summary | Résumé

This PR removes the hCaptcha feature flag. The reasoning: the feature is not going anywhere and will always be ON. 

## Testing

Open a published form and make sure it submits as expected.

### DIFF
--- a/app/(gcforms)/[locale]/(form filler)/id/[...props]/actions.ts
+++ b/app/(gcforms)/[locale]/(form filler)/id/[...props]/actions.ts
@@ -27,11 +27,6 @@ import { shouldCheckCaptcha } from "@lib/utils/shouldCheckCaptcha";
 
 import { randomUUID } from "crypto";
 
-// Block submissions when hCaptcha verification fails, including failed scores, API failures after
-// retries, missing or expired tokens, and invalid configuration.
-// ⚠️ Set to false to allow submissions despite verification failures.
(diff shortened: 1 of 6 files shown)


app/(gcforms)/[locale]/(form filler)/id/[...props]/actions.ts
```diff
@@ -27,6 +27,11 @@ import { shouldCheckCaptcha } from "@lib/utils/shouldCheckCaptcha";
 
 import { randomUUID } from "crypto";
 
+// Block submissions when hCaptcha verification fails, including failed scores, API failures after
+// retries, missing or expired tokens, and invalid configuration.
+// ⚠️ Set to false to allow submissions despite verification failures.
+const HCAPTCHA_FAILURE_BLOCKING_ENABLED = true;
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form filler)/id/[...props]/actions.ts]: Thinking we should remove this and just add a note inside of verifyHCaptchaToken  --- not seeing the benefit to a const here :)

Comment (thiessenp-cds) [reply app/(gcforms)/[locale]/(form filler)/id/[...props]/actions.ts]: Re-thinking it, if we ever need to quickly remove hCaptcha the below IF statement is easy to remove. Sure, will remove it.

timarney (approved):

PR #7837, opened 2026-08-31
chore: move hCaptcha related code to a module

thiessenp-cds:
# Summary | Résumé

Moves hCaptcha related code to a module. The module was published to: 
https://www.npmjs.com/package/@gcforms/hcaptcha

_Note: GCForms will be updated to use the module in a separate PR._

The package currently has two explicit public subpaths:

```ts
@gcforms/hcaptcha/client
@gcforms/hcaptcha/server
```

This separate client/server directories provide clear boundaries, keeps server-only code out of browser bundles, and create focused build outputs. They also make the package API explicit reduce accidental cross-environment imports.


packages/hcaptcha/src/client/useHCaptcha.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+}));
+
+const HookHarness = ({
+  captchaEnabled,
+  failureMode,
+  onCaptchaExpired,
+  onError,
+  onResult,
+}: {
+  captchaEnabled?: boolean;
```
timarney: `captchaEnabled`

Do we need this?

If your using the package I think you would manage this outside of the component

packages/hcaptcha/src/client/useHCaptcha.tsx
```diff
@@ -0,0 +1,156 @@
+"use client";
+
+import HCaptcha from "@hcaptcha/react-hcaptcha";
+import type { ReactNode } from "react";
+import { useCallback, useRef } from "react";
+
+export type HCaptchaFailureMode = "allow" | "block";
+
+export type UseHCaptchaOptions = {
+  captchaEnabled?: boolean;
```
timarney: Per note above I don't think we need this  

```
captchaEnabled?: boolean;
```

packages/hcaptcha/src/client/useHCaptcha.tsx
```diff
@@ -0,0 +1,156 @@
...
+  execute: () => Promise<HCaptchaExecutionResult>;
+  reset: () => void;
+};
+
+const CONFIG_ERROR_CODES = ["invalid-sitekey", "missing-sitekey"];
+
+// Provides CAPTCHA behavior without owning a form, so consumers can integrate execution and reset
+// with their own submission flow, including forms that use uncontrolled inputs.
+export const useHCaptcha = ({
+  captchaEnabled = true,
```
timarney: same

packages/hcaptcha/src/client/useHCaptcha.tsx
```diff
@@ -0,0 +1,156 @@
...
+  }, [captchaEnabled, complete, failureResult]);
+
+  const onVerify = useCallback(
+    (verifiedToken: string) => {
+      complete({ verified: true, token: verifiedToken });
+    },
+    [complete]
+  );
+
+  const captcha = captchaEnabled ? (
```
timarney: same

packages/hcaptcha/tsdown.config.mts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from "tsdown";
```
thiessenp-cds, line 1: The `.mts` extension over `.ts`. This marks it as a module and avoids a Node warning about it having to reparse the config as ES module syntax

vitest.config.ts
```diff
@@ -49,6 +49,8 @@ export default defineConfig({
     tsconfigPaths: true,
     alias: sharedAliases,
     extensions: [".mjs", ".js", ".mts", ".ts", ".jsx", ".tsx", ".json"],
+    // Ensure all workspace packages share the same React instance during tests
+    dedupe: ["react", "react-dom"],
```
thiessenp-cds, line 53: Looks like module vs not can use different instances of React for VItests. This should ensure all tests use the same instance of React.

timarney (approved):
Let's merge this to get it published and we can iterate as need be,

PR #7846, opened 2026-09-01
chore: update Forms to use hCaptcha module

thiessenp-cds:
# Summary | Résumé

This PR updates Forms hCaptcha handling to use the @gcforms/hcaptcha module.

The client hook now owns widget lifecycle, expiry/reset handling, failure states, and duplicate execution protection. FormCaptcha adapts that behavior for the existing Forms/Formik flow, including stale submission protection. The server verifier now provides typed results, configurable score limits, timeouts, and retry handling with exponential backoff.

The old app-local hCaptcha helpers were removed, tests and documentation were updated, and resolver aliases were added so the client/server entrypoints work in Vitest and Next builds.


tsconfig.json
```diff
@@ -29,6 +29,8 @@
       "@i18n/*": ["./i18n/*"],
       "@content/*": ["./public/static/content/*"],
       "@public/*": ["./public/*"],
+      "@gcforms/hcaptcha/client": ["./packages/hcaptcha/src/client/index.ts"],
+      "@gcforms/hcaptcha/server": ["./packages/hcaptcha/src/server/index.ts"],
```
thiessenp-cds, line 33: Allows tests to find the modules. Something like this is needed because the below option will not look in sub directories like client and server.
Another option would be to update the below to include nested subdirectories. I think that could potentially pull in files not needed.

components/clientComponents/forms/Form/Form.tsx
```diff
@@ -41,20 +45,32 @@ import { uploadFile } from "@root/app/(gcforms)/[locale]/(form filler)/id/[...pr
 
 import { SaveAndResumeButton } from "@clientComponents/forms/SaveAndResume/SaveAndResumeButton";
 import { LOCKED_GROUPS } from "@formBuilder/components/shared/right-panel/headless-treeview/constants";
+import { shouldCheckCaptcha } from "@root/lib/utils/shouldCheckCaptcha";
+import { isSuspiciousHCaptchaError } from "@clientComponents/globals/FormCaptcha/isSuspiciousHCaptchaError";
+
+type InternalCaptchaProps = {
```
timarney, line 51: Should these types be in the package and just imported here ?

app/(gcforms)/[locale]/(form filler)/id/[...props]/actions.ts
```diff
@@ -70,8 +73,18 @@ export async function submitForm(
       const shouldVerifyHCaptcha = shouldCheckCaptcha(template?.isPublished, isPreview);
 
       if (shouldVerifyHCaptcha) {
-        const captchaVerified = await verifyHCaptchaToken(captchaToken || "", formId);
-        if (!captchaVerified) {
+        const captchaSecret = process.env.HCAPTCHA_SITE_VERIFY_KEY;
```
Comment (timarney) [inline app/(gcforms)/[locale]/(form filler)/id/[...props]/actions.ts:78]: I think we should do a check here to see if the var is set to "" --- in which case we can early return --- no need to continue as we know it won't be verified

components/clientComponents/forms/Form/Form.tsx
```diff
@@ -416,3 +445,23 @@ export const Form = withFormik<FormProps, Responses>({
     }
   },
 })(InnerForm);
+
+// Keep the CAPTCHA handle outside the Formik higher-order component so its submit handler
+// can access the token and reset it
+export const Form: React.FC<FormProps> = (props) => {
+  const captchaFormRef = useRef<FormCaptchaHandle | null>(null);
+  const setCaptchaFormHandle = useCallback((handle: FormCaptchaHandle | null) => {
```
timarney: Can all these callback be provided by the package   --- provide the ref to 

useX --- seems like we'll want to do all the same for the SSO 

useCaptch(yourRef) => 

return the callbacks

packages/hcaptcha/src/client/HCaptchaForm.test.tsx
```diff
@@ -0,0 +1,202 @@
+/**
+ * @vitest-environment jsdom
```
timarney, line 2: Shouldn't this test be in the package ?

When we pull in packages we don't write tests for the package in our repo.

components/clientComponents/globals/FormCaptcha/FormCaptcha.tsx
```diff
@@ -1,111 +1,149 @@
...
-        if (hCaptchaRef.current) {
-          hCaptchaRef.current.execute();
+import { isSuspiciousHCaptchaError } from "./isSuspiciousHCaptchaError";
+
+export interface FormCaptchaHandle {
+  getToken: () => string | undefined;
+  reset: () => void;
+}
+
+type FormCaptchaProps = Omit<FormHTMLAttributes<HTMLFormElement>, "onError" | "onSubmit"> &
```
timarney: Can this component be exported from the package?

components/clientComponents/forms/Form/Form.tsx
thiessenp-cds: These callbacks are Formik-specific glue, so keeping them here keeps the hCaptcha package reusable and framework-agnostic. I think some of the hCaptcah app level wrapper code could be reused for future controlled-input apps like this. I was going to do it in a follow up PR but might as well do it now.

components/clientComponents/forms/Form/Form.tsx
thiessenp-cds: The type is a bit specific to the Formik Form implementation and may not be that reusable. I don't feel super strongly about it though.

timarney (approved):
Approving this one so we can keep moving forward in a more focussed PR i.e. non formik etc..

anikbrazeau:
Related https://github.com/cds-snc/platform-unified-accounts-user-portal/pull/432

PR #7857, opened 2026-09-02
chore: Remove notification flag

timarney:
# Summary | Résumé

Cleans up "always on" feature flag.

### DIFF
--- a/i18n/translations/en/admin-flags.json
+++ b/i18n/translations/en/admin-flags.json
@@ -10,10 +10,6 @@
   "featureStatus": "Status",
   "featureSwitch": "On/Off",
   "features": {
-    "zitadelAuth": {
-      "title": "Zitadel Auth",
-      "description": "Enable API authentication with Zitadel"
-    },
     "topBanner": {
       "title": "Banner",
       "description": "Enable warning banner to be visible"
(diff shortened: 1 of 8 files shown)


i18n/translations/en/admin-flags.json
```diff
@@ -10,10 +10,6 @@
   "featureStatus": "Status",
   "featureSwitch": "On/Off",
   "features": {
```
thiessenp-cds, line 12: Probably meant to keep this? (same with FR)

lib/cache/types.ts
```diff
@@ -1,12 +1,10 @@
 // TODO: in the future these could pulled in from default_flag_settings.json
 
 export const UserFeatureFlags = {
```
thiessenp-cds, line 3: Probably meant to keep this for now?

timarney: This was missed in the addressComplete feature flag removal

https://github.com/cds-snc/platform-forms-client/pull/7842


So figured I would clean it up here.

i18n/translations/en/admin-flags.json
timarney: Not tied to this PR but just flag cleanup --- 

I don't see anywhere we use 

 "zitadelAuth": {
 
 The key is 
 
 "zitadelLogin": {
      "title": "Zitadel Login",
      "description": "Enable Zitadel as a login method for users"
    },

thiessenp-cds (approved):

PR #7872, opened 2026-09-03
fix: Validation error message persists when navigating back

### LINKED ISSUE #7771: Clear errors at the top of a form when using “Back”
## Description

If a user clicks “Back,” remove the error messages from the previous page.
Error messages should only apply to the page where they occurred.

## Context

https://cds-snc.freshdesk.com/a/tickets/29577

> Error handling needs to be fixed. Currently when a user misses entering data (or making a selection) one to many multiple manual controls and then clicks on the Continue button, the system provides an error message dialogue box with all the error messages for that page and focus is in that dialogue box. Which is great!

> However, when the user clicks on the “Back” button, they system navigates the user to the previous page, but the error message dialogue box for the initial page is now displayed at the top of the previous page. On a “Back”, those error message are not valid and should have been removed. This is a useability/readability issue.

dsamojlenko:
# Summary | Résumé

Fixes #7771 

When navigating forward through a form, validation fires on the visible elements and will prevent navigation and display navigation errors if there are any.

When navigating back, the Formik errors object still contains the errors from the step we were just on.

This PR fixes two things:
- When navigating back, clear the Formik errors object
- When displaying errors, ensure that only errors belonging to the current visible group are displayed


Abi-Nada:
Gave the PR a quick test - LGTM! The issue no longer persists ✅

timarney (approved):

PR #7905, opened 2026-09-10
chore: add admin branding preview page

timarney:
# Summary | Résumé

Adds and admin page to make it easier to spot issues when we adjust branding which can impact all brand logos.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,6 @@
   "type": "module",
   "scripts": {
     "accounts:seed:local": "tsx scripts/generate-local-accounts.ts",
-    "branding:snapshot": "tsx scripts/branding-snapshot.ts",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -20,6 +20,7 @@
   "type": "module",
   "scripts": {
     "accounts:seed:local": "tsx scripts/generate-local-accounts.ts",
+    "branding:snapshot": "tsx scripts/branding-snapshot.ts",
```
craigzour: I can't find any `scripts/branding-snapshot.ts` in this PR

timarney: Right ya will remove --- was an earlier attempt before I had this idea.

timarney: https://github.com/cds-snc/platform-forms-client/pull/7905/changes/9ea51acdc4c7ea91c10baf7524a74526c618a047

craigzour (approved):
Very good idea :)

Issue #7884, opened 2026-09-09
Branding larger than expected in production


anikbrazeau:
Found as part of [#incident-2026-09-09-form-has-groups-set-to-empty-object](https://gcdigital.slack.com/archives/C0C0QMQKQ5S)

<img width="949" height="887" alt="Image" src="https://github.com/user-attachments/assets/f00cfb0e-027f-424b-af38-e9a2325df888" />

timarney:
Note for whoever is working on this we need to ensure it plays nice with https://github.com/cds-snc/platform-forms-client/pull/7746 zooming.

timarney:
Added an admin page to help spot logo issues

https://github.com/cds-snc/platform-forms-client/pull/7905

timarney:
PR: 

https://github.com/cds-snc/platform-forms-client/pull/7907

This updates multiple brand logos that have the same issue

PR #7907, opened 2026-09-10
fix: remove widths and heights from problematic brand logos

timarney:
# Summary | Résumé

Updates to remove the hardcoded widths and heights from some brand logos

Allow the container to better control the widths.


dsamojlenko (approved):

timarney:
fixes: https://github.com/cds-snc/platform-forms-client/issues/7884

PR #7914, opened 2026-09-14
chore: add hCaptcha app setting

thiessenp-cds:
# Summary | Résumé

Adds a new app setting `hCaptchaEnabledSetting` to turn enable or disable hCaptcha. This replaces the old feature flag with a more permanent way to control hCaptcha. When set to "true" hCaptcha will run on both the client and server action. When set to "false" hCaptcha will not run anywhere.

## Testing

Because an app setting is required this can only be tested on Staging. To see whether hCaptcha loads look at the browser requests. This should only happen when `hCaptchaEnabledSetting` is set to true.

You can also test this locally but it takes a few more steps.


app/(gcforms)/[locale]/(form filler)/id/[...props]/page.tsx
```diff
@@ -84,7 +87,7 @@ export default async function Page(props0: {
         saveAndResume={saveAndResume}
         footer={footer}
       >
-        <GCFormsProvider formRecord={formRecord}>
+        <GCFormsProvider formRecord={formRecord} hCaptchaEnabledSetting={hCaptchaEnabledSetting}>
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(form filler)/id/[...props]/page.tsx:90]: In the future if more app settings are used, `hCaptchaEnabledSetting` could be replaced with a list of app settings.

timarney (approved):
Approving so this can be tested on staging

PR #7934, opened 2026-09-17
fix: skip-link app wide

### LINKED ISSUE #7897: “Skip to main content” button was unresponsive when tested with VoiceOver and Chrome.
From client: results of their form in fable testing:

> Ran 20 accessibility tests using Fable (10 in English and 10 in French). The accessibility tests mainly focused on screen reader compatibility, with a couple tests using screen magnification tools.

https://cds-snc.freshdesk.com/a/tickets/30088

**1 issue related to unresponsive “Skip to main content button”** (tester used Voice Over with Chrome). The button did not respond when they tried to select it.

> "On the landing page, the "skip to main content" link appears to do nothing."

**Relevant WCAG criteria: 2.1.1 Keyboard**

thiessenp-cds:
# Summary | Résumé

Fix skip-link navigation so keyboard focus moves reliably to the main content landmark.

## Changes

- Replaced `next/link` with a native `<a href="#content">`.
- Added `tabIndex={-1}` to application-level `<main id="content">` targets
- Ensured each page renders one skip link before the header
- Removed duplicate skip links rendered by shared header components

## Rationale

A skip link uses same-page fragment navigation. A native anchor lets the browser handle that navigation directly, while `next/link` routes through Next.js client-side navigation and does not reliably reproduce native fragment-focus behavior. See [Next.js Link](https://nextjs.org/docs/app/api-reference/components/link).

`tabIndex={-1}` makes `<main>` programmatically focusable without adding it to the normal Tab order. This ensures focus moves from the skip link to the main content, after which pressing Tab continues to the first focusable control.

Some skip-link implementations work without a tabindex, depending on the target element and surrounding document structure. Using `tabIndex={-1}` makes the focus target explicit and provides more consistent behavior. See [GOV.UK Skip link](https://design-system.service.gov.uk/components/skip-link/).

The skip link is rendered before the header as the first meaningful interactive element. This avoids duplicate links and follows the recommended skip-link structure. See [W3C Technique G1](https://www.w3.org/WAI/WCAG22/Techniques/general/G1) and [WCAG 2.4.1: Bypass Blocks](https://www.w3.org/WAI/WCAG22/Understanding/bypass-blocks.html).

## Testing

Try keying the skip-link and it should focus the main landmark area.
Do this from a number of pages but especially from the form-builder and a forms-form.


app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/PreviewClosed.tsx
```diff
@@ -58,7 +58,7 @@ export const PreviewClosed = ({ closedDetails }: { closedDetails: ClosedDetails
       <PreviewNavigation />
       <div className="h-12"></div>
       <div
-        className={`mb-8 border-3 border-dashed border-blue-focus bg-white p-4 ${
+        className={`border-blue-focus mb-8 border-3 border-dashed bg-white p-4 ${
```
Comment (thiessenp-cds) [inline app/(gcforms)/[locale]/(form administration)/form-builder/[id]/preview/PreviewClosed.tsx]: Note: Tailwind class order fixes are also in this PR. The prettier --write ran during the Husky hook automatically. I left them in rather than trying to do a force commit without.

PR #7937, opened 2026-09-17
fix: update hint text to be more readable with magnifiers

thiessenp-cds:
# Summary | Résumé

**The hint text is not a WCAG failure. Automated contrast testing passes**, and the text remains properly associated with its input through `aria-describedby`. This is mainly a UX: concern: the original styling used small text, muted colour, and a 75% width, which can make hints harder to read and follow for people using screen magnification.

The related GCDS implementation is the `gcds-hint` component in `gcds-components`. Its CSS uses `--gcds-hint-text` and `--gcds-hint-margin` and small-text tokens: 1rem on mobile and 1.125rem on desktop, with 150% to 155% line height.

A small change was made to make the text slightly darker and improve wrapping under magnification. This seemed the smallest change to help resolve the issue. We'd need to do user testing to be sure.

| | Before | After |
| -------- | -------- | -------- |
| 100%  | Screenshot 2026-09-17 at 2 33 18 PM   | Screenshot 2026-09-17 at 2 14 18 PM   |
| 200%   | Screenshot 2026-09-17 at 2 33 35 PM   | Screenshot 2026-09-17 at 2 14 10 PM   |


timarney:
We should be using the css variable from the design system 

Screenshot 2026-09-22 at 9 41 04 AM

timarney:
Try 

 color: var(--gcds-hint-text);


https://github.com/cds-snc/gcds-components/blob/main/packages/web/src/components/gcds-hint/gcds-hint.css

PR #7958, opened 2026-09-22
chore: Create local shims for element type-specific types

dsamojlenko:
# Summary | Résumé

Adding local shims for types to reduce duplication.


packages/types/src/index.ts
```diff
@@ -81,9 +82,9 @@ import {
   AddressCompleteLabels,
 } from "./address-complete";
 
+export { AddressCompleNext };
```
timarney: Typo

timarney (approved):

